PR #1, opened 2015-08-13
Add OCaml windows setup directions

kg:


kg:
Rebased; revised based on andreas's suggestion; added detail on right way to get ocaml on linux

PR #9, opened 2015-08-18
Add command (asserteq (invoke ...) expected)

lukewagner:
To write unit tests, it'd be useful to be able to assert an expected return value, otherwise all we're testing is that an error isn't hit.

The SExpr syntax is `(asserteq (invoke export-index args-to-call) expected-results)`.  The reason for the nested `invoke` is to clearly separate invoke args from expected results and to allow us to asserteq other forms later.  An alternative would have been to just have an `assert` that evaluates a boolean condition, but that would require injecting `Invoke` into general expressions or using an `Ast.Call` expression which violate the point of exports (all execution in a module originates with exports; `Eval.eval` sorta violates this already, but I was thinking of changing it to mock up a new module instead of reusing the current module).

### DIFF
--- a/ml-proto/src/lexer.mll
+++ b/ml-proto/src/lexer.mll
@@ -253,6 +253,7 @@ rule token = parse
   | "table" { TABLE }
 
   | "invoke" { INVOKE }
+  | "asserteq" { ASSERTEQ }
 
   | ";;"[^'\n']*eof { EOF }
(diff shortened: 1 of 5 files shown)


lukewagner:
Thanks, updated.

lukewagner:
Oops, I pushed updates to wrong branch, one sec...

PR #10, opened 2015-08-18
Don't Eval.eval in existing module

lukewagner:
Currently `Eval.eval` evaluates expressions inside the current wasm module (kinda like a JS direct eval) which feels like it violates module encapsulation invariants.  This patch mocks up a dummy module.


lukewagner:
Thanks for fixing the nits and merging.  I was pulled away by a meeting or I would've done it for you; happy to learn the right ML style.

PR #18, opened 2015-08-21
Fix mutability bug

lukewagner:
Manifests when running test suite with multiple file arguments.


rossberg-chromium:
Damnit, that's what you get for using state... :)

lukewagner:
I can hear the Haskellers laughing at us now.

rossberg-chromium:
Yes, at first, I was actually thinking of using a monad there, but that's a
bit too cumbersome in OCaml.

PR #20, opened 2015-08-21
Less verbose runtests output. Don't ocamlbuild by default (pass --build as an arg to build).

kg:
Made test run output silent by default (unless it fails). Removed the warning for no expected output (you can uncomment to turn it back on).

ocamlbuild isn't run to build main.native by default anymore. You can do ```./runtests.py --build``` if you want it to be built for you.


rossberg-chromium:
LGTM

I have another feature wish. Can the script check the exit code of the interpreter? Currently, if it fails with e.g. a syntax error, but there is no expected output, it is regarded okay. (I changed the interpreter today so that it returns a non-zero error code on any error.)

kg:
It already checks the exit code.

```python
self.assertEqual(0, exitCode, "test runner failed with exit code %i" % exitCode)
```

```tee``` might have been swallowing the exit code, but I've taken it out, so you should be good.

rossberg-chromium:
Ah, indeed, with your change it already works. Thanks!

PR #21, opened 2015-08-23
Adding Travis build and test of ml-proto.

flagxor:


rossberg-chromium:
Would it be possible to move the shell files to a dedicated subdirectory (say, travis)? It seems like different people like to drop different build/package systems, and it's probably preferable to not have all of their auxiliaries pollute the root dir. Would also be good to have a little README in that directory, explaining its purpose.

flagxor:
Pushed revised version.
PTAL

lukewagner:
I don't know anything about Travis (pretty amazed to see the GH integration), but is it possible to create some sort of image to capture the ocaml installation so it doesn't have to re-run installation on every push?

PR #22, opened 2015-08-25
Add invalid command

lukewagner:
Two patches (on top of #14):

The first patch adds an `(invalid (module ...) "error")` command that checks that the given module fails in the checker (as opposed to during parsing which is outside the spec) with the given `"error"`.  This command is then used to test the corner cases of the `memory` section.

The second patch added tests for the corner cases of exports and caught a minor "bug"&mdash;the AST is allowed to export a single name twice&mdash;which is fixed.


lukewagner:
Oops, this should do it.

rossberg-chromium:
LGTM

PR #27, opened 2015-08-25
Implement unsigned operators

rossberg-chromium:
@lukewagner, @kg, it's rather ugly, going through big_int, but seems to work.


kg:
For consistency and clear definitions, why not make all int ops work using big_int (of size 32 bits/64 bits, respectively)? I feel like that might be a good move for the proto.

lukewagner:
Looks generally good to me.  This made me realize that, without a single-precision ieee754 type, we'll have a hard time representing the precise semantics of `float32` ops.  Probably most tests won't catch the difference, but I bet @sunfishcode will write some...

rossberg-chromium:
Forgot to git add the test file earlier.

@kg, hm, not sure if that would be clearer. You would enlarge the int32/64 value domain with illegal values, and you'd have to implement the wrap-around masking manually. Currently, the hacks are at least localised in a few operators.

@lukewagner: https://github.com/WebAssembly/spec/pull/29

PR #29, opened 2015-08-26
Implement accurate float32 semantics

rossberg-chromium:
@lukewagner, @kg, doing it asm.js-style.

### DIFF
--- a/ml-proto/test/float32.wasm
+++ b/ml-proto/test/float32.wasm
@@ -1,0 +1,52 @@
+(module
+  (func $eq_float32 (param $x f32) (param $y f32) (result i32)
+    (eq.f32 (getlocal $x) (getlocal $y))
+  )
+
+  (func $eq_float64 (param $x f64) (param $y f64) (result i32)
+    (eq.f64 (getlocal $x) (getlocal $y))
+  )
+
+  (func $div_float32 (param $x f32) (param $y f32) (result f32)
+    (div.f32 (getlocal $x) (getlocal $y))
+  )
+
+  (func $div_float64 (param $x f64) (param $y f64) (result f64)
+    (div.f64 (getlocal $x) (getlocal $y))
+  )
+
+  (export "eq_float32" $eq_float32)
+  (export "eq_float64" $eq_float64)
+  (export "div_float32" $div_float32)
+  (export "div_float64" $div_float64)
+)
+
+(asserteq
+  (invoke "eq_float32" (const.f32 1.123456789012345) (const.f32 1.123456789))
+  (const.i32 1)
+)
+
+(asserteq
+  (invoke "eq_float64" (const.f64 1.123456789012345) (const.f64 1.123456789))
+  (const.i32 0)
+)
+
+(asserteq
+  (invoke "eq_float32" (const.f32 1e40) (const.f32 1e50))
+  (const.i32 1)
+)
+
+(asserteq
+  (invoke "eq_float64" (const.f64 1e40) (const.f64 1e50))
+  (const.i32 0)
+)
+
+(asserteq
+  (invoke "div_float32" (const.f32 1.123456789012345) (const.f32 100))
+  (const.f32 0.011234568432)
+)
+
+(asserteq
+  (invoke "div_float64" (const.f64 1.123456789012345) (const.f64 100))
+  (const.f64 0.01123456789012345)
+)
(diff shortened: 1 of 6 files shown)


ml-proto/test/float32.wasm
```diff
@@ -0,0 +1,52 @@
...
+  )
+
+  (export "eq_float32" $eq_float32)
+  (export "eq_float64" $eq_float64)
+  (export "div_float32" $div_float32)
+  (export "div_float64" $div_float64)
+)
+
+(asserteq
+  (invoke "eq_float32" (const.f32 1.123456789012345) (const.f32 1.123456789))
```
sunfishcode: For the AST syntax, it's desirable to issue an error when the literal value can't be exactly represented its type, rather than implicitly rounding it, as this test is testing, since rounding is an interesting operation from the perspective of a language spec.

There is also [a variety of opinions](https://github.com/WebAssembly/design/issues/292) about whether we even want a decimal syntax for floating point values, though I think it's fine for now for exactly representable values.

ml-proto/test/float32.wasm
```diff
@@ -0,0 +1,52 @@
...
+  (const.i32 1)
+)
+
+(asserteq
+  (invoke "eq_float64" (const.f64 1.123456789012345) (const.f64 1.123456789))
+  (const.i32 0)
+)
+
+(asserteq
+  (invoke "eq_float32" (const.f32 1e40) (const.f32 1e50))
```
sunfishcode: Similarly, 1e40 and 1e50 are outside the f32 range.

lukewagner:
This will get us much closer but, unfortunately, there are some ops for which the
```
  x,y : float32
  x f32.op y = float32( float64(x) f64.op float64(y) )
```
identity doesn't hold (e.g. modulus).  The only ops that I know are [proven](http://dl.acm.org/citation.cfm?id=221334) are +, -, *, /, and sqrt.

sunfishcode:
Fortunately, wasm doesn't have a floating point modulus operator :-). I think this technique covers all of the operators in wasm.

lukewagner:
Also, scanning for `float` in [asm.js](http://asmjs.org/spec/latest/#standard-library), ceil, floor, the comparisons.  Eliminating the transcendental functions eliminated a lot of the non-commutative ops.  I was about to ask about whether we could remove float mod (b/c it's in ml-proto), but then Dan just commented :)

rossberg-chromium:
Ah, good point. Removed float mod.

lukewagner:
lgtm.  I also felt the pain of the imprecise decimal representation in #24; perhaps we could open a new issue/PR to use one of the schemes from #292 as @sunfishcode suggested?

ml-proto/test/float32.wasm
rossberg-chromium: I agree, it would be good to have the front-end issue errors whenever there is rounding. But it seems like that would require implementing our own string-to-float conversion. I'm no real expert on that.

At least I changed the tests so that the don't rely on rounding for literals.

lukewagner: The proposed floating formats just represent the bits of the IEEE754
float32/float64 directly, so all you need at the end is an
`Int64.float_of_bits`.

On Wed, Aug 26, 2015 at 1:01 PM, rossberg-chromium <notifications@github.com

sunfishcode: A simple thing we could do for the parser for now would be to just test if string_of_float on the float value gives us a string equal to the string passed to float_of_string, and issue an error if it's not. That won't be everything we'll ever want, but it'll work for many simple purposes.

jfbastien:
@lukewagner's cross-repo link should have been https://github.com/WebAssembly/design/issues/292 :-)

ml-proto/test/float32.wasm
rossberg-chromium: Unfortunately, string representations for floats aren't unique. So that would prevent you from ever writing e.g. `1e10`, which I think would be rather bad in terms of usability.

But I agree that we should add hex float literals. I would still want to keep the ability to write human readable constants in the text format, though.

jfbastien: Hex float is perfectly readable / writable by humans :-)
More seriously, I think it's pretty important that text parsers reject numbers which aren't precisely representable. It also important that all parsers know exactly how to do this so that they all accept the same inputs! Hex float is the easiest way to get this, but I'm open to supporting more.

I'd like to make sure that we discuss any other representation in https://github.com/WebAssembly/design/issues/292, and that we write down the outcome so it's easy to search in the future. These PR discussions aren't searchable!

Issue #26, opened 2015-08-25
Bring names in line with AST semantics sooner rather than later


titzer:
Before we get too far along I think we need to have a referendum on the naming going on in the spec. In particular, I think the spec should follow naming conventions that have already been agreed upon in AstSemantics.md. Also, we should solve some of the outstanding issues with naming in AstSemantics.md.

In particular, the operation.type in the spec bothers me. This might come from a viewpoint where operations are parameterized over types in some way. That kind of doesn't make sense, since there are many floating point operations available that don't work on integers and vice versa. Therefore I think it's more logical to do what we were doing before and consider operations to be members of types.

E.g.

type.operation
type.operation[type]

kg:
We really need this. I've spent a while reading over the ml-proto and I still can't figure out how to mentally map name -> ast node -> name. The way it works currently is very confusing to me.

rossberg-chromium:
I'm fine with inverting the naming scheme in the proto. FWIW, it was inspired by good old M68k syntax, not by parameterisation.

rossberg-chromium:
The biggest other discrepancy probably is with the zoo of conversion operators. One reason for the difference was that the naming in the AstSemantics is rather irregular, so kind of impossible to factorise in a pleasant manner. Can we perhaps do something about that?

sunfishcode:
Could you be more specific about the irregularity you observe? The current set of names is aimed at emphasizing semantic differences among the conversions.

rossberg-chromium:
@sunfishcode, I understand the motivation, but I'm not convinced it buys anything in a code format not targeted at humans anyway. In fact, find it rather difficult to memoize all the different names, and S-expr generators likewise will have to go through extra hoops to select them.

Conversions essentially form a 2D matrix (with a few holes; 3D if you want to count sign handling). It's unfortunate if different elements in this matrix require different naming schemes. (Or if both index dimensions are annotated in different ways, for that matter.)

rossberg-chromium:
I'll probably have plenty of time on the plane tomorrow, will do a first round of lining up naming.

sunfishcode:
@rossberg-chromium Humans finding it difficult to memorize all of the different names is a human-oriented concern :-). Using a single name to represent different kinds of things is an abstraction.

[...]

PR #30, opened 2015-08-27
Remove near|far split and rename (Get|Set)Memory

lukewagner:
This patch moves the linear memory operations a bit closer to the [design doc](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#linear-memory-operations).  It doesn't attempt to fix the type naming situation (#26) but it does:
* remove the near/far distinction since >4GiB heaps are [future feature](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#linear-memory-bigger-than-4gib).  The SExpr opcode names are future-compatible under the assumption that we bias toward near ops ("" means near, "far" means far);
* rename `get*` to `load` and `set*` to `store`.

### DIFF
--- a/ml-proto/src/ast.ml
+++ b/ml-proto/src/ast.ml
@@ -60,8 +60,7 @@ type binop = (Int32Op.binop, Int64Op.binop, Float32Op.binop, Float64Op.binop) op
 type relop = (Int32Op.relop, Int64Op.relop, Float32Op.relop, Float64Op.relop) op
 type cvt = (Int32Op.cvt, Int64Op.cvt, Float32Op.cvt, Float64Op.cvt) op
 
-type dist = Near | Far
-type memop = {dist : dist; align : Memory.alignment; mem : Memory.mem_type}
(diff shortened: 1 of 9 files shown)


MikeHolman:
Can you update in the readme as well? Otherwise lgtm

ml-proto/src/ast.ml
```diff
@@ -86,8 +85,8 @@ and expr' =
   | SetLocal of var * expr
   | GetGlobal of var
   | SetGlobal of var * expr
```
jfbastien: While you're here... ASTSemantics also names `load_global` and `store_global`.

rossberg-chromium:
Other than the README, lgtm.

ml-proto/src/ast.ml
lukewagner: You bet

lukewagner:
Merging based on lgtm

lukewagner:
Oh oops, I didn't see I only updated half the README.md, will update the "external syntax" now.

PR #31, opened 2015-08-27
Refine alignment

lukewagner:
Extending #30, this patch moves the alignment aspect of linear memory ops to what's in the [design repo](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#alignment) as follows:
* remove `Memory.alignment` from `Memory` since alignment makes no semantic difference
* add `Ast.alignment`, since load/store AST nodes do have an alignment field
* switches `alignment` from a binary `Aligned|Unaligned` to an `int` which is validated to be {1,2,4,8} and smaller than the element size

### DIFF
--- a/ml-proto/src/check.ml
+++ b/ml-proto/src/check.ml
@@ -240,8 +240,7 @@ and check_arm c t ts arm =
   check_expr c (if fallthru then [] else ts) e
 
 and check_memop memop at =
-  require (Lib.is_power_of_two memop.align) at "non-power-of-two alignment";
-  require (memop.align <= Memory.mem_size memop.mem) at "alignment too big"
+  require (Lib.is_power_of_two memop.align) at "non-power-of-two alignment"
 
 
 (*
--- a/ml-proto/src/check.ml
+++ b/ml-proto/src/check.ml
@@ -189,10 +189,12 @@ let rec check_expr c ts e =
     check_type [] ts e.at
 
   | Load (memop, e1) ->
+    check_memop memop e.at;
     check_expr c [Int32Type] e1;
     check_type [type_mem memop.mem] ts e.at
 
   | Store (memop, e1, e2) ->
+    check_memop memop e.at;
     check_expr c [Int32Type] e1;
     check_expr c [type_mem memop.mem] e2;
     check_type [] ts e.at
@@ -236,6 +238,13 @@ and check_arm c t ts arm =
   let {value = l; expr = e; fallthru} = arm.it in
   check_literal c [t] l;
   check_expr c (if fallthru then [] else ts) e
+
+and check_memop memop at =
+  require (memop.align = 1 || memop.align = 2 ||
+           memop.align = 4 || memop.align = 8)
+    at "invalid alignment";
+  require (memop.align <= (Memory.mem_size memop.mem))
+    at "alignment too big"
 
 
 (*
--- a/ml-proto/src/check.ml
+++ b/ml-proto/src/check.ml
@@ -243,7 +243,7 @@ and check_memop memop at =
   require (memop.align = 1 || memop.align = 2 ||
            memop.align = 4 || memop.align = 8)
     at "invalid alignment";
-  require (memop.align <= (Memory.mem_size memop.mem))
+  require (memop.align <= Memory.mem_size memop.mem)
     at "alignment too big"
 
 
--- a/ml-proto/src/check.ml
+++ b/ml-proto/src/check.ml
@@ -240,11 +240,8 @@ and check_arm c t ts arm =
   check_expr c (if fallthru then [] else ts) e
 
 and check_memop memop at =
-  require (memop.align = 1 || memop.align = 2 ||
-           memop.align = 4 || memop.align = 8)
-    at "invalid alignment";
-  require (memop.align <= Memory.mem_size memop.mem)
-    at "alignment too big"
+  require (Lib.is_power_of_two memop.align) at "invalid alignment";
+  require (memop.align <= Memory.mem_size memop.mem) at "alignment too big"
 
 
 (*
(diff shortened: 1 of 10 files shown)


ml-proto/src/check.ml
```diff
@@ -241,6 +239,13 @@ and check_arm c t ts arm =
   check_literal c [t] l;
   check_expr c (if fallthru then [] else ts) e
 
+and check_memop memop at =
+  require (memop.align = 1 || memop.align = 2 ||
+           memop.align = 4 || memop.align = 8)
```
sunfishcode: Since we're checking that the alignment is at most the access size below, would it make sense to just check that the alignment is a power of 2 here, instead of having an explicit list? That way it'll automatically generalize to SIMD types.

jfbastien: Agreed.

Extra advantage: subsequent indexed memory accesses can also key off the first access' alignment (base has an alignment, other indices don't because the alignment is implied).

ml-proto/test/memory.wasm
```diff
@@ -54,8 +66,8 @@
           (break)
         )
         (setlocal 2 (mul.i32 (getlocal 0) (const.i32 4)))
-        (setnears.i32 (getlocal 2) (getlocal 0))
-        (setlocal 1 (getnears.i32 (getlocal 2)))
+        (stores.i32 (getlocal 2) (getlocal 0))
+        (setlocal 1 (loads.i32 (getlocal 2)))
```
jfbastien: Is there a way for the test to validate what alignment this implicitly corresponds to? IIRC the design said "natural alignment" when unspecified?

rossberg-chromium:
Lgtm, other than README ;)

ml-proto/src/check.ml
lukewagner: Righto

ml-proto/test/memory.wasm
lukewagner: I think at the AST/spec level we shouldn't have an "unspecified" option, but rather leave this as a text-format-level syntactic sugar (as is the case in this PR) and a binary-format-level encoding optimization.  Of course we'd want tests for the text/binary format, but I think the way we should test this case (and surely others to follow) is to have testing commands that parse a given binary/text string and assert that the resulting AST has the expected values.

lukewagner:
Gosh, the README, it is my blind spot.  lmk if `Lib.is_power_of_two` is the right place to put this.

ml-proto/test/memory.wasm
```diff
@@ -26,6 +27,17 @@
   (module (memory 100 1000 (segment 0 "a") (segment 2 "b") (segment 1 "c")))
   "data segment not disjoint and ordered")
 
+;; Test alignment annotation rules
+(assertinvalid (module (func (loadu.2.i8 (const.i32 0)))) "alignment too big")
```
jfbastien: Too big alignment is [part of the design](https://github.com/WebAssembly/design/blob/ce509cdb9ec33fc439e389c82dffc1e641c1925d/AstSemantics.md#alignment) so this test isn't correct.

ml-proto/test/memory.wasm
```diff
@@ -26,6 +27,17 @@
...
+(assertinvalid (module (func (loadu.4.i16 (const.i32 0)))) "alignment too big")
+(assertinvalid (module (func (loadu.8.i32 (const.i32 0)))) "alignment too big")
+(assertinvalid (module (func (load.8.f32 (const.i32 0)))) "alignment too big")
+(assertinvalid (module (func (loadu.0.i64 (const.i32 0)))) "invalid alignment")
+(assertinvalid (module (func (loadu.3.i64 (const.i32 0)))) "invalid alignment")
+(assertinvalid (module (func (loadu.5.i64 (const.i32 0)))) "invalid alignment")
+(assertinvalid (module (func (loadu.6.i64 (const.i32 0)))) "invalid alignment")
+(assertinvalid (module (func (loadu.7.i64 (const.i32 0)))) "invalid alignment")
```
jfbastien: I think "non-power-of-two alignment" is a better reason.

ml-proto/test/memory.wasm
lukewagner: The design document isn't expected to be a complete enumeration of all invalid conditions, that's what `spec` is for and so this test *is* correct, wrt the spec as updated by this PR.  If you would like to disagree and argue that we *should* allow too-big alignment in the spec, that's another matter entirely, we shouldn't expect the high-level design repo to have to reiterate every single minute detail; that's not its purpose.

jfbastien: Agreed that the design doc isn't expected to be complete, but this specific case was discussed when creating the design, and we agreed that alignment just had to be a positive power of 2, no limits. This spec PR contradicts the design and what we agreed on, so the design is now wrong.

In cases like this the discussion should occur in the design repo: it revisits a decision that was made in the design repo. What the point of agreeing on anything in the design repo otherwise? I'm not saying that the design is golden and shouldn't change (witness the issues I've filed recently), I'm saying that we want to avoid diverging slightly in every place we implement wasm.

It's great the the spec repo finds unforeseen problems with the design, but this one wasn't unforeseen!

lukewagner: I was not aware that anyone had discussed this issue; could you provide a link?  This seems like the sort of decision that should be explicitly documented since it's nonobvious whether "power of 2" is just waving its hands at the boundary conditions (as we often do) or really means "no, seriously, every power of 2!" (up to what limit?).  I'm happy to remove this check here for now since ultimately it's a minor detail and people have actually discussed it.

ml-proto/test/memory.wasm
lukewagner: Anyhow, now it's explicit in code and a test; the process is working!

ml-proto/test/memory.wasm
lukewagner: Anyhow, now it's explicit in code and a test; the process is working!

lukewagner:
Merging based on lgtm above and no outstanding concerns.

PR #33, opened 2015-08-28
Rename opcodes to match AST semantics

rossberg-chromium:
@lukewagner, @kg, PTAL. Does not reverse S-expr opcode order yet.


ml-proto/README.md
```diff
@@ -175,7 +174,7 @@ expr:
   ( break )                           ;; = (break 0)
   ( switch.<type> <expr> <case>* <expr> )
   ( call <var> <expr>* )
-  ( dispatch <var> <expr> <expr>* )
+  ( callindirect <var> <expr> <expr>* )
```
lukewagner: `call_indirect`

ml-proto/src/lexer.mll
```diff
@@ -210,35 +210,34 @@ rule token = parse
...
-  | "convert."(fxx as t)".f32"
-    { CONVERT (floatop t F32.ToFloat32 F64.ToFloat32) }
-  | "convert."(fxx as t)".f64"
-    { CONVERT (floatop t F32.ToFloat64 F64.ToFloat64) }
-
-  | "cast.i32.f32" { CONVERT (Values.Int32 I32.ToFloatCast) }
-  | "cast.i64.f64" { CONVERT (Values.Int64 I64.ToFloatCast) }
-  | "cast.f32.i32" { CONVERT (Values.Float32 F32.ToIntCast) }
-  | "cast.f64.i64" { CONVERT (Values.Float64 F64.ToIntCast) }
+  | "extends.i32.i64" { CONVERT (Values.Int64 I64.ExtendSInt32) }
```
lukewagner: While you said you're waiting to move result type to the front, perhaps you could switch `*s` to `_signed` in this PR?

lukewagner:
Looks like a step in the right direction.  I think the semantics of float-to-int conversion still needs more work to [match](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#datatype-conversions-truncations-reinterpretations-promotions-and-demotions).

rossberg-chromium:
Reversed opcode order. Also made memory ops match AST semantics.

I tentatively used "/" to separate opcode index parameters, since using "[]" there does not fit the S-expression format all too well. Let me know what you think.

rossberg-chromium:
@lukewagner, @kg, do you think this part is good to land? The float truncation semantics should probably be addressed elsewhere (Dan has a WIP patch).

lukewagner:
For brackets vs. slash: ideally, the opcode names in the SExpr format would *exactly* match those in AstSemantics.md.  Since we're not actually trying to feed our SExpr language to a real LISP, is it ok if we bend the SExpr rules a bit and allow `[`/`]` in the identifier names?  Or is there a basic parsing ambiguity?

titzer:
+1 to exact match

-1 to some of the names currently in AstSemantics
-1 to brackets

And there went my karma for today

rossberg-chromium:
Using `[]` wouldn't currently cause a problem with the parser, but would still be rather odd. Maybe that's just me, though.

I'll be off the grid soon, so I hope it's okay if I land the current patch as is and leave the final bikeshedding for later. Should be trivial to tweak separately.

I'll also leave further changes to memory access ops to a future patch, since we are still discussing the best approach. This patch remains close to what's currently in AstSemantics, so should be good.

lukewagner:
That makes sense; this patch already puts us a lot closer.

PR #35, opened 2015-08-29
Cross-compiled C# wasm tests

kg:
This isn't necessarily something we want to merge yet, but I wanted to make a PR so people can look at it and we can start a discussion.

I did some work to cross-compile simple test cases from C# to wasm. I have some basic stuff working. CountUp passes; Fannkuch starts, runs, then fails (which is progress, since it was hanging before!)

See https://gist.github.com/kg/3f66f50e2428d9d47a41 for the source to Fannkuch and https://gist.github.com/kg/4cd6e23a73ac232dcca5 for the source to CountUp.

My goal here was to start translating more-realistic functions and workflows from a source language to wasm, instead of just hand-writing wasm tests. This will be helpful in the interim until LLVM's wasm backend is working, and we might still want to have another source language for tests just to make sure we aren't overly focused on LLVM.

It would be great to get help stepping through fannkuch to figure out why it's broken. The current state of the interpreter is basically impossible to debug - if it hangs, it just hangs, and ```-t``` doesn't seem to trace the actual instructions being executed. I may take a shot at adding more instrumentation to the interpreter once I can figure out the code.

Some related code quality/syntax questions came up while porting these tests, so I'll file issues about them separately.

### DIFF
--- a/ml-proto/runtests.py
+++ b/ml-proto/runtests.py
@@ -9,6 +9,9 @@ import sys
 
 class RunTests(unittest.TestCase):
   def _runTestFile(self, shortName, fileName, interpreterPath):
+    # HACK: Windows python compatibility
+    fileName = fileName.replace("\\", "/")
+
     logPath = fileName.replace("test/", "test/output/").replace(".wasm", ".wasm.log")
     try:
       os.remove(logPath)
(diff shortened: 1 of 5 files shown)


titzer:
I think this is valuable work and having multiple frontends generating wasm
tests is a good sanity check. Thanks for doing this and keep it coming.

Is Fannkuch long-running? If so, let's try to separate it from "unit" tests
that try to poke on individual opcodes, features, memory behavior, control
structures, etc.

kg:
> Is Fannkuch long-running?
With n=7 the C# version takes about 1ms (including startup overhead) and n=8 takes about 5ms.

kg:
Added a working Sieve of Eratosthenes test (and implemented fp sqrt, since it needs it).

PR #38, opened 2015-08-29
Rename round to nearestint.

sunfishcode:
This is what the design repo calls it. And in particular, it's useful to
avoid calling it "round" because it's semantically different from both
JS's Math.round and C's round.

### DIFF
--- a/ml-proto/src/arithmetic.ml
+++ b/ml-proto/src/arithmetic.ml
@@ -185,7 +185,7 @@ struct
       | Ceil -> ceil
       | Floor -> floor
       | Trunc -> fun _ -> 0.0  (* TODO *)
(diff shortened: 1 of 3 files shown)


jfbastien:
Not for this PR< but it would be good to test that this is indeed the behavior of the ML prototype.

sunfishcode:
It's in the TestingTodo.md already :-) :
> - test that nearestint doesn't do JS-style Math.round or C-style round(3) rounding

rossberg-chromium:
LGTM, but why not simply name it "Nearest"? The Int suffix seems kind of redundant and inconsistent with the other float rounding operators.

sunfishcode:
I'm ok with "nearest". I updated the PR accordingly.

PR #39, opened 2015-08-29
Simplify memops

rossberg-chromium:
Eliminates mem_type in favour of providing access size and optional signedness flag directly in the memory opcodes.

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -152,18 +152,15 @@ The S-expression syntax is defined in `parser.mly`, the opcodes in `lexer.mll`. 
 
 ```
 type: i32 | i64 | f32 | f64
-memtype: <type> | i8 | i16
-
+size: 8 | 16 | 32 | 64
 value: <int> | <float>
 var: <int> | $<name>
 
 unop:  neg | abs | not | ...
 binop: add | sub | mul | ...
 relop: eq | neq | lt | ...
-sign: s|u
-align: 1|2|4|8|...
-memop: (<sign>.)?(<align>.)?
-cvtop: trunc_s | trunc_u | extend_s | extend_u | ...
+cvtop: wrap | trunc_<sign> | extend_<sign> | ...
+sign: s | u
 
 expr:
   ( nop )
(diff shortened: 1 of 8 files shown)


ml-proto/README.md
```diff
@@ -152,18 +152,15 @@ The S-expression syntax is defined in `parser.mly`, the opcodes in `lexer.mll`.
...
 var: <int> | $<name>
 
 unop:  neg | abs | not | ...
 binop: add | sub | mul | ...
 relop: eq | neq | lt | ...
-sign: s|u
-align: 1|2|4|8|...
-memop: (<sign>.)?(<align>.)?
-cvtop: trunc_s | trunc_u | extend_s | extend_u | ...
+cvtop: wrap | trunc_<sign> | extend_<sign> | ...
```
sunfishcode: I guess this was already present, but is there a reason why there are signed and unsigned forms of `trunc` here?

ml-proto/src/ast.ml
```diff
@@ -62,7 +62,8 @@ type binop = (Int32Op.binop, Int64Op.binop, Float32Op.binop, Float64Op.binop) op
 type relop = (Int32Op.relop, Int64Op.relop, Float32Op.relop, Float64Op.relop) op
 type cvt = (Int32Op.cvt, Int64Op.cvt, Float32Op.cvt, Float64Op.cvt) op
 
-type memop = {ty : Types.value_type; mem : Memory.mem_type; align : int}
+type memop =
+  {ty : Types.value_type; size : int; align : int; signed : bool option}
```
lukewagner: Perhaps `size : int` could be replaced by `mty : memory_type`?  That type seems a bit "tighter" (without ruling out interesting cases).  In theory, having `size` be a simple integer lets you deal with size independently of float vs. integer, but I didn't see that helping in practice here, although I could've missed it.

lukewagner:
Looks like another step in the right direction.  Since this doesn't end up exactly at the op name syntax in AstSemantics.md#linear-memory, perhaps good to start with a design-repo PR that changes the names to match here?

jfbastien:
+1 on editing design repo. I like this proposal, it would be good to tie it to https://github.com/WebAssembly/design/issues/82 as it seems to resolve the objections from that issue?

ml-proto/README.md
sunfishcode: Oh, right. This is as it should be.

rossberg-chromium:
I created issue https://github.com/WebAssembly/design/issues/326 -- I think there are two general options for simplifying loads & stores.

lukewagner:
Can this PR be closed now?

rossberg-chromium:
Yes, I think it has been superseded by Luke's patch.

PR #40, opened 2015-09-01
wasm-compatible float32 and float64 arithmetic

sunfishcode:
This PR is not yet ready for merging!

This is some exploratory code implementing wasm-compatible float64 arithmetic in OCaml. This is my first ML code and I'm still learning, so please don't be gentle. Also, it doesn't compile yet because I haven't figured out how to use int64 yet. And, it doesn't have tests yet. And nothing is using it yet. And it may be buggy. I'm interested in any and all feedback about the approach, how best to structure the code, coding style, or anything.

I expect to eventually refactor this code to support float32 as well.

### DIFF
--- a/ml-proto/src/Makefile
+++ b/ml-proto/src/Makefile
@@ -5,7 +5,7 @@
 NAME = wasm
 INCLUDES = -I host -I given -I spec
 MODULES = \
-  host/flags given/lib given/source spec/error \
+  host/flags given/lib given/source given/float32 given/float64 spec/error \
   spec/types spec/values spec/memory spec/ast \
   spec/check spec/arithmetic spec/eval \
   host/print host/builtins host/script \
--- a/ml-proto/src/given/float32.ml
+++ b/ml-proto/src/given/float32.ml
@@ -1,0 +1,95 @@
+(* WebAssembly-compatible float32 implementation *)
+
+(*
+ * We represent float32 as its bits in an int32 so that we can be assured that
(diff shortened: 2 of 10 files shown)


sunfishcode:
Updated to address review feedback. Lots of changes including using modules and abstract types, and using the new operators in spec code. This still lacks tests.

sunfishcode:
The functor refactoring is looking like it's going to take me a while, and this is starting to slow progress elsewhere; would you be ok if we deferred the functor refactoring and just checked in this branch as it stands here? The redundancy isn't actually that bad.

Also, while there are no tests here yet, I am planning to write extensive tests for all the code here, and it would actually be more convenient to submit these tests incrementally instead of bundling them all up into a single monolithic patch.

rossberg-chromium:
On 16 September 2015 at 18:14, Dan Gohman <notifications@github.com> wrote:

> The functor refactoring is looking like it's going to take me a while, and
> this is starting to slow progress elsewhere; would you be ok if we deferred
> the functor refactoring and just checked in this branch as it stands here?
> The redundancy isn't actually that bad.
>
> Also, while there are no tests here yet, I am planning to write extensive
> tests for all the code here, and it would actually be more convenient to
> submit these tests incrementally instead of bundling them all up into a
> single monolithic patch.
>

Sure. I'm also happy to do the refactoring after you land this version.

sunfishcode:
Ok, great. I'll merge this now and we can continue to iterate on this code in tree.

PR #42, opened 2015-09-01
Rename Mod to Rem.

sunfishcode:
The signed remainder operator in particular is not clearly described as
a modulus function, because it returns negative values when the dividend
is negative. This patch renames it from "mod" to "rem", which is also in
accordance with AstSemantics.md.


rossberg-chromium:
LGTM

PR #44, opened 2015-09-01
Sync the integer unary operators with AstSemantics.md.

sunfishcode:
AstSemantics.md presently has no integer neg, abs, or not operators. It
does have a popcnt operator.


rossberg-chromium:
LGTM, but I am rather confused why we have neg & abs for floats but not for ints. What's the motivation?

kg:
Same question. Shouldn't we just add int neg/abs/not to astsemantics? Is there a reason to omit them? All three seem useful and common

jfbastien:
The FP operations cannot be emulated with other FP operations (though they technically can be with integer operations and a healthy assumption that sign is MSB).

The int operations are redundant, and we've gone back and forth on which should be in the IR and which shouldn't.

sunfishcode:
We discussed integer `abs` at one point -- it's interesting because the alternative reads the operand twice, which blocks single-use expression tree formation, but there was concern with the overall number of opcodes. I did add it as a possible [future feature](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#additional-integer-operations) though.

jfbastien:
I think the != discussion led us to decide that it probably wasn't worth being stingy on opcodes, though. It's probably worth opening an issue for these, and basing our decision on data once we have large codebases?

titzer:
+1 for waiting on larger applications. I think we've got everything we need
for completeness and everything else is just convenience.

jfbastien:
So it sounds like for now this is lgtm to merge, and file an issue on the design repo to gather data?

sunfishcode:
Merging because I believe we basically have consensus here, at least with respect to bringing the spec repo into accordance with the design repo. If people want integer abs/neg/not or other things, they're welcome to pursue them in the design repo first.

PR #46, opened 2015-09-02
Add min and max operators.

sunfishcode:
This adds floating point min and max, from AstSemantics.md. They don't have semantics yet; I'm working on that in the [float-bits](https://github.com/WebAssembly/spec/tree/float-bits) branch.


rossberg-chromium:
LGTM

Issue #47, opened 2015-09-03
Calls to the embedder


jfbastien:
The design repo is pretty vague about how calls to the embedder work, mostly because we were waiting on an implementation to really get a good understanding. I''m hoping that the spec repo can try things out, and help inform the design.

Could the spec repo implement something so that I can e.g. compile "Hello World!" and print that string to the screen?

I think the current agreement is that embedder calls look like declared-but-not-defined functions in the module. It's pretty similar to what dynamic linking would do.

lukewagner:
Improvements/clarifications always welcome, but the design repo does state that the way host environments expose host-specific functionality through [builtin modules](https://github.com/WebAssembly/design/blob/master/Modules.md#imports-and-exports) (see also NonWeb.md).

Concretely, the way I was expecting this to be reflected in ml-proto:
* we'd add an `imports` field to `Ast.module` (reflecting [imports](https://github.com/WebAssembly/design/blob/master/Modules.md#imports-and-exports) in Modules.md)
* an import would roughly look like `(import @importName "func_name" "module_name" (signature ...))` such that you could `(call @importName ...)` (like everything else, the name would be optional and an index could be used instead).  `check.ml` would require the call's sig to match the import's sig.
* [per design repo](https://github.com/WebAssembly/design/blob/master/Modules.md#imports-and-exports), how an import is resolved is totally up to the host environment.  To reflect this in ml-proto, `eval.ml` would be extended so that `Eval.init` also took a list imported functions (assumed to match the signatures of `Ast.imports`) and the impl of `call` of an import simply called one of these functions.
* the host part of ml-proto (`script.ml`, which is not logically part of the spec, but part of the host environment) would then, after parsing the `Ast.module` from SExprs would iterate down the list of `module.imports`, look them up in a list of builtin functionality we'd like to expose in ml-proto (stdio, etc) to build the list of imports passed to `Eval.init` (or fail on unknown imports).

Thus, just like the C/C++ standard (Section 1.9 Program Execution: "The semantic descriptions in this International Standard define a parameterized nondeterministic abstract machine."), the wasm semantics would be parameterized by the host environment (the import arguments to `init`).  When we add `page_size`, that value would, I expect, be another parameter to `init`.

[...]

PR #48, opened 2015-09-04
Split files into given, spec and host dirs

lukewagner:
This patch splits the files into the three dirs as described in [#47](https://github.com/WebAssembly/spec/issues/47#issuecomment-137542803).


ml-proto/runtests.py
```diff
@@ -51,7 +51,7 @@ def find_interpreter(path):
 def rebuild_interpreter(path):
   print("// building %s" % path)
   sys.stdout.flush()
-  exitCode = subprocess.call(["ocamlbuild", "-libs", "bigarray, nums, str", "main.native"], cwd=os.path.abspath("src"))
+  exitCode = subprocess.call(["ocamlbuild", "-libs", "bigarray, nums, str", "-Is", "given, spec, host", "host/main.native"], cwd=os.path.abspath("src"))
```
sunfishcode: ml-proto/README.md also has a copy of this build command; as long as it's still there, would you mind updating it too?

rossberg-chromium:
lgtm

PR #49, opened 2015-09-04
Add assert_heap_eq script command to check contents of a heap region

kg:
Added an ```assert_heap_eq``` script command alongside ```assert_eq```. You specify a base offset into the module heap along with an expected series of bytes (as a string with escapes, like memory segments).

This allows straightforward tests for algorithms that don't merely produce a few bytes ([this test](https://github.com/WebAssembly/ilwasm/blob/master/third_party/tests/Strcat.cs), for example.)

### DIFF
--- a/ml-proto/src/host/lexer.mll
+++ b/ml-proto/src/host/lexer.mll
@@ -249,6 +249,7 @@ rule token = parse
   | "assert_invalid" { ASSERTINVALID }
   | "assert_eq" { ASSERTEQ }
   | "assert_trap" { ASSERTTRAP }
+  | "assert_heap_eq" { ASSERTHEAPEQ }
   | "invoke" { INVOKE }
 
   | name as s { VAR s }
--- a/ml-proto/src/host/parser.mly
+++ b/ml-proto/src/host/parser.mly
@@ -50,7 +50,7 @@ type context =
   {funcs : space; imports : space; locals : space; labels : int VarMap.t}
 
 let empty () = {map = VarMap.empty; count = 0}
-let c0 () =
+let empty_context () =
   {funcs = empty (); imports = empty ();
    locals = empty (); labels = VarMap.empty}
 
@@ -99,7 +99,7 @@ let anon_label c = {c with labels = VarMap.map ((+) 1) c.labels}
 %token CONST UNARY BINARY COMPARE CONVERT
 %token FUNC PARAM RESULT LOCAL MODULE MEMORY SEGMENT IMPORT EXPORT TABLE
 %token PAGESIZE MEMORYSIZE RESIZEMEMORY
-%token ASSERTINVALID ASSERTEQ ASSERTTRAP INVOKE
+%token ASSERTINVALID ASSERTEQ ASSERTTRAP ASSERTHEAPEQ INVOKE
 %token EOF
 
 %token<string> INT
@@ -330,7 +330,7 @@ module_fields :
       | None -> {m with memory = Some $1} }
 ;
 modul :
-  | LPAR MODULE module_fields RPAR { $3 (c0 ()) @@ at() }
+  | LPAR MODULE module_fields RPAR { $3 (empty_context ()) @@ at() }
 ;
 
 
@@ -340,11 +340,13 @@ cmd :
   | modul { Define $1 @@ at() }
   | LPAR ASSERTINVALID modul TEXT RPAR { AssertInvalid ($3, $4) @@ at() }
   | LPAR INVOKE TEXT expr_list RPAR
-    { Invoke ($3, $4 (c0 ())) @@ at() }
+    { Invoke ($3, $4 (empty_context ())) @@ at() }
   | LPAR ASSERTEQ LPAR INVOKE TEXT expr_list RPAR expr RPAR
-    { AssertEq ($5, $6 (c0 ()), $8 (c0 ())) @@ at() }
+    { AssertEq ($5, $6 (empty_context ()), $8 (empty_context ())) @@ at() }
   | LPAR ASSERTTRAP LPAR INVOKE TEXT expr_list RPAR TEXT RPAR
-    { AssertTrap ($5, $6 (c0 ()), $8) @@ at() }
+    { AssertTrap ($5, $6 (empty_context ()), $8) @@ at() }
+  | LPAR ASSERTHEAPEQ INT TEXT RPAR
+    { AssertHeapEq (int_of_string $3, $4) @@ at() }
 ;
 cmd_list :
   | /* empty */ { [] }
--- a/ml-proto/src/host/script.ml
+++ b/ml-proto/src/host/script.ml
@@ -134,6 +134,8 @@ and char_at_memory_offset m base offset =
     | Int32 i -> Int32.to_int i
     | _ -> assert false
   in
+  assert (code >= 0);
+  assert (code <= 255);
   Char.chr code
 
 and memory_as_string m offset count =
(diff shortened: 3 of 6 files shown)


ml-proto/src/script.ml
```diff
@@ -71,6 +75,42 @@ let run_command cmd =
       Error.error cmd.at "assertion failed"
     end
 
+  | AssertHeapEq (offset, expected) ->
```
rossberg-chromium: Instead of checking each character individually, how about having a loop that reads the actual bytes into a `bytes` buffer of suitable length, and then just compare `Bytes.to_string buf = expected`. That way, you don't need to implement a second loop just for error printing.

ml-proto/src/script.ml
```diff
@@ -71,6 +75,42 @@ let run_command cmd =
...
+    let m = match !current_module with
+      | Some m -> m
+      | None -> Error.error cmd.at "no module defined to examine heap"
+    in
+    let mem = Eval.get_module_memory m in
+    let count = String.length expected in
+
```
rossberg-chromium: Style nit: Ocaml code does not typically use empty lines in places like this (similarly below)

ml-proto/src/script.ml
```diff
@@ -71,6 +75,42 @@ let run_command cmd =
...
+    for i=0 to (count-1) do
```
rossberg-chromium: Style nit: put spaces around binops; also, redundant parens

ml-proto/src/script.ml
```diff
@@ -71,6 +75,42 @@ let run_command cmd =
...
+      let expected_ch = Char.code expected.[i] in
+      let actual_ch = match Memory.load mem i UInt8Mem Int32Type with
```
rossberg-chromium: You probably want to make sure that the index is within bounds (either catch the Bounds exception or by adding a `size` function to the `memory` interface that you can compare against).

ml-proto/src/script.ml
```diff
@@ -71,6 +75,42 @@ let run_command cmd =
...
+        | Int32 i -> Int32.to_int i
+        | _ -> Error.error cmd.at "memory load returned wrong type"
```
rossberg-chromium: This would be an internal error in (the memory module of) the interpreter, not an error in the program, so the `Error` exception seems inappropriate. One idiomatic way of indicating such (hopefully) unreachable code in Ocaml is `assert false`.

ml-proto/src/script.ml
```diff
@@ -71,6 +75,42 @@ let run_command cmd =
...
+      in
+
+      if expected_ch <> actual_ch then begin
+        print_string "Result: \"";
+        
+        for j=0 to (count-1) do
```
rossberg-chromium: Nit: spaces

ml-proto/src/script.ml
```diff
@@ -71,6 +75,42 @@ let run_command cmd =
...
+          let ch = match Memory.load mem j UInt8Mem Int32Type with
+            | Int32 i -> Char.chr (Int32.to_int i)
+            | _ -> Error.error cmd.at "memory load returned wrong type"
```
rossberg-chromium: Same here.

ml-proto/src/eval.mli
```diff
@@ -9,3 +9,4 @@ val init : Ast.modul -> module_instance
 val invoke : module_instance -> string -> value list -> value list
   (* raise Error.Error *)
 val eval : Ast.expr -> value (* raise Error.Error *)
+val get_module_memory : module_instance -> Memory.t
```
rossberg-chromium: Nit: don't use `get` prefixes in functional programming.

ml-proto/src/parser.mly
```diff
@@ -51,7 +51,7 @@ type context =
   {funcs : space; globals : space; locals : space; labels : int VarMap.t}
 
 let empty () = {map = VarMap.empty; count = 0}
-let c0 () =
+let emptyContext () =
```
rossberg-chromium: Nit: empty_context (CamlCasing is only used for uppercase names in Ocaml, i.e., constructors and modules)

kg:
Rebased and fixed based on feedback

kg:
@jfbastien proposes that instead of doing this we just add a way to print a heap range to stdout. I'd be happy with that. We'd need to make sure any other wasm interpreter output goes to stderr, though. the ```runtests.py``` harness already has a mechanism for dumping interpreter output to a file and asserting it against an ```expected-output/foo.wasm.log``` file.

A host export that can be imported and called (printf?) would also be great, but I don't know if we're far enough along to do that yet?

Larger tests are going to need to assert against larger sets of values/heap ranges and string literals in the test are pretty awkward. See https://github.com/WebAssembly/ilwasm/blob/master/third_party/tests/Gradient.cs for a stripped down example of such a scenario. Given this PR it generates:

```
(assert_eq (invoke "generateGradient" (i32.const 0)  ) (i32.const 211) )
(assert_heap_eq 0 "\00\00\02\00\00\00\00\00\00\00\00\00\08\00\08\00\18\00\00\00\00\00\00 \00\00@\00\00`\00\00\80\00\00\A0\00\00\C0\00\00\E0 \00\00 \00  \00@ \00` \00\80 \00\A0 \00\C0 \00\E0@\00\00@\00 @\00@@\00`@\00\80@\00\A0@\00\C0@\00\E0`\00\00`\00 `\00@`\00``\00\80`\00\A0`\00\C0`\00\E0\80\00\00\80\00 \80\00@\80\00`\80\00\80\80\00\A0\80\00\C0\80\00\E0\A0\00\00\A0\00 \A0\00@\A0\00`\A0\00\80\A0\00\A0\A0\00\C0\A0\00\E0\C0\00\00\C0\00 \C0\00@\C0\00`\C0\00\80\C0\00\A0\C0\00\C0\C0\00\E0\E0\00\00\E0\00 \E0\00@\E0\00`\E0\00\80\E0\00\A0\E0\00\C0\E0\00\E0")
```

lukewagner:
FWIW, big longs strings of heap dumps like you mention won't be much fun to debug (whether it's in `assert_heap_eq` or asserted text output).  In asm.js tests, we do a bunch of heap testing, but on specific heap locations so we're never asserting against long strings.  This makes it a lot easier to understand failures.  We of course should have big compiled workloads that walk all over the heap (for example, we run bullet in the asm.js unit tests and assert the final positions of boxes), but as already discussed, that's probably separate from our unit test suite.

kg:
That's only if the heap dumps are raw binary goo. This also addresses stdout testing (my first test was in fact a strcat test)

lukewagner:
That's a good point.  In that case, I can see both the assert command and a new stdio builtin to print strings being useful.

jfbastien:
So should we merge this, and agree to use it only for small things, versus using "golden" output files for longer output validation?

kg:
How would you feel about an ```assert_heap_eq_file``` referencing an expected output file living somewhere in the test repo? Would that be better?

kg:
er, wtf. why did github close this?

kg:
This isn't working anymore because of some strange internals issue, so it's not ready to merge yet. Have to change my test cases to conform with all the opcode name changes before I can dig in.

kg:
Can we come to an agreement on this? I'm tired of rebasing it over and over.

kg:
It seems to be impossible to rebase this now, though maybe I can figure it out with my graphical git tools at home...

kg:
Yeah, the extensive changes to the codebase (file moves, deletes, renames, splits, merges) have made this impossible to continue to rebase.

PR #7, opened 2015-08-17
arithmetic: Add Ctz function

eminemix:
Signed-off-by: Valentin Ilie <valentin.ilie@intel.com>

PR #50, opened 2015-09-07
implementation and tests for clz, ctz, and popcnt

naturaltransformation:
A short, illustrative implementation of integer bit operations, unoptimized since I assume the spec isn't prescribing particular optimizations. Test cases can benefit from hex integer literals though.

### DIFF
--- a/ml-proto/src/spec/arithmetic.ml
+++ b/ml-proto/src/spec/arithmetic.ml
@@ -33,6 +33,7 @@ sig
   val shift_right : t -> int -> t
   val shift_right_logical : t -> int -> t
   val to_int : t -> int
+  val of_int : int -> t
   val of_int32 : int32 -> t
   val of_int64 : int64 -> t
   val to_float : t -> float
@@ -42,6 +43,8 @@ sig
   val of_big_int_u : Big_int.big_int -> t
   val to_value : t -> value
   val of_value : int -> value -> t
+  val zero : t
+  val one : t
 end
 
 let to_big_int_u_for size to_big_int i =
@@ -118,11 +121,35 @@ struct
   let unsigned big_op i j = big_op (Int.to_big_int_u i) (Int.to_big_int_u j)
 
   let unop op =
+    let open Int in 
     let f = match op with
-      | Clz -> fun i -> i  (* TODO *)
-      | Ctz -> fun i -> i  (* TODO *)
-      | Popcnt -> fun i -> i  (* TODO *)
-    in fun v -> Int.to_value (f (Int.of_value 1 v))
+      | Clz -> 
+        let rec loop acc n =
+          if n = zero then 
+            size 
+          else if logand n (shift_left one (size - 1)) <> zero then 
+            acc 
+          else 
+            loop (1 + acc) (shift_left n 1) 
+        in loop 0
+      | Ctz -> 
+        let rec loop acc n =
+          if n = zero then 
+            size
+          else if logand n one = one then 
+            acc
+          else 
+            loop (1 + acc) (shift_right_logical n 1) 
+        in loop 0
+      | Popcnt -> 
+        let rec loop acc i n =
+          if n = zero then 
+            acc
+          else 
+            let acc' = if logand n one = one then acc + 1 else acc in
+            loop acc' (i - 1) (shift_right_logical n 1)
+        in loop 0 size
+    in fun v -> to_value (of_int (f (of_value 1 v)))
 
   let binop op =
     let f = match op with
--- a/ml-proto/test/int32.wasm
+++ b/ml-proto/test/int32.wasm
@@ -1,0 +1,37 @@
+(; Int arith operations ;)
+
+(module
+  (func $clz (param $x i32) (result i32)
+    (i32.clz (get_local $x))
+  )
+
+  (func $ctz (param $x i32) (result i32)
+    (i32.ctz (get_local $x))
+  )
+
+  (func $popcnt (param $x i32) (result i32)
+    (i32.popcnt (get_local $x))
+  )
+    
+  (export "clz" $clz)
+  (export "ctz" $ctz)
+  (export "popcnt" $popcnt) 
+)
+
+(assert_eq (invoke "clz" (i32.const -1)) (i32.const 0)) ;; 0xFFFFFFFF
+(assert_eq (invoke "clz" (i32.const 0)) (i32.const 32))
+(assert_eq (invoke "clz" (i32.const 32768)) (i32.const 16)) ;; 0x00008000 
+(assert_eq (invoke "clz" (i32.const 255)) (i32.const 24)) ;; 0xFF
+(assert_eq (invoke "clz" (i32.const -2147483648)) (i32.const 0)) ;; 0x80000000
+(assert_eq (invoke "clz" (i32.const 1)) (i32.const 31))
+(assert_eq (invoke "clz" (i32.const 2)) (i32.const 30)) 
+
+(assert_eq (invoke "ctz" (i32.const -1)) (i32.const 0))
+(assert_eq (invoke "ctz" (i32.const 0)) (i32.const 32))
+(assert_eq (invoke "ctz" (i32.const 32768)) (i32.const 15)) ;; 0x00008000
+(assert_eq (invoke "ctz" (i32.const 65536)) (i32.const 16)) ;; 0x00010000
+(assert_eq (invoke "ctz" (i32.const -2147483648)) (i32.const 31)) ;; 0x80000000
+
+(assert_eq (invoke "popcnt" (i32.const -1)) (i32.const 32))
+(assert_eq (invoke "popcnt" (i32.const 0)) (i32.const 0))
+(assert_eq (invoke "popcnt" (i32.const 32768)) (i32.const 1)) ;; 0x00008000


jfbastien:
(drive-by confirmation that George is part of the [W3C CG](https://www.w3.org/community/webassembly/participants) under his company, thanks for the contribution!)

wcarr:
This contains code from the closed #7 that was dropped due to not being from a CG participant.  Just to avoid any worries about that, this is to confirm that code was also from Intel and one of our WebAssembly CG reps has contributed it this time (I'm Intel's W3C AC rep).

ml-proto/test/int32.wasm
```diff
@@ -0,0 +1,37 @@
...
+    (i32.clz (get_local $x))
+  )
+
+  (func $ctz32 (param $x i32) (result i32)
+    (i32.ctz (get_local $x))
+  )
+
+  (func $pcnt32 (param $x i32) (result i32)
+    (i32.popcnt (get_local $x))
+  )
```
sunfishcode: It'd be nice to keep the function name identical to the opcode name.

ml-proto/src/arithmetic.ml
```diff
@@ -119,10 +123,36 @@ struct
...
   let unop op =
     let f = match op with
-      | Clz -> fun i -> i  (* TODO *)
-      | Ctz -> fun i -> i  (* TODO *)
-      | Popcnt -> fun i -> i  (* TODO *)
-    in fun v -> Int.to_value (f (Int.of_value 1 v))
+      | Clz -> fun (v : Int.t) -> 
+                 let rec helper n acc =
+                   match n with 
+                     | n when n = Int.zero -> Int.size
```
sunfishcode: Is this line equivalent to `| Int.zero -> Int.size`?

naturaltransformation: Yeah, I'd like that, but unfortunately, Int.zero is a value and not a constructor so it can't be used in the pattern match directly, only as a guard. In my earlier iteration, I was thinking of dispensing with the match altogether but it does seem to be clearer in the later cases.

ml-proto/src/spec/arithmetic.ml
```diff
@@ -119,10 +123,36 @@ struct
...
+                   match n with 
+                     | n when n = Int.zero -> Int.size
+                     | n when Int.logand n Int.one = Int.one -> acc
+                     | _ -> helper (Int.shift_right n 1) (1+acc)
+                 in helper v 0
+      | Popcnt -> fun v ->
+                    let rec helper n acc i =
+                      match (n, i) with 
+                       | (n,_) when n = Int.zero -> acc
+                       | (_,0) -> acc
```
sunfishcode: Is this line ever reached? It's not clear that this line will ever match if the first line doesn't.

ml-proto/src/spec/arithmetic.ml
```diff
@@ -119,10 +123,36 @@ struct
...
+                       | _ -> 
+                           let n' = Int.shift_right n 1
```
sunfishcode: I'm still very much an OCaml beginner, but I'd find this code a little clearer if you put the computation of n' after the computation of acc', since it uses the original n. And if you do that, then you don't need to use the apostrophes; you can just reuse the names "acc" and "n" (a style which at least one OCaml tutorial I read encourages, but I'm happy to be corrected).

sunfishcode:
This code lgtm, with a few comments above. Eventually it'll be desirable to factor this code into functions outside of arithmetic.ml, but I think this is a good first step. Also, it'd be good for someone with more OCaml experience than I to review this.

ml-proto/src/spec/arithmetic.ml
naturaltransformation: The first line matches n=Int.zero and the second when i=0. It would have been clearer and more succinct if this could be have turned into an OR pattern but alas OR patterns are incompatible with guards. OCaml will helpfully identify some redundant or non-exhaustive pattern cases.

ml-proto/src/spec/arithmetic.ml
naturaltransformation: The variable shadowing you suggest here seems fine. In the more general case, however, the ML family's liberal shadowing can easily get one into trouble. For example, shadowing of exception names can result in some pretty non-intuitive behaviors.

ml-proto/src/spec/arithmetic.ml
sunfishcode: From my reading of the code, when i=0, n must have been shifted down to zero too, which would mean this case isn't ever taken.

naturaltransformation: Thanks! Actually, that is a great catch. I was using the wrong shift operator (it should be shift_right_logical instead of shift_right, now fixed). That case happened to mask the issue by handling negative integer inputs.

ml-proto/src/spec/arithmetic.ml
```diff
@@ -42,6 +43,9 @@ sig
   val of_big_int_u : Big_int.big_int -> t
   val to_value : t -> value
   val of_value : int -> value -> t
+  val compare : t -> t -> int
```
rossberg-chromium: Compare is polymorphic in Ocaml, you don't need to parameterise over it.

ml-proto/src/spec/arithmetic.ml
```diff
@@ -119,10 +123,30 @@ struct
 
   let unop op =
     let f = match op with
-      | Clz -> fun i -> i  (* TODO *)
-      | Ctz -> fun i -> i  (* TODO *)
-      | Popcnt -> fun i -> i  (* TODO *)
-    in fun v -> Int.to_value (f (Int.of_value 1 v))
+      | Clz -> let rec loop acc n =
```
rossberg-chromium: Nit: break line and avoid over-indentation (same below).

ml-proto/src/spec/arithmetic.ml
```diff
@@ -119,10 +123,30 @@ struct
...
+                 match n with
```
rossberg-chromium: Nit: If all your cases are just `when` conditions then there is no point in using `match` instead of a simple `if` cascade.

ml-proto/src/spec/arithmetic.ml
```diff
@@ -119,10 +123,30 @@ struct
...
   let unop op =
     let f = match op with
-      | Clz -> fun i -> i  (* TODO *)
-      | Ctz -> fun i -> i  (* TODO *)
-      | Popcnt -> fun i -> i  (* TODO *)
-    in fun v -> Int.to_value (f (Int.of_value 1 v))
+      | Clz -> let rec loop acc n =
+                 match n with 
+                     | n when n = Int.zero -> Int.size 
+                     | n when Int.logand n (Int.shift_left Int.one (Int.size-1))
```
rossberg-chromium: Nit: spacing around `-`

ml-proto/src/spec/arithmetic.ml
rossberg-chromium: Also, how about adding a `let open Int in` here to avoid all the repetition? (similarly for the other cases)

ml-proto/src/spec/arithmetic.ml
```diff
@@ -119,10 +123,30 @@ struct
...
+      | Ctz -> let rec loop acc n =
+                 match n with 
+                     | n when n = Int.zero -> Int.size
+                     | n when Int.logand n Int.one = Int.one -> acc
+                     | _ -> loop (1+acc) (Int.shift_right_logical n 1) 
+               in loop 0
+      | Popcnt -> let rec loop acc i n =
+                    if n = Int.zero then acc
+                    else let acc = 
+                           if Int.logand n Int.one = Int.one
```
rossberg-chromium: Nit: odd indentation. Also, the shadowing of `acc` is confusing. I'd write this as
```
let open Int in
let rec loop acc i n =
  if n = zero then
    acc
  else
    let acc' = if logand n one = one then acc + 1 else acc in
    loop acc' (i - 1) (shift_right_logical n 1)
```

ml-proto/test/int32.wasm
rossberg-chromium: On 16 September 2015 at 01:36, Dan Gohman <notifications@github.com> wrote:

> In ml-proto/test/int32.wasm
> <https://github.com/WebAssembly/spec/pull/50#discussion_r39579393>:
>
> > @@ -0,0 +1,37 @@
> > +(; Int arith operations ;)
> > +
> > +(module
> > +  (func $clz32 (param $x i32) (result i32)
> > +    (i32.clz (get_local $x))
> > +  )
> > +
> > +  (func $ctz32 (param $x i32) (result i32)
> > +    (i32.ctz (get_local $x))
> > +  )
> > +
> > +  (func $pcnt32 (param $x i32) (result i32)
> > +    (i32.popcnt (get_local $x))
> > +  )
>
> It'd be nice to keep the function name identical to the opcode name.
>

Btw, despite it being what Intel picked on a bad day, `popcnt` really is an
odd name out. It is both inconsistent with the other count operators, and
it sounds like some kind of stack operator.

Would anybody mind if we changed it? For example, how about `cnz` (count
non-zeros), which would be in line with the other count operators. (If you
say it's a worthwhile idea I can create an issue or PR.)

ml-proto/src/spec/arithmetic.ml
naturaltransformation: Ok. But question though, does val compare : 'a -> 'a -> int vs. val compare: t -> t -> int make a difference to specialization either way?

ml-proto/src/spec/arithmetic.ml
naturaltransformation: Does Taureg mode give this indentation or is there another OCaml mode I should be using?

ml-proto/src/spec/arithmetic.ml
rossberg-chromium: No, they are the same (would be rather terrible if they weren't :) ).

ml-proto/src/spec/arithmetic.ml
rossberg-chromium: I assume it does, but then, I don't use any mode. :)

ml-proto/test/int32.wasm
naturaltransformation: I may be wrong, but IIRC, the “population count” terminology long precedes Intel. It is used in GHC base package, GCC, Sparc, Cray, and CDC (the latter ones long before Intel). Even the Mark I had it but called it Sideways Add.

naturaltransformation: On UltraSparc, it is known as POPC if that is any better.

naturaltransformation:
Any further feedback on this one? Thanks!

rossberg-chromium:
lgtm modulo nit

sunfishcode:
Merging with lgtm and nit addressed. Any further feedback can be addressed in followup patches. Thanks!

PR #51, opened 2015-09-08
Fixed i32.wrap/i64 being parsed as i64.wrap/i64

AndrewScheidecker:
It was parsing i64.wrap/i64, but I believe that's an accidental divergence from the AstSemantics doc, so I changed it to i32.wrap/i64.

I also added tests for i64.extend_s/i32, i64.extend_u/i32, i32.wrap/i64, i32.trunc_s/f32, i32.trunc_u/f32, i64.trunc_s/f32, and i64.trunc_u/f32.

I had to use negative constants for some of the tests to work around an apparent bug that const.iXX doesn't handle values larger than the largest signed value for that integer size.


rossberg-chromium:
lgtm

jfbastien:
@AndrewScheidecker can you resolve conflicts so we can merge the issue?

kg:
LGTM. Please merge, I'm blocked on this

sunfishcode:
This has lgtms and is now mergable. Merging.

PR #54, opened 2015-09-10
Add import section to modules

lukewagner:
This patch builds on #53 and implements imports of builtin functions as discussed in #47.  The main difference is that, instead of injecting imports into the index space of functions (such that, after importing a function you call it with `(call ...)`, the patch creates a separate index space for imports that are called through a new `(call_import ...)`.  The reason for this is that I think it makes everything (codegen, spec, impl) simpler.


ml-proto/README.md
```diff
@@ -172,13 +168,13 @@ expr:
   ( if <expr> <expr> )                     ;; = (if <expr> <expr> (nop))
   ( loop <expr>* )                         ;; = (loop (block <expr>*))
   ( label <name>? <expr>* )                ;; = (label (block <expr>*))
-  ( break <var> <expr>* )
+  ( break <var> <expr> )
```
binji: shouldn't this just be ( break <var> ) now? I believe the expression list was used as a return value to the label.

ml-proto/README.md
```diff
@@ -233,7 +230,7 @@ script: <cmd>*
 cmd:
   <module>                                       ;; define, validate, and initialize module
   ( invoke <name> <expr>* )                      ;; invoke export and print result
-  ( asserteq (invoke <name> <expr>* ) <expr>* )  ;; assert expected results of invocation
+  ( asserteq (invoke <name> <expr>* ) <expr> )   ;; assert expected results of invocation
```
binji: While you're here, these have been renamed to assert_eq and assert_invalid

ml-proto/README.md
lukewagner: You're right there a bug here, but, because labels still return a value (this patches changes that from `value list` to `value option`) this should be `(break <var> <expr>?)`.

ml-proto/README.md
lukewagner: Good point, done.

kg:
LGTM

lukewagner:
Thanks for reviewing.  I'm inclined to wait for @rossberg-chromium to return before merging, unless he was going to be offline for a while.  Anyone know?

lukewagner:
@rossberg-chromium Updated

jfbastien:
Could you do the corresponding change in AstSemantics before merging? The idea sounds fine to me, I just want to make sure the more-oft-observed design repo gets a say in this!

lukewagner:
@jfbastien I was about to say that this PR simply implements what's already in `design` (most of the interesting stuff we're discussing, like the overloading, is in `host/`, not `spec/`), but I forgot that I did add `call_import` which should indeed be reflected in AstSemantics.md#call.

rossberg-chromium:
LGTM

PR #55, opened 2015-09-11
Update load/store to match design changes

lukewagner:
This patch updates ml-proto to match the changes in #341.  The only remaining difference is that, for all opcode names, not just load/store, ml-proto uses the "i32" form vs. "int32".  (I could see this difference resolved in either spec or design, so I'll leave that for a separate issue.)


lukewagner:
@rossberg-chromium Updated with `align = _` and rebased over single-return-type.

rossberg-chromium:
LGTM

jfbastien:
(updated description to https://github.com/WebAssembly/design/pull/341, not this repo's non-existent #341)

PR #57, opened 2015-09-11
initial tests for globals

marianoguerra:
this are some simple tests to check that globals are set and return the set value both from inside the module and from outside and that setting two different globals doesn't affect each other.

this is not in sync with the spec which says that store should return the stored value, will look into it now.

### DIFF
--- a/ml-proto/src/spec/check.ml
+++ b/ml-proto/src/spec/check.ml
@@ -186,7 +186,7 @@ let rec check_expr c ts e =
 
   | StoreGlobal (x, e1) ->
     check_expr c [global c x] e1;
-    check_type [] ts e.at
+    check_type [global c x] ts e.at
 
   | Load (memop, e1) ->
     check_memop memop e.at;
--- a/ml-proto/src/spec/check.ml
+++ b/ml-proto/src/spec/check.ml
@@ -179,7 +179,7 @@ let rec check_expr c ts e =
 
   | SetLocal (x, e1) ->
     check_expr c [local c x] e1;
-    check_type [] ts e.at
+    check_type [local c x] ts e.at
 
   | LoadGlobal x ->
     check_type [global c x] ts e.at
--- a/ml-proto/src/spec/check.ml
+++ b/ml-proto/src/spec/check.ml
@@ -197,7 +197,7 @@ let rec check_expr c ts e =
     check_memop memop e.at;
     check_expr c [Int32Type] e1;
     check_expr c [memop.ty] e2;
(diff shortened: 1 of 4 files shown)


lukewagner:
Thanks!  I think you also need to add a rule to check.ml to allow expressions like assigning a `set_global` to a local to succeed.  Just a fair warning, we're seriously considering removing globals from the MVP (and so ml-proto) in #154.  However the analogous change you've made here also applies to `set_local` and `store`.

marianoguerra:
* added the change in check for store_global
* added test to check that I can use the result of store_global
* added the ability to return the value being set on set_local
* added test to check that I can use the result of set_local

regarding removing globals, no problem, I'm using this as a training to understand the project and the spec so no problem for me throwing this later :). Just out of curiosity, will globals be removed for good or there are plans to reintroduce them later? I think globals are easily "shimed" with module memory so maybe they are not that primitive...

lukewagner:
@marianoguerra Globals would just be removed from the MVP.  See the new #344 (which, when merged, will be followed by a PR in this repo) for more explanation.

ml-proto/src/spec/check.ml
```diff
@@ -179,14 +179,14 @@ let rec check_expr c ts e =
 
   | SetLocal (x, e1) ->
     check_expr c [local c x] e1;
-    check_type [] ts e.at
+    check_type [local c x] ts e.at
```
lukewagner: Here and below, it might be nice to factor out via `let t1 = local c x in` to avoid duplicating the expression.

marianoguerra: I will, just need to read an introduction to ocaml somewhere, any good recommendation? (I already code in clojure, erlang and scheme so it can be a high level introduction)

rossberg-chromium: FWIW, I actually find it more readable (since shorter) as is. :)

lukewagner: Hehe, fair enough.  I was on the edge anyhow for such a simple duplicated expr.

jfbastien:
Should we avoid merging this, based on https://github.com/WebAssembly/design/pull/344 removing globals from MVP?

marianoguerra:
@jfbastien I can remove globals and merge the other changes if it's ok for you.

jfbastien:
Sounds good.

lukewagner:
If you update this PR to remove globals, I think we can merge it.

lukewagner:
Created #106 which subsumes this PR.

PR #58, opened 2015-09-11
add some tests for call

marianoguerra:
the idea is to test call and call_indirect, but since I can't get to call call_indirect without a syntax error I start with simple call tests.

notice that at the bottom there's a commented test that fails.

### DIFF
--- a/ml-proto/test/calls.wasm
+++ b/ml-proto/test/calls.wasm
@@ -1,0 +1,37 @@
+(module
+ (func $one
+    (i32.const 1))
+
+ (func $two
+    (i32.const 2))
+
+ (func $add (param i32) (param i32) (result i32)
+    (i32.add (get_local 0) (get_local 1)))
+
+ (func $callone (call $one))
+ (func $calltwo (call $two))
+ (func $calladd (param i32) (param i32) (result i32)
+  (call $add (get_local 0) (get_local 1)))
+
+ (export "one" $one)
+ (export "two" $two)
+ (export "add" $add)
+ (export "callone" $callone)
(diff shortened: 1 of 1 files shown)


ml-proto/test/calls.wasm
```diff
@@ -0,0 +1,37 @@
...
+(assert_eq (invoke "add" (i32.const 1) (i32.const 2)) (i32.const 3))
+
+(assert_eq (invoke "callone") (i32.const 1))
+(assert_eq (invoke "calltwo") (i32.const 2))
+(assert_eq (invoke "calladd" (i32.const 1) (i32.const 2))
+           (i32.const 3))
+
+(; the following test fails
+(assert_eq (invoke "calladd" (invoke "callone") (invoke "calltwo"))
+           (i32.const 3))
```
lukewagner: The reason this fails is that the first argument to `assert_eq` isn't an arbitrary expression, it must be a fixed form `(assert_eq (invoke ...) ...)`.  This relates to trying to limit how much expression evaluation happens in the scripting language, compared to within real wasm code.

marianoguerra: should I create a second module that exports functions called "test{feature}" and then on outer asserts call those?

lukewagner: I don't quite follow what you're suggesting.  In general, we try to do most of the computation inside wasm and then just assert the result (or, when #54 merges, printing output and asserting output match with a .log file in expected-output).

lukewagner:
Tentatively closing since patch is rather out of date and calls now have a lot more tests.  Feel free to reopen.

PR #61, opened 2015-09-14
Parse hex literals

binji:


ml-proto/src/host/lexer.mll
```diff
@@ -97,7 +97,8 @@ let escape = ['n''t''\\''\'''\"']
 let character = [^'"''\\''\n'] | '\\'escape | '\\'hexdigit hexdigit
 
 let num = ('+' | '-')? digit+
-let int = num
+let hexnum = ('+' | '-')? "0x" digit+
```
jfbastien: Shouldn't this be `hexdigit`?

binji: Whoops, yes. Fixed!

ml-proto/test/hexnum.wasm
```diff
@@ -0,0 +1,39 @@
+(module
+  (func $test32 (result i32)
+    (return (i32.const 0x0badd00d))
```
jfbastien: Also test uppercase hex?

PR #62, opened 2015-09-14
Update memory/segment definition in README

binji:


ml-proto/README.md
```diff
@@ -187,22 +187,22 @@ case:
...
+global:  ( global <type>* ) | ( global <name> <type> )
+table:   ( table <var>* )
+memory:  ( memory <int> <int>? <segment>* )
+segment: ( segment <int> "<char>*" )
 ```
 
 Here, productions marked with respective comments are abbreviation forms for equivalent expansions.
 
-The data string is used to initialise the lower end of the memory. It is an ASCII string, that can have the usual escape sequences, or hex escapes of the form `\xx` to denote a single byte.
+The segment string is used to initialise the lower end of the memory. It is an ASCII string, that can have the usual escape sequences, or hex escapes of the form `\xx` to denote a single byte.
```
jfbastien: What's the "usual" escape sequences? I'd like to have something a bit more precise!
Is there a pre-existing standard we can use? Presumably Unicode would work.
How are quotes escaped?

I guess we don't define `char` at all here, which is the true problem.

binji:
How's this?

jfbastien:
I'd rather define `char` instead of special-casing segment. I'd define it as something that's already standardized, so we don't have to invent something new.

binji:
Perhaps that is worth doing in a separate PR. This was meant to fix the broken definition of the memory expression that was not updated when the code was.

jfbastien:
I'd be OK with that if the text were only "The segment string is used to initialize the memory at the given offset." (with typo fix in there!), and an associated issue or PR to fix `char`.

rossberg-chromium:
I agree with @jfbastien. Better define `char` than describing it in prose. In this or a separate PR.

binji:
Added issue #64, and removed the prose definition of char.

jfbastien:
lgtm, thanks! Merging since @rossberg-chromium lgtm'd as well.

PR #66, opened 2015-09-15
Refactor the Float32 and Float64 modules into a Float functor.

sunfishcode:
@rossberg-chromium Thanks for the help! I made the "with type t = ..." change you suggested, and the naming changes.

I also prefixed the given/* modules with "Basic" to anticipate disambiguating them from arithmetic.ml's versions, which will need to add conversions to/from value, unless we want value to live in given too. But I'm open to better naming schemes.

However, I'm still getting errors I don't understand:

```
ocamlopt -c -I host -I given -I spec  given/basicfloat.ml
File "given/basicfloat.ml", line 149, characters 32-51:
Error: Signature mismatch:
       ...
       Type declarations do not match:
         type t = Basicint.BasicInt32.t
       is not included in
         type t = Float32Traits.t
       File "given/basicfloat.ml", line 62, characters 68-85:
         Expected declaration
Makefile:72: recipe for target 'given/basicfloat.cmx' failed
```

Do you know what this error message means? Thanks.


rossberg-chromium:
When you annotate a module with a signature (aka module type) then that is the signature the module gets coerced to. In particular, if the signature specifies some types abstractly (as the types `t` and `bits` in `FLOAT_TRAITS`) then the identity of these types is hidden, and consequently, they are not compatible to their underlying representation at use sites.

The fix in this case is to either not annotate the signature at the definition of the Float Traits (module typing is structural, so you do not need to write it), or to explicitly make the types transparent, by saying e.g.
```
module Float32Traits : FLOAT_TRAITS with type t = int32 = ...
```
You probably need to do the same for type `bits` and the `BASIC_FLOAT` signature in the definition of the `Float` functor.

The applications of that functor OTOH can be simplified to
```
module BasicFloat32 = Float (Float32Traits) (Basicint.BasicInt32)
module BasicFloat64 = Float (Float64Traits) (Basicint.BasicInt64)
```

I'll think about naming some more.

sunfishcode:
Closing this PR for now; the floating point code is evolving; we can look at refactoring it more once it's settled down more.

PR #67, opened 2015-09-15
Rename 'neq' to 'ne'.

sunfishcode:
This is what the design repo calls it, and it's convenient to have all
the comparison operators be the same number of characters (ignoring the
_s or _u).


rossberg-chromium:
lgtm

sunfishcode:
Merging with lgtm and with this being what's in the design repo.

PR #73, opened 2015-09-17
Fix a bug in the detection of end-of-file.

### LINKED ISSUE #72: Mystery syntax error on file suspiciously near 65536 bytes in size
My branch here:

https://github.com/WebAssembly/spec/tree/large-file-syntax-error

has a (work-in-progress) test in test/float64.wasm. It currently fails mysteriously with 
```
test/float64.wasm:524.130-524.130: syntax error
```

Line 524 is syntactically correct. Deleting seemingly any single the assert line from the file makes the syntax error go away. Suspiciously, the file size is just over 65536 bytes.

Anyone have any clues as to what could be going on here?

sunfishcode:
input returns 0 when it reaches the end of the file; other values may
indicate merely a short read. This fixes #72.

### DIFF
--- a/ml-proto/src/host/main.ml
+++ b/ml-proto/src/host/main.ml
@@ -12,7 +12,7 @@ let load file =
   let rec loop () =
     let len = input f buf 0 size in
     let source = Bytes.sub_string buf 0 len in
-    if len < size then source else source ^ loop ()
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
Oops, thanks for spotting. LGTM

PR #74, opened 2015-09-17
Add assert_fault script command

lukewagner:
This patch adds a command `(assert_fault (invoke "foo" ...args) "...fail regex")` that asserts a specified runtime failure.


rossberg-chromium:
lgtm modulo comment

PR #76, opened 2015-09-17
Rename assert_fault to assert_trap to match design repo terminology

lukewagner:
@sunfishcode Pointed out that the consistent term we use through the design repo for these runtime errors is [trap](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#abstract-syntax-tree-semantics), so this renames "fault" to "trap" in #74.


rossberg-chromium:
lgtm

sunfishcode:
I added an extra commit to do a few more obvious renames (which fixes the test failures). Merging with lgtm above.

PR #77, opened 2015-09-17
Avoid relying on correct rounding of mod_float.

sunfishcode:
mod_float is not always correctly rounded on all platforms. This patch uses a safer algorithm to check whether a number is even.


rossberg-chromium:
lgtm

PR #78, opened 2015-09-17
Fix the min and max computation for the case where the operands are equal

sunfishcode:
This fixes a thinko in the min and max functions. Tests for this will be added soon; they're in development on the [float-tests](https://github.com/WebAssembly/spec/tree/float-tests) branch.


rossberg-chromium:
lgtm

PR #79, opened 2015-09-18
Rename 'or' to 'ior', to match the design repo.

sunfishcode:
I don't have a strong preference either way; I just want spec and design to match here.


rossberg-chromium:
Can we change the design doc instead?  `ior` seems like a very non-standard name for no good reason.

sunfishcode:
Ok. I opened https://github.com/WebAssembly/design/pull/359 to change the design repo. We should merge exactly one of these two PRs.

sunfishcode:
https://github.com/WebAssembly/design/pull/359 was merged instead of this one.

Issue #70, opened 2015-09-16
Asserts for floating point tests


sunfishcode:
`assert_eq` uses ocaml's <> which returns false for `-0 <> 0` and true for `NaN <> x` ∀x. This is the usual behavior, but for our floating point unit tests we have unusual needs: we do actually need to distinguish between -0 and 0, test for specific NaNs sometimes, and test for any NaN sometimes. What do people think of adding the following?

 - `assert_eq_bits` - like `assert_eq`, but reinterpret-casts both operands as same-size integers first
 - `assert_nan` - has one operand and asserts that it is a NaN (of any kind)

rossberg-chromium:
I'd suggest simply changing the implementation of `assert_eq` to use the stricter equality operator. Or is there any reason to ever use IEEE equality in the tests we want to write?

As for NaN tests, it is easy to express that in WASM itself, so I don't think we need to extend the assertion language just for that either.

rossberg-chromium:
Btw, once your new float implementations land, `=` and `<>` on those types will automatically mean bitwise equality. So maybe it isn't even necessary to change anything here.

sunfishcode:
You're right; there is no need for the looser equality in our tests. It just feels weird to me to give `eq` a subtly different meaning in this context than it has in wasm itself.

As for NaN tests, it seems like it will be quite inconvenient to express any-NaN tests in terms of assert_eq regardless of whether it's the strict or loose version. The strict one tests for a specific bit pattern when we want to accept many, and the loose version requires we name the value so that we can reference it twice and form `x <> `. Or are you proposing we add NaN test functions inside the wasm modules themselves?

sunfishcode:
BTW, you're right that = and <> now automatically do bitwise equality with my float implementations. How does that work, given that the types are abstract?

sunfishcode:
Ah, I found [a blog post](https://syntaxexclamation.wordpress.com/2011/08/15/an-ocaml-hack-recover-abstraction-of-abstract-types/) which explains it.

jfbastien:
Agreed with what I think @rossberg-chromium is saying on NaNs: this should be done with a wasm function, e.g.: `assert(is_nan(v))`.

rossberg-chromium:
>
> BTW, you're right that = and <> now automatically do bitwise equality with
> my float implementations. How does that work, given that the types are
> abstract?
>

In Ocaml, equality is generic, and does the "right" thing depending on the
runtime type of the value.

[...]

PR #81, opened 2015-09-19
Add float conversion tests

AugustusHuang:


jfbastien:
Confirming that Xuxing Huang is a member of [the W3C CG](https://www.w3.org/community/webassembly/participants). Thanks!

rossberg-chromium:
lgtm

AugustusHuang:
Picked +/-0, +/- INT_MAX, +/-1 to test Int2Float, 0, 1, +/- MOST-POSITIVE-DOUBLE(SINGLE)-FLOAT, +/- LEAST-POSITIVE-DOUBLE(SINGLE)-FLOAT to test reinterpretFloat2Int, some of the results of reinterpretInt2Float will generate NaN so I didn't add them. All tests have one or more random cases :)

ml-proto/test/conversions.wasm
```diff
@@ -89,14 +146,113 @@
...
+(assert_eq (invoke "$f64_convert_s_i64" (i64.const 9223372036854775807)) (f64.const 9223372036854775807))
+(assert_eq (invoke "$f64_convert_s_i64" (i64.const -9223372036854775808)) (f64.const -9223372036854775808))
+(assert_eq (invoke "$f64_convert_s_i64" (i64.const 4669201609102990)) (f64.const 4669201609102990)) ;; Feigenbaum
+
+;; f32.convert_u/i32
+(assert_eq (invoke "$f32_convert_u_i32" (i32.const 1)) (f32.const 1.0))
+(assert_eq (invoke "$f32_convert_u_i32" (i32.const 0)) (f32.const 0.0))
+(assert_eq (invoke "$f32_convert_u_i32" (i32.const 2147483647)) (f32.const 2147483647))
+(assert_eq (invoke "$f32_convert_u_i32" (i32.const -2147483648)) (f32.const 2147483648))
+(assert_eq (invoke "$f32_convert_u_i32" (i32.const 305419896)) (f32.const 305419896)) ;; 0x12345678
```
sunfishcode: The wasm parser does recognize hexadecimal constants now, so you can write the constant directly now instead of writing it in decimal and having a comment :-).

sunfishcode:
This looks good!

One thing to watch out for: Constants such as 2147483647, 9223372036854775807, and 1234567890 are not exactly representable as f32 values; the wasm parser is silently rounding them. This is obviously not ideal. Eventually we should implement hexadecimal float literal parsing, and then we'll probably convert most of these tests to use it, but if you want to be tidy in the mean time, it'd be nice to avoid values that need rounding.

If you're interested in writing more conversion tests, here are some further ideas :-) :
 - Conversions with unsigned inputs will need to be tested with the maximum unsigned value of the type. For example, we'll need to test f32.convert_u/i32 against 0xffffffff and so on.
 - Conversions involving floating point will need to be tested for correct rounding. For example, the `trunc` conversions are defined to round toward zero, so we should make sure that a positive number rounds down even when the nearest integer is up, and a negative number rounds up even when the nearest integer is down, and so on. For conversions that round to nearest, we should test that a positive number is rounded down to nearest, a negative number is rounded up to nearest, and a number halfway between two integers is rounded to the even integer.
 - `trunc` conversions will need to test that values just outside the i32 range can be rounded into a valid i32 value.

We'll also eventually need to add -0, NaN, Infinity, and trapping cases, though as you noticed these aren't all convenient to do right now. I have some possible ideas for this in [this branch](https://github.com/WebAssembly/spec/tree/float-tests), though there's discussion ongoing about them in #70.

PR #82, opened 2015-09-21
Factor out i32 and i64 from arithmetic.ml; implement wasm-compatible semantics

sunfishcode:
This factors out the integer computation out of spec/arithmetic.ml and moves it into i32 and i64 modules under given, similar to the floating point types. It also is a first draft of design-compatible integer semantics, complete with trapping as appropriate.

This adds tests for all the plain integer operations, but the conversions are not yet tested. I am working on tests for them, though there are some floating point parsing issues that need to be resolved first, and [assert_eq_bits and assert_nan](https://github.com/WebAssembly/spec/issues/70) would be useful.


rossberg-chromium:
Looks good, although I haven't checked all the details.

I wonder if it's still adequate to put these modules into given/, considering that they implement rather Wasm-specific interfaces and semantics now. They are arguably in the same equivalence class as Memory.

Any reason you defunctorised & duplicated the existing code in arithmetic?

sunfishcode:
Moving them to spec makes sense. I added a commit for this now.

I defunctorized and duplicated the arithmetic code because while the existing code was generic over Int32 and Int64, it wasn't clear to me how to make a generic interface over I32_convert and I64_convert. I wouldn't necessarily object if someone wants to re-functorize everything from here, though I do appreciate the current property of the code that it's easy to follow, so I'd be interested to see how the factored code would look first.

rossberg-chromium:
lgtm, but somebody else might want to have a closer look at the tests, too.

lukewagner:
lgtm too!

sunfishcode:
Merging with lgtms above. If any bugs are found, please file new issues. If there are any language design concerns, please file new issues against the design repo, which this PR seeks to implement.

PR #83, opened 2015-09-21
Now using ocamlbuild for building. Rewrote Makefile, introduces _tags.

lindig:
.gitignore is simplified since now almost all build files reside in _build.
Using "git archive" to build a zip file in target "zip".

### DIFF
--- a/ml-proto/src/.gitignore
+++ b/ml-proto/src/.gitignore
@@ -1,16 +1,7 @@
-lexer.ml
-parser.ml
-parser.mli
-*.cmi
-*.cmo
-*.cmx
-*.exe
-*.o
-*.output
-*.depend
 *.native
 *.byte
 *.opt
 *.zip
 _build
 wasm
+unopt
--- a/ml-proto/src/Makefile
+++ b/ml-proto/src/Makefile
@@ -1,83 +1,44 @@
 #
-# (c) 2015 Andreas Rossberg
+# This Makfile assumes we can use ocamlfind to locate libraries. This
+# is the case in a typical OCaml environment that uses Opam. Use 
+#
+#	make check
+#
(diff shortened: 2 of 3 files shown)


lindig:
I'd like to suggest to use OCamlbuild for the build process (but still controlled by a `Makefile`). This results in a simpler build specification and a cleaner build structure as source and build files are better separated. The main change is in the `Makefile`, which could be integrated as `Makefile.ocb` to keep new and old versions for a while.

lindig:
I didn't realise that using ocamlbuild is already discussed in another pull request. This is somewhat simpler in comparison as it want to use ocamlbuild but does not pull in other tools like merlin or oasis.

lindig:
This Makefile uses ocamlbuild but no longer requires ocamlfind. Hence, it now works with the OCaml standard distribution.

rossberg-chromium:
Looks good, but for legal reasons, you will have to join the W3C WebAssembly community (https://www.w3.org/community/webassembly/).

@kg, should the test script be adapted to go through make now, so that it doesn't have to duplicate the build parameter logic?

kg:
> @kg, should the test script be adapted to go through make now, so that it doesn't have to duplicate the build parameter logic?

This would mean the only way to use this stuff on Windows would be to install Cygwin. But that seems to basically be the case already w/ocaml, so I guess I'm fine with it.

rossberg-chromium:
Is it very likely that somebody has Python but not some form of Make
installed on their Windows machine?

lindig:
I joined W3C and the WebAssembly Community: https://www.w3.org/users/83026.

rossberg-chromium:
LGTM. Test runner changes can be a follow-up.

PR #84, opened 2015-09-22
Got rid of Obj.magic use in memory module

### LINKED ISSUE #52: Memory bug on Windows x64
On my x64 Windows 10 box memory.wasm fails mysteriously:

```
-- Assert invoking...
Result: 42. : f64
Expect: 42. : f64
-- Error:
test/memory.wasm:151.1-151.45: assertion failed
```

Not sure what's going on here, don't know ocaml well enough to debug.

AndrewScheidecker:
Instead of doing magical casts of Bigarrays, it unpacks values into bytes before storing them, and packs them from bytes on load.

I don't know whether this is the best approach, but it fixes #52 and relies on less casting magic.


sunfishcode:
Do Bigarrays use host endianness? If so, this patch also corrects the endianness on big-endian hosts, since WebAssembly is uses little-endian ordering.

rossberg-chromium:
LGTM

ml-proto/src/spec/memory.ml
```diff
@@ -91,38 +76,56 @@ let address_of_value = function
...
+  Int32.logor (load8 mem a NX) (Int32.shift_left (load8 mem (a+1) ext) 8)
+
+let load32 mem a =
+  Int32.logor (load16 mem a NX) (Int32.shift_left (load16 mem (a+2) NX) 16)
+
+let load64 mem a =
+  Int64.logor (Int64.of_int32 (load32 mem a)) (Int64.shift_left (Int64.of_int32 (load32 mem (a+4))) 32)
+
+let store8 mem a bits =
+  !mem.{a} <- Int32.to_int (Int32.logand bits (Int32.of_int 255))
```
sunfishcode: Is this &255 necessary? I guess it doesn't hurt, but I was initially confused about what this code is doing and wasn't clear whether it's doing something that needs the high bits to be cleared.

AndrewScheidecker: AFAIK it's not necessary. I think there's some value in making it clear there's a truncation happening there, but I'm happy to remove it if you want.

sunfishcode: The syntax was indeed unfamiliar to me at first, so I do appreciate clarifying it. I'd suggest a comment, something like "store the least significant byte of `bits` at memory index `a`". With the &255, I was wondering whether it was perhaps doing a bitwise or of `bits` with memory such that it actually needed the high bits to be zero or something.

PR #85, opened 2015-09-24
Replaced explicit truncation on memory store with implicit truncation+comment

AndrewScheidecker:
As requested by @sunfishcode [here](https://github.com/WebAssembly/spec/pull/84#discussion_r40200166)

### DIFF
--- a/ml-proto/src/spec/memory.ml
+++ b/ml-proto/src/spec/memory.ml
@@ -91,7 +91,8 @@ let load64 mem a =
   Int64.logor (Int64.of_int32 (load32 mem a)) (Int64.shift_left (Int64.of_int32 (load32 mem (a+4))) 32)
 
 let store8 mem a bits =
-  !mem.{a} <- Int32.to_int (Int32.logand bits (Int32.of_int 255))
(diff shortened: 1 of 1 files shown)


ml-proto/src/spec/memory.ml
```diff
@@ -91,7 +91,8 @@ let load64 mem a =
   Int64.logor (Int64.of_int32 (load32 mem a)) (Int64.shift_left (Int64.of_int32 (load32 mem (a+4))) 32)
 
 let store8 mem a bits =
-  !mem.{a} <- Int32.to_int (Int32.logand bits (Int32.of_int 255))
+  (* Store "bits" at byte index a; this implicitly truncates all but the lowest 8 bits of "bits". *)
```
sunfishcode: I apologize for being pedantic, but "truncate" in wasm spec lingo means discarding least-significant digits of a value. How about "this stores just the low 8 bits of "bits", discarding the rest"?

AndrewScheidecker: Updated.

sunfishcode:
lgtm; thanks! Merging based on my own lgtm because this is a minor change that I don't expect will be controversial.

PR #86, opened 2015-09-25
Update load/store to match design repo changes

lukewagner:
Updating the spec repo to re-include ops re-added by [design/#366](https://github.com/WebAssembly/design/pull/366).

In the process, I applied the idea we discussed in the other issue of splitting `Load` into `Load`/`LoadExtend` and `Store` into `Store`/`StoreTrunc`.  Given that the extend/trunc are really doing semantically different things and take additional immediate arguments, this feels like the right thing to do and makes the `Load`/`Store` cases really short and sweet.

Patch also contains minor fix of source info in `Const` for better error reporting.


rossberg-chromium:
LGTM, I like it. (The one odd case that is not prevented in the AST by construction is `i32.load32`/`i32.store32`, but perhaps that's bearable.)

Does it still make sense to call it `mem_type`, though? How about
```
type mem_size = Mem8 | Mem16 | Mem32
```

lukewagner:
> LGTM, I like it. (The one odd case that is not prevented in the AST by construction is 
> i32.load32/i32.store32, but perhaps that's bearable.)

I know and it bugged me too :)  I was thinking we could have some new variant that enumerated the valid `(mem_size, value_type)` pairs, but that seemed a bit worse overall.

`mem_size` sounds good, I'll try that.

rossberg-chromium:
Something could probably be done with polymorphic variants and GADTs, but I
don't think it's worth going there.

PR #87, opened 2015-09-25
Conversion tests

sunfishcode:
This adds some more tests for conversion operators, including coverage of overflow boundary conditions. It also fixes some the overflow range checks to properly detect overflows.

It also tidies up some existing tests and merges unsigned.wasm into i32/i64/conversions.wasm.


rossberg-chromium:
lgtm

sunfishcode:
Merging with lgtm above. If any bugs are found, please file new spec issues. For discussion of any of the semantics, please file design issues as this is implementing what's currently in the design.

PR #88, opened 2015-09-29
Rename "const" to "literal" to match the language in the design.

sunfishcode:
Per the discussion in https://github.com/WebAssembly/design/pull/377 the preferred name for these nodes is "literal".


sunfishcode:
Closing, since the resolution is to stay with `const`.

PR #89, opened 2015-09-29
Add constant offsets to spec

lukewagner:
This PR adds the immediate offset described in [AstSemantics.md#addressing](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#addressing).  I'd rather not block on the broader question of opcode naming in [design/#374](https://github.com/WebAssembly/design/pull/374), so this PR just does what seems like the obvious extension to the current scheme of putting load/store immediates in the opcode name (prefixed with `+` for the offset).

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -124,7 +124,7 @@ unop:  ctz | clz | popcnt | ...
 binop: add | sub | mul | ...
 relop: eq | ne | lt | ...
 sign: s|u
-align: 1|2|4|8|...
+align: align=(1|2|4|8|...)
 cvtop: trunc_s | trunc_u | extend_s | extend_u | ...
 
 expr:
@@ -145,8 +145,8 @@ expr:
   ( return <expr>? )                             ;; = (break <current_depth> <expr>?)
   ( get_local <var> )
   ( set_local <var> <expr> )
-  ( <type>.load((8|16)_<sign>)?(/<align>)? <expr> )
-  ( <type>.store(/<align>)? <expr> <expr> )
+  ( <type>.load((8|16)_<sign>)? <align>? <expr> )
+  ( <type>.store <align>? <expr> <expr> )
   ( <type>.const <value> )
   ( <type>.<unop> <expr> )
   ( <type>.<binop> <expr> <expr> )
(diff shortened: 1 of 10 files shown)


kg:
Not okay with this. The parsing rules and opcode names are already too complicated; this is worse and  in a non-general way. I'll move over to 374 for discussion, I guess.

ml-proto/README.md
```diff
@@ -126,10 +126,10 @@ type expr =
...
   | GetLocal of var                         (* read local variable
   | SetLocal of var * expr                  (* write local variable
-  | LoadGlobal of var                       (* read global variable
-  | StoreGlobal of var * expr               (* write global variable
-  | Load of loadop * expr                   (* read memory address
-  | Store of storeop * expr * expr          (* write memory address
+  | Load of memop * expr                    (* read memory address
+  | Store of memop * expr * expr            (* write memory address
+  | LoadExtend of extendop * expr           (* read memory address with sign- or zero-extension
+  | StoreTrunc of truncop * expr * expr     (* write memory address with truncation
```
sunfishcode: Here and elsewhere in this patch, the design repo calls this "wrap" instead of "trunc", for consistency with the conversion opcodes.

lukewagner: Good point, filed #91 to address.

lukewagner:
@kg The point is that the naming can, has been, and will continue to evolve separately from semantics and so there's no reason to derail with name bikeshedding.  However, if we can reach a quick consensus in design/#374 then I'm happy to update this PR.  I really don't care that much naming at this point.

lukewagner:
(Getting back to this now; will update with new patch to implement [new immediate syntax](https://github.com/WebAssembly/design/pull/374#issuecomment-144436947).)

lukewagner:
Ok, updated and ready for review.

ml-proto/spec/memory.ml
```diff
@@ -79,9 +80,13 @@ let grow mem n =
   Array1.blit (Array1.sub !mem 0 host_old_size) (Array1.sub after 0 host_old_size);
   mem := after
 
-let rec loadn mem n a =
+let effective_address a o =
+  if (Int64.sub Int64.max_int a) < o then raise Bounds;
```
sunfishcode: This is doing a signed comparison. The design text describes the effective address as an unsigned computation, so it seems like this should be checking for unsigned overflow instead.

sunfishcode:
lgtm, with one comment above.

rossberg-chromium:
lgtm

ml-proto/spec/memory.ml
lukewagner: Good point!

PR #91, opened 2015-09-30
Rename StoreTrunc to StoreWrap for consistency

lukewagner:
@sunfishcode pointed out that we consistently use "wrap" instead of "truncate" to describe larger-to-smaller integer conversions, so this PR renames `StoreTrunc` to `StoreWrap` which is nicer anyway since it's shorter and replaces an abbreviation with a real word.  Also tidies up README.md.


lukewagner:
Merging since this seems pretty uncontroversial.

PR #92, opened 2015-09-30
Fix the path to the wasm executable to match what the Makefile builds.

### LINKED ISSUE #59: multiple build processes, processes disagree
When using src/Makefile, one gets an executable at src/wasm. However, runtests.py expects an executable at src/main.native. It'd be useful for these to agree; which one is preferable?

sunfishcode:
This fixes #59. I have no idea if it works on Windows. If this fix is inappropriate, please tell me what I can do to fix this instead. Thanks!

### DIFF
--- a/ml-proto/runtests.py
+++ b/ml-proto/runtests.py
@@ -58,7 +58,7 @@ def rebuild_interpreter(path):
     raise Exception("Interpreter has not been built. Looked for %s" % path)
 
 if __name__ == "__main__":
-  interpreterPath = os.path.abspath("./main.native")
+  interpreterPath = os.path.abspath("./wasm")
 
   try:
     os.makedirs("test/output/")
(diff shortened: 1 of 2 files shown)


pjuftring:
The travis-testscript still creates "main.native" and therefore fails.
Everything's fine on Windows, at least on Cygwin.

sunfishcode:
@pjuftring Fixed, thanks!

rossberg-chromium:
I suspect this breaks the `--build` option of the script, unless you also
adapt the `rebuildInterpreter` function. Now that the Makefile is using
ocamlbuild, the function should probably just invoke that instead of
duplicating the parameters.

sunfishcode:
I updated --build to just invoke the Makefile. It Works For Me (tm).

sunfishcode:
Merging, because it fixes #59 and no one has objected so far.

PR #93, opened 2015-09-30
Use the arcane ocamlbuild incantation that actually gets debug symbols.

sunfishcode:
Apparently ocamlbuild decides whether or not to include debug symbols sufficient to perform a backtrace based on the output executable file name. This commit changes the name used, so that we can get backtraces.


rossberg-chromium:
LGTM (one reason why I'm not a big fan of ocamlbuild yet)

PR #95, opened 2015-09-30
Move the contents of ml-proto/src into ml-proto.

sunfishcode:
The extra level of directory hierarchy doesn't appear to have a lot of
benefit, and it makes working with the code less fun (see e.g. the
diffs in README.md).


lukewagner:
I think `src` made sense when it contained all the files directly, but now that they are in the three subdirs, I agree it's mostly just an annoyance.

rossberg-chromium:
LGTM

sunfishcode:
Great, merging.

PR #96, opened 2015-09-30
Replace the AST definition in README.md with a link to ast.ml.

sunfishcode:
The README.md copy keeps going out of date with asm.ml. The ML source
code ought to be readable as documentation, since it also serves as a
specification, so if there is a readability issue with ast.ml, we
should fix it.

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -106,48 +106,13 @@ Here, the external format is S-expressions, but similar considerations would app
 
 ## Internal Syntax
 
-The core language is defined in `ast.ml`, and looks as follows:
-
-```
-type var = int
-
-type expr =
-  | Nop                                     (* do nothing
-  | Block of expr list                      (* execute in sequence
-  | If of expr * expr * expr                (* conditional
-  | Loop of expr                            (* infinite loop
-  | Label of expr                           (* labelled expression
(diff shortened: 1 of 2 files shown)


ml-proto/src/spec/ast.ml
```diff
@@ -74,31 +74,31 @@ type literal = value Source.phrase
...
+  | Switch of value_type * expr * arm list * expr (* switch, latter expr is default *)
+  | Call of var * expr list                       (* call function *)
+  | CallImport of var * expr list                 (* call imported function *)
+  | CallIndirect of var * expr * expr list        (* call function through table *)
+  | Return of expr option                         (* return 0 to many value *)
+  | GetLocal of var                               (* read local variable *)
+  | SetLocal of var * expr                        (* write local variable *)
+  | Load of memop * expr                          (* read memory address *)
+  | Store of memop * expr * expr                  (* write memory address *)
   | LoadExtend of extendop * expr
```
rossberg-chromium: Can you fill in the missing comments on these two?

rossberg-chromium:
LGTM, but don't forget to update the link when you commit the patch
removing src/. :)

sunfishcode:
Link updated; thanks!

PR #97, opened 2015-09-30
Rename 'modul' to 'module_'.

sunfishcode:
I suppose this is personal familiarity, but 'modul' initially made me wonder if it's some strange OCaml keyword that I hadn't previously encountered.


rossberg-chromium:
LGTM, both are sufficiently ugly choices, so I don't care either way :)

PR #99, opened 2015-10-01
Labels for blocks, loops, & switches

rossberg-chromium:
Inspired by Dan's larger patch (https://github.com/WebAssembly/spec/pull/98/), this one introduces sugar to define a label directly as part of a `block`, `loop`, or `switch`. A `loop` allows up to two labels, one to exit, the other to continue the loop.

The patch also redefines `return` as sugar for a `break`, slightly simplifying the spec, and making the core language compositional (you can now substitute function bodies, up to variable renaming).


lukewagner:
So in both this and #98, I'm wondering if it's a good idea to desugar ops into other ops in `host/parser.mly` b/c that does sorta make them vanish from `spec/` (in particular `ast.ml`) which is what I'm considering "the spec".  I definitely like simplifying the spec, but I feel that the type of complexity I want to avoid is cross-cutting concerns or ad hoc stuff; adding new cases that are almost the same as other cases seems like no big deal; we're not defining a calculus here and nobody is trying to print out wasm semantics on a poster (thought I bet with K we could... ;).

Alternatively, there is another strategy that has already been on my mind: since there is already some collapsing going on where many "ops" (that have different names in AstSemantics.md) are given a single ML constructor in `ast.ml`, I was wondering if it would make sense to have two ML AST datatypes: a "surface" AST that is generated by `parser.mly` and has exactly 1 constructor for every AstSemantics.md op name, and a "kernel" AST that is basically what we have now (including your PR).  Then, the first thing we do in `spec/` is to desugar the surface AST into the kernel AST (and thus all of `check.ml` and `eval.ml` would stay the same).  That way, we're making the desugaring very clear in "the spec".

rossberg-chromium:
Yes, it's a good point, and has been bothering me as well. I'll give it a
try to introduce a "sugared ast" and move the current desugaring out of the
parser.

lukewagner:
Great!

rossberg-chromium:
One question of course is what to do with the use and resolution of symbolic names. I think it would be inadequate to have those appear inside `spec`, since they are purely an artifact of the textual representation. So I'll probably leave that in the parser, although it is the actual tricky part of the logic -- the remaining desugaring is super trivial atm.

lukewagner:
Agreed.

AndrewScheidecker:
Could this also get rid of `label`? It seems redundant if `block` has a branch target.

I also think it's less confusing to call `break` `branch` once it can be used to continue loops.

rossberg-chromium:
It's only surface syntactic sugar for `label`, so no, at least not yet.

I'd be fine with `branch`, but I actually think `break` still is more accurate, since it isn't an arbitrary jump, but always terminates a surrounding expression. This applies to the continue label as well, which terminates the loop's body.

AndrewScheidecker:
> It's only surface syntactic sugar for `label`, so no, at least not yet.

I mean that you could get rid of the `label` node in the sugared syntax.

rossberg-chromium:
Ah, I see. Well, then it wouldn't be a desugaring any more, but a proper
translation. I think it's better to keep the input syntax a superset of the
core syntax, and allow access to desugared constructs directly. If for
nothing else, then for writing specific tests of malformed code.

rossberg-chromium:
Superseded by #117

PR #104, opened 2015-10-01
Parsing, testing infrastructure, and tests for f32 and f64

sunfishcode:
- support for infinity, NaN with arbitrary bit pattern, and hexadecimal float literals
 - assert_eq_bits and assert_nan for testing floating point results
 - tests for all floating point operators


ml-proto/host/lexer.mll
```diff
@@ -90,10 +90,16 @@ let tick = '\''
...
-let num = ('+' | '-')? digit+
-let hexnum = ('+' | '-')? "0x" hexdigit+
+let sign = ('+' | '-')?
+let num = sign digit+
+let hexnum = sign "0x" hexdigit+
 let int = num | hexnum
-let float = (num '.' digit+) | num ('.' digit+)? ('e' | 'E') num
+let float = (num '.' digit+)
+          | num ('.' digit+)? ('e' | 'E') num
+          | sign "0x" hexdigit* '.'? hexdigit* 'p' sign digit*
```
rossberg-chromium: This would allow weird things like `0xp` or `0x.p` to be valid float literals. Is that intended?

rossberg-chromium:
I would prefer kicking out the existing `assert_eq` and rename `assert_eq_bits` to it. There is no good reason to ever use the former. (I feel somewhat strongly about this because we actually had this very situation in the V8 test suite, which provides both AssertEqual and AssertSame. Everybody was intuitively using the former, because it has the more natural name. At some point we discovered that many tests actually were broken because of this, and not testing what the authors thought they would. And that had been hiding some real bugs in the features being tested.)

ml-proto/host/lexer.mll
sunfishcode: No; I've now updated this to require at least one leading significand digit and at least one exponent digit, and added a test for a significand with a trailing dot, as in `0x1.p4`.

sunfishcode:
Ok. I updated this PR to merge `assert_eq` and `assert_eq_bits` into a single `assert_same`. I think I like that name better than `assert_eq_bits` because it anticipates possible future types which don't have exposed bitwise representations (GC handles, etc.).

rossberg-chromium:
How about naming it `assert_return`? That would be more analogous to `assert_fail`, and would avoid the confusion we already saw a couple of times where people expected the assertion to be symmetric, with arbitrary expressions being allowed on either side. (`assert_nan` should probably be `assert_return_nan` in that case.)

sunfishcode:
`assert_return` and `assert_return_nan` sound good to me. I updated the PR accordingly.

rossberg-chromium:
lgtm

Issue #102, opened 2015-10-01
'make' generates 4 executables


lukewagner:
After the recent Makefile changes (which makes things much nicer), the first run of 'make' generates 'wasm' and 'unopt'.  The second run adds 'main.native' and 'main.d.byte'.  Ideally, 'make' would only generate 1 executable.  For ease of use, maybe this should be the unopt build and you have to type 'make opt' to get the opt build?

PR #105, opened 2015-10-02
Makefile improvements

rossberg-chromium:
- Only build opt version by default.
- Mark `main.native` and `main.d.byte` as intermediate targets.
- Introduce test target to invoke test runner.

This should address issue #102 .

### DIFF
--- a/ml-proto/Makefile
+++ b/ml-proto/Makefile
@@ -5,8 +5,8 @@
 #
 
 NAME =		wasm
-NAME_OPT =      $(NAME)
-NAME_UNOPT =    $(NAME).unopt
+NAME_OPT =      $(NAME).opt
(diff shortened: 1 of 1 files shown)


lukewagner:
Thanks!

sunfishcode:
The debug build has much more extensive backtraces, and since performance isn't really important for the spec interpreter, would it make sense to make the debug build be the default?

rossberg-chromium:
Done

sunfishcode:
lgtm

lukewagner:
The issue I see now is that `make` and `runtests.py` don't seem to observe file changes so i have to force rebuild.

rossberg-chromium:
Hm, okay, will look into that.

PR #107, opened 2015-10-02
Only validate memops when there is a memory section

lukewagner:
This PR makes it a validation error to use linear memory operations (load, store, page_size, resize_memory, memory_size) when there is no memory section declared in the module.  Since it is possible to have a `(memory 0)` section (which allows memory ops but doesn't require allocation), there is no loss of expressiveness for producers.  Rather, this is a simple a priori way to capture the wasm producer's intent to never use linear memory that allows the impl to take advantage of this (to reduce footprint, avoid corner cases in codegen) without having to worry about a linear memory later appearing (via resize_memory or dynamic linking).  This restriction could later be backwards-compatibly relaxed if necessary.  The main expected use is apps using [GC](https://github.com/WebAssembly/design/blob/master/GC.md) for all memory.

### DIFF
--- a/ml-proto/spec/eval.ml
+++ b/ml-proto/spec/eval.ml
@@ -25,7 +25,7 @@ type instance =
   imports : import list;
   exports : export_map;
   tables : func list list;
-  memory : Memory.t;
+  memory : Memory.t option;
   host : host_params
 }
 
@@ -87,6 +87,9 @@ let numerics_error at = function
   | Numerics.InvalidConversionToInteger ->
       error at "runtime: invalid conversion to integer"
   | exn -> raise exn
+
+let some_memory c =
+  match c.module_.memory with Some m -> m | _ -> assert false
 
 let some v at =
   match v with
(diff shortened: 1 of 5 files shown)


titzer:
We should make this the default, as memory just causes so many problems.

Haha, jk. I like this idea.

sunfishcode:
lgtm

ml-proto/spec/eval.ml
```diff
@@ -88,6 +88,9 @@ let numerics_error at = function
       error at "runtime: invalid conversion to integer"
   | exn -> raise exn
 
+let some_memory c =
+  match c.module_.memory with Some m -> m | _ -> assert false
```
rossberg-chromium: Can we make this a proper runtime error instead of an assertion failure? As a rule of thumb, always assume that the evaluator can be fed unchecked code. In fact, we even do so for `invoke` arguments.

lukewagner: As brought up in #109, I wonder if we can stop feeding the evaluator unchecked code.  In that case, I think it'd be nice if our user of `assert` in `eval` specifically asserted "checked earlier" and so we had a clear documented distinction between actual runtime errors and things that should never happen, assuming soundness.  That is, if someone did a soundness proof, they'd prove we never hit an assert.

lukewagner: ... but I should probably leave that for a separate issue so it can be considered orthogonally; I'll put in the runtime error here for now.

rossberg-chromium: I think we still want to distinguish two classes of execution errors (aside
from traps): bugs in the semantics (i.e., the static semantics not being
sound wrt the operational semantics), and bugs in the implementation. The
latter would be assertion failures, but the former should be observed by a
dedicated exception.

lukewagner: That sounds good.  So maybe a new `Soundness` exn type?  Since this shouldn't ever happen, seems like it could be made more terse to raise and leave out position/string.

rossberg-chromium: Maybe let's call it `Crash`. Soundness then is the property that a crash
cannot happen for a checked program. :)

Not sure why you'd want to drop the error info. There aren't many places
that raise this, so I would err on the side of being more informative in
those cases.

lukewagner: Not strong reason: just because the source info clutters a bit and while it's generally worth it because it helps us when developing, the `Crash`es should never happen, even when developing.  That's why I liked `assert`.  Since impl=spec (sortof), I do wonder if we need to discriminate bugs in impl and bugs in semantics.

rossberg-chromium: > Since impl=spec (sortof), I do wonder if we need to discriminate bugs in impl and bugs in semantics.

I think the "sortof" is key here. As much as we try to approximate
that ideal, there's still a lot of non-declarative code in the
implementation, and not everything is in the spec directory anyway. So
I think the distinction is useful.

lukewagner: I can see your point and I don't have a strong opinion on this.

lukewagner:
Merging with comments addressed.

Issue #80, opened 2015-09-18
OCaml's Bigarray module can't handle large linear memory sizes


sunfishcode:
memory.ml is [currently](https://github.com/WebAssembly/spec/blob/master/ml-proto/src/spec/memory.ml#L5) using OCaml's [Bigarray](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Bigarray.html) module, specifically [Bigarray.Array1](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Bigarray.Array1.html), to represent linear memory.

Bigarray's interfaces all use OCaml's [int](http://caml.inria.fr/pub/docs/manual-ocaml/values.html#sec77) type for array extents and index values. While int has a host-dependent size, it is 31 bits on some common systems. It is also signed, so it can only hold values less than 1<<30.

Consequently, on 32-bit hosts, it seems the WebAssembly reference interpreter is limited to linear memory sizes less than 1 GiB, even when the underlying host is capable of allocating that much memory.

Am I understanding everything here correctly? And if so, are there any alternatives to Bigarray which allow for bigger sizes?

sunfishcode:
One more note: I noticed this while working on wasm's integer operator semantics, and noticed that `address_of_value` currently does `Int32.to_int i`, which silently wraps if the value is outside the OCaml int range, which is not the intended semantics.

rossberg-chromium:
I think you're correct. This is a limitation on 32-bit systems.

I suggest defining `address` and `size` to be `nativeint`. Then
`address_of_value` would no longer wrap.

Furthermore, the `create` function then needs to check whether the size
overflows an int, and if so, raise `Out_of_memory`. That way, sizes too
large are not observably different from OOM, which seems like suitable
behaviour.

Analogously, load & store functions in memory.ml need to check whether the
index overflows int range and raise `Bounds` in that case.

lukewagner:
I consider this basically resolved by #108 which at least confines the hand waving to [one well-commented block](https://github.com/WebAssembly/spec/blob/6867335cdb77ab71980af7b2ff74ecc6676626cb/ml-proto/spec/memory.ml#L27).

PR #108, opened 2015-10-02
Use int64 for addresses and sizes, isolate BigArray int limitation

lukewagner:
This PR is intended to address some of the issues raised in #80 by using `int64` throughout `spec`/.  The BigArray `int` limitation is, as suggested in #80, addressed by modeling creation of an array bigger than `max_int` as an `Out_of_memory` trap.

### DIFF
--- a/ml-proto/spec/eval.ml
+++ b/ml-proto/spec/eval.ml
@@ -14,7 +14,7 @@ let error = Error.error
 type value = Values.value
 type func = Ast.func
 type import = value list -> value option
-type host_params = {page_size : int}
+type host_params = {page_size : Memory.size}
 
 module ExportMap = Map.Make(String)
 type export_map = func ExportMap.t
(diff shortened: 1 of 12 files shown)


ml-proto/spec/memory.mli
```diff
@@ -4,11 +4,9 @@
 
 type memory
 type t = memory
-type address = int
```
rossberg-chromium: Specific reason to remove these instead of making them synonyms for int64? They have some documentation value.

ml-proto/spec/eval.ml
```diff
@@ -275,7 +274,7 @@ and eval_func (m : instance) (f : func) (evs : value list) =
 let init_memory ast =
   match ast with
   | None ->
-    Memory.create 0
+    Memory.create Int64.zero
```
rossberg-chromium: You can write `0L` for this (same below).

ml-proto/spec/memory.mli
lukewagner: The reason I removed was that, before, we half treated these as opaque types and, with this PR, we fully assume they are int64s, so it seemed like an unnecessary abstraction.  But I wasn't considering the documentation value; I can add them back.

lukewagner:
Updated for comment and rebased.

lukewagner:
@rossberg-chromium lgty?

rossberg-chromium:
LGTM modulo nits

lukewagner:
Merging with nits fixed.

PR #111, opened 2015-10-05
Refactoring float32 and float64 prims into a single FloatPrims functor

naturaltransformation:
This is a simple refactoring of F32 and F64 to avoid fairly significant duplication.


rossberg-chromium:
Looks good, but I would prefer to factor this slightly differently:

- Rename `Float.FloatPrims` to `Float.Make`.
- Move the instantiations of `Float.Make` into separate modules `f32` and `f64` like existed before, which now just contain `include Float.Make(struct ... end)`
- Drop `float.mli`.
- Move the `FLOAT` signature into the `Float` module, and remove the `Floatsig` module.

This way, (1) you avoid all the extra `Float.FXY` abbreviations, and (2) match the idioms of the stdlib more closely.

naturaltransformation:
Good point. I am matching the stdlib style now, including signature names in the module.

rossberg-chromium:
LGTM modulo the TODO.

ml-proto/spec/float.ml
```diff
@@ -0,0 +1,145 @@
...
+  val abs : t -> t
+  val neg : t -> t
+  val copysign : t -> t -> t
+  val eq : t -> t -> bool
+  val ne : t -> t -> bool
+  val lt : t -> t -> bool
+  val le : t -> t -> bool
+  val gt : t -> t -> bool
+  val ge : t -> t -> bool
+  val size : int
```
sunfishcode: `size` wasn't in the interface before. Is there a need for it?

naturaltransformation: Doesn't seem to be anymore. size was there a few commits ago when int cast was integrated in float, but I guess I just let it reappear. In any case, I removed it now. Thanks!

ml-proto/spec/float.ml
```diff
@@ -0,0 +1,181 @@
...
+      of_signless_string (String.sub x 1 (len - 1)) (len - 1)
+    else
+      of_signless_string x len
+
+  let to_string x =
+    (if x < Rep.zero then "-" else "") ^
+      let a = abs x in
+      let af = arith_of_bits a in
+      if af <> af then
+        ("nan(0x" ^ Rep.print_nan a ^ ")")
```
sunfishcode: As a minor nit, `print_nan` here should be named something like `print_nan_significand_digits` since it's just printing that one part, rather than the whole NaN.

naturaltransformation:
Ok, done with the rename.

sunfishcode:
lgtm; thanks!

PR #118, opened 2015-10-08
Split f32.ml and f64.ml into multiple files.

sunfishcode:
These tests are pretty big relative to the other tests; splitting out
the comparison operators into a separate file shrinks them significantly.


rossberg-chromium:
lgtm

PR #120, opened 2015-10-09
Implement has_feature.

sunfishcode:
This implements `has_feature` as described in [the design](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#feature-test).

### DIFF
--- a/ml-proto/spec/check.ml
+++ b/ml-proto/spec/check.ml
@@ -90,17 +90,16 @@ let type_cvt at = function
     | DemoteFloat64 -> error at "invalid conversion"
     ), Float64Type
 
-type type_hostop_type = {
-  hostop_ins : value_type list;
-  hostop_out : expr_type;
-  hostop_hasmem : bool
-}
-
+(*
+ * This function returns a tuple of a func_type and a bool, with the bool
+ * indicating whether the given function requires a memory declaration to be
+ * present in the module.
+ *)
 let type_hostop = function
-  | PageSize -> {hostop_ins = []; hostop_out = Some Int32Type; hostop_hasmem = true}
-  | MemorySize -> {hostop_ins = []; hostop_out = Some Int32Type; hostop_hasmem = true}
-  | GrowMemory -> {hostop_ins = [Int32Type]; hostop_out = None; hostop_hasmem = true}
-  | HasFeature str -> {hostop_ins = []; hostop_out = Some Int32Type; hostop_hasmem = false}
+  | PageSize -> ({ins = []; out = Some Int32Type}, true)
+  | MemorySize -> ({ins = []; out = Some Int32Type}, true)
+  | GrowMemory -> ({ins = [Int32Type]; out = None}, true)
+  | HasFeature str -> ({ins = []; out = Some Int32Type}, false)
 
 
 (* Type Analysis *)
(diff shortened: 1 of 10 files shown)


sunfishcode:
This is now rebased on top of the host opcode grouping changes, which required teaching the host opcode code about host opcodes that don't need memory.

kg:
It would be cool to add a known-always-present feature so we can test both the positive and negative case in the automated test. That's a pretty small thing though. Otherwise this LGTM.

sunfishcode:
Good idea. I added a feature called "wasm" which is claimed to be unconditionally supported :-).

ml-proto/spec/check.ml
```diff
@@ -90,10 +90,17 @@ let type_cvt at = function
     | DemoteFloat64 -> error at "invalid conversion"
     ), Float64Type
 
+type type_hostop_type = {
```
rossberg-chromium: Instead of introducing another variant of function type for a singular use, can't `type_hostop` just return a pair `func_type * bool`?

ml-proto/spec/eval.ml
```diff
@@ -298,6 +304,9 @@ and eval_hostop host mem hostop vs at =
     Memory.grow mem delta;
     None;
 
+  | HasFeature str, [] ->
+    Some (Int32 (if (host.has_feature str) then 1l else 0l))
```
rossberg-chromium: Nit: redundant parens

rossberg-chromium:
I'm afraid I still don't understand how a feature test _inside_ the language is supposed to make sense -- AFAICS it is utterly useless as long as there's no defined way to actually handle unsupported features _inside_ the language as well. And I have no idea how that would be done or specced or implemented. The design docs don't explain that either, only how to do feature testing on the meta level (which doesn't require the operator). It seems to me that putting in the operator now is starting with a (useless) step 2 before we have a clue about step 1.

But other than these fundamental doubts about the premise, and the nits above, LGTM. :)

lukewagner:
@rossberg-chromium The intended usage I'm anticipating in the MVP is described [here](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#better-feature-testing-support).

rossberg-chromium:
@lukewagner, yes, that's what I looked at, but as far as the MVP is concerned, the only workable suggestions there are doing meta-level feature detection (or am I misreading that?). So what still puzzles me is how _in-language_ feature detection is useful before you also add the actual ability to "decode-through" unrecognised features?

lukewagner:
@rossberg-chromium Because we'd want implement layer 1 decoder (which also performed the polyfilling) in wasm.  That (tiny) module needs to learn what features are supported *somehow*, so might as well be from within wasm; otherwise we'd have to add it as a JS feature and that seems weird.

rossberg-chromium:
I see! Okay, sorry that I didn't get that from the description. I wouldn't mind clarifying this use case (and perhaps moving the explanation to the FeatureTest page itself). ;)

lukewagner:
@rossberg-chromium Will do; I forgot we even had FeatureTest.md :)

kg:
> So what still puzzles me is how in-language feature detection is useful before you also add the actual ability to "decode-through" unrecognised features?

For the record, my binary encoding proposal already covers being able to decode unrecognized AST nodes.

ml-proto/spec/check.ml
sunfishcode: That works.

sunfishcode:
The patch is updated to address the comments above. Any further comments?

rossberg-chromium:
lgtm

PR #121, opened 2015-10-09
Add stdio.write import for writing a sequence of bytes to stdout

kg:


ml-proto/test/stdio_write.wast
```diff
@@ -0,0 +1,14 @@
+(module 
+  (import $write "stdio" "write" (param i32 i32))
+
+  (memory 4096 4096 (segment 0 "hello, world!\0D\0A\00"))
```
sunfishcode: In C/C++ code compiled to wasm, strings don't contain the "\0D", because stdout is in "text" mode, which ordinarily means that the C library translates newline into carriage-return-line-feed on Windows. However in wasm's case, the C library won't know that it's running on Windows, so it won't know that it has to do this. I think it's best if stdio.write handles this itself so that wasm programs don't have to worry about it.

kg: That would make it unusable for binary data.

sunfishcode: The alternative is to be unusable for text strings in C/C++. How about a separate function for binary data?

kg: ```write``` is a base primitive for writing whatever to stdout. if you want to print text, your stdlib would do any transforms like ```\n``` -> ```\r\n``` on windows.

sunfishcode: How should the stdlib detect that is is on Windows?

kg: I still don't understand why this even matters.

sunfishcode: Which stdlib is going to do `\n` -> `\r\n` translation, the stdlib linked into the wasm app, or the stdlib linked into ocaml? If the former, it doesn't know that it's on Windows, because wasm presents an OS-independent abstraction, so it doesn't know whether it needs to do that or not. If the latter, then that's what I'm asking for, and the testcase here doesn't need its `\r`.

kg: Why do we need newline translation? What problem does it solve? This is a testcase primitive for writing to an output file, which will be compared against expected output.

sunfishcode: It is very useful to be able to run the testcases by hand, with output going to the console, and if the Windows console needs us to print a `\r`, then that's a problem we need to solve.

sunfishcode: Alternatively, if we just care about writing to an output file and reading it with python, another option would be to just remove the `\r` from the testcase.

kg: runtests already puts the output somewhere you can look at it, in a file. every modern text editor I've ever used (except maybe win32 notepad?) can handle unix newlines. Newline translation would break most of my tests that generate files, so I'm not adding it to write. You're welcome to add write_text or something that does platform-specific translation.

Next time I'm on windows I'll see whether the windows console needs crlf, but I'm not sure it does.

kg: > Alternatively, if we just care about writing to an output file and reading it with python, another option would be to just remove the \r from the testcase.

The ```\r``` is there specifically to ensure that you can write binary data. I should probably make sure writing a raw ```\00``` works as well.

sunfishcode: What I'm actually concerned about here is that there's wasm code with a `\r\n` in it, which I find very concerning. I assumed you put it there because of Windows; if not, then can we just remove the `\r`?

jfbastien: IIUC this is meant to test that `\r` is unchanged. I strongly believe that `fwrite` shouldn't change the bits it's given, so the test should exist in some form. Are you suggesting the form be changed?

sunfishcode: Would you mind making a separate test for binary output then? I saw this test as the "hello world" test that would be used as an example of what wasm code looks like. I don't want to risk anyone thinking that wasm code needs a `\r` to print hello world.

As you say, a separate test for binary output could test `\0` and other interesting characters.

titzer: "write" for binary data, "print" for text?

jfbastien: I think separate test files is fine.

kg:
Welp. Looks like windows ocaml is doing line normalization. I probably need to find another way to write raw bytes to stdout.

```
======================================================================
FAIL: test/stdio_write.wast (__main__.RunTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./runtests.py", line 37, in <lambda>
    return lambda self : self._runTestFile(*rec)
  File "./runtests.py", line 34, in _runTestFile
    self.assertEqual(expectedText, actualText)
AssertionError: '\x89PNG\r\n\x1a\n\x00' != '\x89PNG\r\r\n\x1a\r\n\x00'
```

pjuftring:
I added the line

```ocaml
set_binary_mode_out stdout true;
```

And it seems to work, but I don't know if it does on linux.

kg:
Switched to print_bytes, that might disable newline conversion. Or it might not, since ```Bytes``` is initialized from ```char``` (?????)

rossberg-chromium:
I don't think `print_bytes` disables conversion (have you tried it?).

It's just like in the C or C++ stdlibs: newline conversion is a property of the stream, not the source. And `stdout` is a text stream. If you want to disable newline conversion on Windows, you need to go to binary mode. As @pjuftring pointed out, you can use `set_binary_mode_out` to change the mode after the fact. (Just make sure to switch it back before the function returns, so that you don't break other output to stdout.)

kg:
> I don't think print_bytes disables conversion (have you tried it?).

Nope, I don't have access to Windows at work :-) I'll come up with a fix on my Windows machine.
Good point about the pattern match, didn't realize you could use them on lists.

kg:
Yeah, ```print_bytes``` doesn't disable conversion. Weird :( I'll just do ```set_binary_mode_out```. One caveat is that you can't query the current binary mode state, so there's no way to restore it after setting it - as a result this function will trample over the output mode. I don't like that, but I can't see an alternative.

Your suggested pattern-match doesn't seem like it type-checks, at least as written. I'm not actually sure why, as it looks correct to me.

```
Error: This expression has type
         Eval.instance ->
         ((int32, 'a, 'b, 'c) Values.op * (int32, 'd, 'e, 'f) Values.op) list ->
         'g option
       but an expression was expected of type
         Eval.instance -> Values.value list -> 'h option
       Type (int32, 'a, 'b, 'c) Values.op * (int32, 'd, 'e, 'f) Values.op
       is not compatible with type
         Values.value = (I32.t, I64.t, F32.t, F64.t) Values.op
```

Should I just update the PR without using the pattern-match, or do you not want to merge without it? The pattern-match does seem like the right way to do this.

rossberg-chromium:
Ah, sorry, I keep forgetting that OCaml separates list items by semicolons. That should fix it.

kg:
OK, I've revised it so that it works correctly on Windows now.

Had some confusion over the fact that to disable newline normalization, you do ```set_binary_mode_out stdout false``` - the meaning of the arg is reversed.

rossberg-chromium:
lgtm

ml-proto/host/builtins.ml
```diff
@@ -1,9 +1,50 @@
 open Source
+open Eval
```
sunfishcode: From a readthrough, the only use of the Eval module in this file is the Eval.memory_for_module. Can this open be removed?

ml-proto/host/builtins.ml
```diff
@@ -1,9 +1,50 @@
...
+open Memory
```
sunfishcode: It looks like this open may be similarly unneeded.

ml-proto/host/builtins.ml
```diff
@@ -1,9 +1,50 @@
...
+open Types
```
sunfishcode: (Sorry for nit-picking these `open` directives; my experience with OCaml so far has been that having lots of `open`s leads to increased chances of confusion from name collisions or struct type literal inferences). Is this `open` needed?

ml-proto/host/builtins.ml
```diff
@@ -1,9 +1,50 @@
...
+  if List.length vs != 2 then
+    Error.error at "stdio.write expects 2 arguments (offset, count)";
+
+  let mem = Eval.memory_for_module at m in
+  match vs with
+  | [Values.Int32 _offset; Values.Int32 _count] ->
+    let offset = (Int64.of_int32 _offset) in
+    let count  = (Int64.of_int32 _count) in
+
+    set_binary_mode_out stdout false;
```
sunfishcode: As you mentioned elsewhere, this API does the opposite of what it first looks like. Would you mind adding a comment about this?

ml-proto/host/builtins.ml
```diff
@@ -1,9 +1,50 @@
...
 
+let rec stdout_write_inner at mem offset count i =
+  let load_result = Memory.load_extend mem (Int64.add offset i) Mem8 ZX Int32Type in
+
+  begin
+    match load_result with
+    | Values.Int32 byte ->
+      print_char (Char.chr (Int32.to_int byte))
+    | _ ->
+      ignore (Error.error at "load_extend returned wrong type")
```
sunfishcode: Would this be a use case where an `assert` would be appropriate? The type will only be wrong if `Memory.load_extend` produces a value with the wrong type.

ml-proto/host/builtins.ml
```diff
@@ -1,9 +1,50 @@
...
+  end;
+
+  let next = Int64.succ i in
+  let should_iterate = (Int64.compare next count) < 0 in
```
sunfishcode: Is this equivalent to using plain `<` on int64 types, which we use elsewhere?

ml-proto/host/builtins.ml
```diff
@@ -1,9 +1,50 @@
...
+
+  if should_iterate then
+    ignore (stdout_write_inner at mem offset count next)
+  else
+    ();
```
sunfishcode: If I understand what's going on here, the `else ()` here is redundant; is that right, or am I missing something subtle?

ml-proto/host/builtins.ml
kg: Sure

ml-proto/host/builtins.ml
kg: I think I was getting type errors before, but it might have been because I had a diagnostic print in there. It'll probably work if I remove the else.

ml-proto/host/builtins.ml
kg: When I tried ```<``` before it wasn't working, but I had a bunch of strange problems with this code. I'll see if ```<``` works.

ml-proto/host/builtins.ml
kg: Dunno. Don't know anything about ```assert```.

ml-proto/host/builtins.ml
kg: I have no idea. How do the rules work?

sunfishcode: I assume it's just: if you're not using any of the things in the module, you don't need the open. I was hoping it would be just a better of deleting the open, seeing if everything still compiles, and if so, go with it.

rossberg-chromium: A good rule is to limit `open` to a small number of imports that are heavily used, e.g. those that define datatypes with many constructors (like Ast), but is best avoided in most other cases, because it quickly harms readability. (In general, I feel that we have already started overusing it.)

rossberg-chromium:
Closing, there doesn't seem to be a strong need

PR #126, opened 2015-10-12
Group host operators into structured opcode

rossberg-chromium:
Avoids some spec duplication (esp when more such opcodes get added), and perhaps also addresses some of the concerns raised in https://github.com/WebAssembly/design/issues/362 .


lukewagner:
lgtm

PR #128, opened 2015-10-12
Coerce result of void functions to None & remove `host_eval`

### LINKED ISSUE #122: invoke outputs garbage
When invoking functions with a ```void``` return type, ml-proto acts like they have a return type and outputs garbage (the last expression in the function body, seems like?)

```
(module 
  (import $write "stdio" "write" (param i32 i32))

  (memory 4096 4096 (segment 0 "\89\50\4e\47\0d\0a\1a\0a\00"))

  (func $write_png_header
    (call_import $write (i32.const 0) (i32.const 9))
    (i32.const 7)
  )

  (export "write_png_header" $write_png_header)
)

(invoke "write_png_header")
```

prints ```7 : i32``` at the end.

rossberg-chromium:
Fixes #122.

The `host_eval` function breaks once `None` results are coerced, so I had to remove it for good. Only const nodes are still allowed as arguments to asserts (I also looked for ways to introduce a leaner assert syntax while at it, but the explicit type info from the const opcodes is still needed, so I left it as is for the time being).

Also did a few stylistic clean-ups.


lukewagner:
lgtm

Issue #129, opened 2015-10-12
Split tests out into a separate repository?


sunfishcode:
We want to make it as easy as possible for people developing WebAssembly implementations to run the official testsuite, because that's one of our main mechanisms for ensuring portable behavior. @Teemperor suggested on irc that we split the tests out of the spec repo into their own repo, so that they can be included as submodules in other projects.

I think this sounds like a good idea. Any objections?

titzer:
sgtm

lukewagner:
I'm not familiar with how submodules work.  Would it still be possible to have a single PR with both the ml-proto and test changes?  Would it still easy to get a coherent view of ml-proto and tests at a single point in time?  Instead, would it be sufficient if we moved `test/` up to the top-level so that the tests felt like one symmetric part of the spec (sibling to `ml-proto`)?

binji:
You could have a single PR that updates the spec repo and bumps the test repo SHA, but that would require landing the change to the test repo prior.

I think it's better to hold off on this, or do something simpler, like mirror the test directory to a separate repo.

Teemperor:
I think just moving /test to the top level directory of the repo is fine for now (as @lukewagner and @binji said).

sunfishcode:
I opened #130 to move the test directory to the top level. I also renamed in `testsuite` to be slightly more descriptive.

kg:
PRs with dependencies across submodules are a nightmare, so I want to make it clear that we shouldn't do that unless we can't avoid it. Contributors to my personal projects have had a lot of trouble there.

Moving to the top-level sounds good for now, but it's still going to be awkward having a lot of traffic in that folder mixed with spec changes... hrm.

jcbeyler:
Yes moving it to the top-level is really a good idea but will, as kg says, have problems for spec as it changes the definition of the tests. There is no great way of doing those kind of changes:

- Spec changes that breaks the tests
   - Need to update both at the same time

However, not having this forces all projects to have local copies or reference the test suite via the spec repository. And I think that pro outweighs the cons even though as spec still changes things, the developers of spec will see more complications in updating their code. Now two pushes when they break a test.

Finally, it makes adding a feature without tests easier to miss because the feature and the tests are now separated.

sunfishcode:
[...]

PR #130, opened 2015-10-12
Move the test suite into a top-level directory.

sunfishcode:
See discussion in https://github.com/WebAssembly/spec/issues/129 .


sunfishcode:
Now that we have an [official testsuite mirror](https://github.com/WebAssembly/testsuite), this PR is not urgent.

PR #131, opened 2015-10-13
Implement br_if and br_switch.

sunfishcode:
This implements `br_if` and `br_switch`, essentially as was suggested in https://github.com/WebAssembly/design/pull/299. This implements them as spec syntax sugar on top of the existing `if` and `switch` for now.


sunfishcode:
As with #403, I'm closing this for now.

PR #133, opened 2015-10-13
Move signatures into module types table

lukewagner:
In preparation for updating function pointers to match [design/#392](https://github.com/WebAssembly/design/pull/392), a module-level table of signatures is needed which is indexed by `call_indirect` and `func` definitions (replacing the signature stored inside `func` with an index into the table).

This PR allows explicit declaration of the type table (currently only containing function types, but one could imagine other compound types in the future) via `(types (func $T1 (param f32) (result i64)) ...)` (where `T1` desugars to index, like other names).  The existing syntax for defining functions still works and desugars to adding the signature to the table (if it's not already there) and using the resulting index.  However, types can be explicitly referenced via `(func (type $T1) ...)`.  Both can be used (so that parameters can be named), in which case the signatures must match.


ml-proto/host/parser.mly
```diff
@@ -238,35 +284,49 @@ cases :
...
-    { fun c -> bind_local c $3; let f = $6 c in
-      {f with params = $4 :: f.params} }
-  | LPAR RESULT value_type RPAR func_fields
-    { let at = at () in
-      fun c -> let f = $5 c in
-      match f.result with
-      | Some _ -> Error.error at "more than one return type"
-      | None -> {f with result = Some $3} }
+    { {(fst $5) with ins = $3 @ (fst $5).ins},
+      fun c -> anon_locals c $3; (snd $5) c }
```
lukewagner: Just a note to reviewers: so the return value of `func_fields` is now a pair, instead of just the function.  This was necessary to extract the type during the first pass (when the global scope is built) instead of the second pass (where function bodies are generated).  Alternatively, I played a bit with moving the param/result fields out of `func_fields` and into another production that came before `func_fields` in `func`, but I ran into shift/reduce and reduce/reduce conflicts that seemed to require changing the text format to better isolate the signature.

ml-proto/test/func_ptrs.wast
```diff
@@ -0,0 +1,27 @@
...
+        (func $U (param i32))
+    )
+
+    (func $one (type 0) (i32.const 13))
+    (export "one" $one)
+
+    (func $two (type $T) (i32.add (get_local 0) (i32.const 1)))
+    (export "two" $two)
+
+    (func $three (type $T) (param $a i32) (result i32) (i32.sub (get_local 0) (i32.const 2)))
```
sunfishcode: Could you add a comment here confirming to the reader who may not have read all the rules elsewhere that it's explicitly ok to have both a type and params/results, provided they match, because this lets the params be named?

lukewagner: Done

rossberg-chromium:
Mostly looks good, but I have a couple of bike-shedding comments.

- It seems unnecessary to force grouping type definitions into a "section" -- we don't require that for other things either. Just have module fields of the form (type ...). That's both more flexible and simplifies the grammar some.

- Don't nest the bound type name inside the type expression -- that's such a C-ism ;). Can we write type defs as (type $t (func (param ...) (result ...)))? Besides being more logical, such a syntax also scales better to other forms of type definition we might want to add later.

Also, README update is missing. :)

lukewagner:
@rossberg-chromium My motivation wasn't so much C as matching the structure of `(func ...)` definitions which was arbitrary.  Anyhow, both changes you suggested sound good.  All comments should be addressed with the most recent push.

Note: we could greatly simplify `func` and probably reuse `type_def` if we switched the structure of func definitions to: `(func $foo (type (param ...) (result ...)) (local ...) ...)`.  This changes the text format and makes it more verbose in the common case, though.  wdyt?

rossberg-chromium:
That would require changing the syntax of function types, though, so that you can still bind variable names, doesn't it?

rossberg-chromium:
Anyway, LGTM

lukewagner:
> That would require changing the syntax of function types, though, so that you can still bind variable
> names, doesn't it?

Oh, hah, right; I realized this when I was trying it out earlier but then forgot.

PR #134, opened 2015-10-14
Rename `resize_memory` to `grow_memory`, following the design.

sunfishcode:
The previous code interpreted `resize_memory` as an absolute size; this
patch also changes it to be a delta as described in the design.

Also, refactor resizing.wast to be independent of the host page size.


rossberg-chromium:
Looks good to me, other than the question about trap vs OOM.

sunfishcode:
Yes, a dedicated exception would be better. Also, this makes me notice bugs due to not using unsigned arithmetic. I reused I64's unsigned operators here to fix this, though that breaks its encapsulation a little, it seems significantly more convenient than alternatives.

Patch updated. PTAL, thanks!

rossberg-chromium:
LGTM, although I find the name `AddressOverflow` a bit weird in this context. How about `GrowNegative` or s.th like that?

sunfishcode:
What do you think of the name `MemorySizeOverflow`?

rossberg-chromium:
Hm, I'm confused in what sense it is an overflow at all.

sunfishcode:
When the mathematical value of size + delta is too great to be represented as an unsigned number in the result type, it's an unsigned overflow.

rossberg-chromium:
Ah, sorry, you are right, I missed that due to these values being assumed to be unsigned, an overflow is the only reason that new_size < old_size can fail.

Okay, in that case, `SizeOverflow` or just `Overflow` sounds good to me (I wouldn't include the `Memory` prefix, since that's the module's name already).

Sorry for the noise!

sunfishcode:
Renamed AddressOverflow to SizeOverflow. Merging with LGTM above and the rename. Thanks!

JSStats:
Would I be reading this right that the maximum memory this supports is 2GB-1?

Is this just an implementation limitation for the current spec tools, or has a decision been made that the size is interpreted as a signed number?

sunfishcode:
This patch is attempting to interpret the size as unsigned everywhere, and even corrects some code that previously interpreted it as signed. Did I miss any places?

sunfishcode:
Ah, looking at it again, I did miss some places. max_int is the max signed int. I'll submit a new PR to fix that.

PR #135, opened 2015-10-14
Second left-to-right test

pjuftring:
This tests if store instructions and calls evaluate their operands from left to right, too.

### DIFF
--- a/ml-proto/TestingTodo.md
+++ b/ml-proto/TestingTodo.md
@@ -22,7 +22,7 @@ Operator semantics:
  - ~~test that shifts don't mask their shift count. 32 is particularly nice to test.~~
  - test that `page_size` returns a power of 2
  - ~~test that arithmetic operands are evaluated left-to-right~~
- - test that call and store operands are evaluated left-to-right too
+ - ~~test that call and store operands are evaluated left-to-right too~~
  - ~~test that add/sub/mul/wrap/wrapping-store silently wrap on overflow~~
  - ~~test that sdiv/udiv/srem/urem trap on divide-by-zero~~
  - ~~test that sdiv traps on overflow~~
--- a/ml-proto/TestingTodo.md
+++ b/ml-proto/TestingTodo.md
@@ -23,6 +23,7 @@ Operator semantics:
  - test that `page_size` returns a power of 2
  - ~~test that arithmetic operands are evaluated left-to-right~~
  - ~~test that call and store operands are evaluated left-to-right too~~
+ - test that call and argument operands of call_indirect are evaluated left-to-right, too
  - ~~test that add/sub/mul/wrap/wrapping-store silently wrap on overflow~~
  - ~~test that sdiv/udiv/srem/urem trap on divide-by-zero~~
  - ~~test that sdiv traps on overflow~~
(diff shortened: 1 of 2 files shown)


ml-proto/test/left-to-right.wast
```diff
@@ -1,20 +1,22 @@
...
-  (func $get (result i32) (i32.load8_u (i32.const 0)))
-  (func $i32_left (result i32) (i32.store8 (i32.const 0) (i32.const 1)) (i32.const 127))
-  (func $i32_right (result i32) (i32.store8 (i32.const 0) (i32.const 2)) (i32.const 141))
-  (func $i64_left (result i64) (i32.store8 (i32.const 0) (i32.const 1)) (i64.const 163))
-  (func $i64_right (result i64) (i32.store8 (i32.const 0) (i32.const 2)) (i64.const 155))
-  (func $f32_left (result f32) (i32.store8 (i32.const 0) (i32.const 1)) (f32.const 124))
-  (func $f32_right (result f32) (i32.store8 (i32.const 0) (i32.const 2)) (f32.const 145))
-  (func $f64_left (result f64) (i32.store8 (i32.const 0) (i32.const 1)) (f64.const 163))
-  (func $f64_right (result f64) (i32.store8 (i32.const 0) (i32.const 2)) (f64.const 164))
+  (func $reset (i32.store8 (i32.const 8)(i32.const 0)))
```
sunfishcode: I see this was already present in the previous patch, but could you add a space between the ')' and '(' here?

ml-proto/test/left-to-right.wast
```diff
@@ -1,20 +1,22 @@
 (module
-  (memory 1)
+  (memory 9)
   
   ;; The idea is: We reset the memory, then the arithmetic instruction calls $*_left and $*_right.
   ;; $*_left stores the number 1, and $*_right stores the number 2 (both at address 0)
   ;; Then we read the value at address 0. If it's 0, the VM did an unwanted optimzation.
```
sunfishcode: This comment is now stale, since the loads and stores below use address 8. Also, is there a reason for switching from 0 to 8?

ml-proto/TestingTodo.md
```diff
@@ -22,7 +22,7 @@ Operator semantics:
  - ~~test that shifts don't mask their shift count. 32 is particularly nice to test.~~
  - test that `page_size` returns a power of 2
  - ~~test that arithmetic operands are evaluated left-to-right~~
- - test that call and store operands are evaluated left-to-right too
+ - ~~test that call and store operands are evaluated left-to-right too~~
```
sunfishcode: One additional test we can add here is to test that `call_indirect` evaluates its call operand before the argument operands. This is probably best done after #136 is in though, so feel free to just add it to TestingTodo.md for now.

sunfishcode:
Other than comments above, looks good to me.

ml-proto/test/left-to-right.wast
pjuftring: I needed some space for the store operations to store their data in. I use the first 8 bytes to do so and use the 9th one to store the order information.

ml-proto/test/left-to-right.wast
pjuftring: Oops, next time I'll scan my code for ')(', so hopefully this won't happen again.

sunfishcode:
lgtm, thanks!

PR #137, opened 2015-10-14
Experimental interpreter tracing code.

sunfishcode:
This PR isn't ready as-is; I'm just looking for feedback.

I've found this debugging feature, which which prints out all values of all operations as they are executed, to be quite handy. The patch here could be prettier, and it'd be nice to print the opcode too, but my main question is: This patch breaks the "tail call" property of eval_expr, which was mentioned a while ago as an important property; is there a convenient way to do what this patch is doing with doing that?

### DIFF
--- a/ml-proto/spec/eval.ml
+++ b/ml-proto/spec/eval.ml
@@ -131,6 +131,7 @@ let func_type instance f =
  *)
 
 let rec eval_expr (c : config) (e : expr) =
+  let result_value = (
   match e.it with
   | Nop ->
     None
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
Yes, tail calls are essential to the current formulation of the operational semantics. If you break it then e.g. a loop will no longer run in constant space but blow the stack. Other constructs like blocks and `if` would also eat up stack space linear in nesting depth.

The loop issue could be avoided by implementing it with a loop instead of a tail call, but of course that would no longer be declarative, i.e., "speccy". Maybe that's okay in this isolated case. The linearity in nesting depth could not be avoided, however.

PR #138, opened 2015-10-14
Fix i32 unsigned range tests.

sunfishcode:
When testing whether a value is in the i32 unsigned range, use the
unsigned max value (aka minus_one), rather than max_int, which is the
signed max value.


JSStats:
It appears Int64.of_int32 sign extends, so this might not be want is needed for wasm32.

Might it be better to spec the maximum index rather than memory size to support 4GB cleanly?

Is seems really confusing to be using the wasm storage types in the spec for general computations. Can the spec just use arbitrate precision integer maths.

sunfishcode:
Arg, good catch. I changed them to proper unsigned conversions now.

This caused some tests to expose the problem that the initial memory size is not rounded up to a page size. I've now added a patch which fixes that too.

The spec is using int64 as a "sufficiently large integer math" type, even when the index size is i32. In fact, the spec doesn't support wasm64 at all right now.

ml-proto/spec/eval.ml
```diff
@@ -109,7 +109,7 @@ let mem_size v at =
  * we currently only support i32, just test that.
  *)
 let mem_overflow x =
-  I64.gt_u x (Int64.of_int32 Int32.max_int)
+  I64.gt_u x (I64_convert.extend_u_i32 Int32.minus_one)
```
rossberg-chromium: Nit: `-1l` (here and below)

rossberg-chromium:
LGTM modulo nits

ml-proto/spec/eval.ml
```diff
@@ -314,7 +318,8 @@ let init m imports host =
   assert (host.page_size > 0L);
   assert (Lib.Int64.is_power_of_two host.page_size);
   let {memory; funcs; exports; tables; _} = m.it in
-  let memory' = Lib.Option.map init_memory memory in
+  (* Heads up! The following line curries the init_memory function! *)
```
lukewagner: I don't think this comment is necessary; we do this all over the place.

ml-proto/spec/eval.ml
sunfishcode: When I make that change, I get:

File "spec/eval.ml", line 117, characters 14-38:
Error: This expression has type I32.t -> I64.t
       but an expression was expected of type int

which I don't understand.

rossberg-chromium: Oh yes, it's interpreting the `-` as a binary operator. Need to put parens around the `-1l`.

sunfishcode:
Looking at this again, since OCaml doesn't actually support unsigned integers, and since we do a proper check for unsigned overflow in the `grow_memory` check, I don't think we need the main change that was being made here. I removed this from the branch, and now all that remains is just the trivial test function renaming, which I'll now merge.

PR #139, opened 2015-10-15
Makefile tweaks

sunfishcode:
This just makes some minor Makefile tweaks to make the Makefile a little friendlier.


rossberg-chromium:
lgtm

PR #141, opened 2015-10-15
Fix a copy+paste error in a test name, and add a new test for symbol names

sunfishcode:
Inspired by https://github.com/WebAssembly/design/issues/415 here is a PR that fixes the test name in question, and adds a new test for symbol name handling.

This tests a bunch of interesting names, but one thing it doesn't test is for newline characters embedded in names. That doesn't presently appear to with in the spec interpreter; I'll file a separate bug.

### DIFF
--- a/ml-proto/test/float_exprs.wast
+++ b/ml-proto/test/float_exprs.wast
@@ -272,29 +272,29 @@
 ;; Test that x/-0 is not folded away.
 
 (module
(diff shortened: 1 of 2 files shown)


rossberg-chromium:
lgtm

sunfishcode:
Rebased and trimmed tests that contain characters which are no longer accepted.

sunfishcode:
Merging with lgtm (from a while ago) above, and because I believe this is uncontroversial.

PR #144, opened 2015-10-16
Distinguish different categories of errors

rossberg-chromium:
This patch introduces different error exceptions to distinguish between static errors, traps, crashes, and assertion failures. It also does various other clean-ups around error handling.


ml-proto/spec/eval.ml
```diff
@@ -29,6 +27,36 @@ type instance =
 }
 
 
+(* Errors *)
+
+module Trap = Error.Make ()
+module Crash = Error.Make ()
+
+exception Trap = Trap.Error
+exception Crash = Crash.Error
```
sunfishcode: Can you add a comment explaining the difference between a Trap and a Crash?

ml-proto/spec/eval.ml
```diff
@@ -29,6 +27,36 @@ type instance =
...
+  | Memory.Bounds -> Trap.error at "out of bounds memory access"
+  | Memory.SizeOverflow -> Trap.error at "memory size overflow"
+  | exn -> raise exn
+
+let type_error at v t =
+  Crash.error at
+    ("type error, expected " ^ Types.string_of_value_type t ^
+      ", got " ^ Types.string_of_value_type (type_of v))
+
+let arithmetic_error at at1 at2 = function
```
sunfishcode: Can you add a comment documenting the meaning of at, at1, and at2?

ml-proto/host/builtins.ml
```diff
@@ -2,6 +2,9 @@ open Source
 open Types
 open Ast
 
+module Unknown = Error.Make ()
+exception Unknown = Unknown.Error
```
sunfishcode: Can you add a comment explaining what "Unknown" means in this context?

ml-proto/host/script.ml
```diff
@@ -21,35 +21,20 @@ type script = command list
 
 (* Execution *)
 
+module Syntax = Error.Make ()
+module AssertFailure = Error.Make ()
+
+exception Syntax = Syntax.Error
+exception AssertFailure = AssertFailure.Error
```
sunfishcode: Can you add a comment explaining that this AssertFailure is for failures in asserts in the .wast script, rather than assertions in the spec code itself?

ml-proto/spec/float.ml
```diff
@@ -50,7 +50,8 @@ sig
   val le : t -> t -> bool
   val gt : t -> t -> bool
   val ge : t -> t -> bool
-  val zero: t
+  val zero : t
+  val bare_nan : t
```
sunfishcode: Did something change in this patch to require bare_nan in the float interface? This value is an implementation detail -- and in fact, it's not actually a NaN pattern until some non-zero bits get or'd in -- so it'd be nice to keep its visibility limited if possible.

rossberg-chromium: Oops, this was an unintended remained of some earlier experiment. Removed.

rossberg-chromium:
Addressed comments, PTAL.

ml-proto/host/main.ml
```diff
@@ -22,11 +22,16 @@ let parse name source =
...
     {lexbuf.Lexing.lex_curr_p with Lexing.pos_fname = name};
-  try Parser.script Lexer.token lexbuf with Error.Error (region, s) ->
+  try Parser.script Lexer.token lexbuf with Script.Syntax (region, s) ->
     let region' = if region <> Source.no_region then region else
       {Source.left = Lexer.convert_pos lexbuf.Lexing.lex_start_p;
        Source.right = Lexer.convert_pos lexbuf.Lexing.lex_curr_p} in
-    raise (Error.Error (region', s))
+    raise (Script.Syntax (region', s))
+
+let error at category msg =
```
naturaltransformation: Would it make sense to make error take (at, msg) as a tuple and a category string to simplify the exception handling pattern match below?

rossberg-chromium: Could do that, but I don't think introducing a constructor indirection is worth it here. Makes all creation sites slightly more verbose and saves a little bit in this single place.

rossberg:
Friendly ping.

ml-proto/spec/eval.ml
```diff
@@ -294,18 +284,21 @@ and eval_hostop host mem hostop vs at =
...
-    if mem_overflow (Int64.add (Memory.size mem) delta) then
-      error at "runtime: grow_memory overflow";
+      Trap.error at "growing memory by non-multiple of page size";
+    let new_size = Int64.add (Memory.size mem) delta in
+    if I64.lt_u new_size (Memory.size mem) then
+      Trap.error at "memory size overflow";
+    (* Test whether the new size overflows the memory type.
+     * Since we currently only support i32, just test that. *)
+    if I64.gt_u new_size (Int64.of_int32 Int32.max_int) then
+      Trap.error at "growing memory out of bounds";
```
sunfishcode: "out of bounds" is a phrase we use to describe accesses that extend beyond the size. How about "memory size exceeded implementation limit"?

sunfishcode:
lgtm, with one wording nit above.

ml-proto/spec/eval.ml
rossberg-chromium: Good point, done.

rossberg-chromium:
Thanks! Wording nit addressed, merging.

PR #145, opened 2015-10-18
Tests for correct page size and float conversion

pjuftring:
This combines two little tests:
The first one tests that page_size is a power of two.
The second one converts both integer types to the corresponding floating point types (and back again) and tests that those conversions aren't folded away.


sunfishcode:
lgtm!

PR #147, opened 2015-10-20
refactor of int ops (i32/i64) into functor

naturaltransformation:
A simple refactor of the i32 and i64 operators into a single functor


rossberg-chromium:
lgtm

ml-proto/spec/int.ml
```diff
@@ -0,0 +1,197 @@
+(* WebAssembly-compatible i32 implementation *)
```
sunfishcode: This comment needs updating.

ml-proto/spec/int.ml
```diff
@@ -0,0 +1,197 @@
...
+    op (Rep.add x Rep.min_int) (Rep.add y Rep.min_int)
+
+  (*
+   * Unsigned division and remainder in terms of signed division; algorithm from
+   * Hacker's Delight, Second Edition, by Henry S. Warren, Jr., section 9-3
+   * "Unsigned Short Division from Signed Division".
+   *)
+  let divrem_u n d =
+    if d = Rep.zero then raise Numerics.IntegerDivideByZero else
+      let t = Rep.shift_right d (32 - 1) in
```
sunfishcode: This 32 - 1 was 64 - 1 in the original i64 code. I don't yet know if this means the algorithm works either way or if we need more tests.

naturaltransformation: Good catch. git rebase regression unfortunately.

naturaltransformation: Probably need another test case too. I can propose one in a subsequent PR if needed.

sunfishcode:
That works. Feel free to file a new issue for investigating the div_u/rem_u question and adding more tests if needed, or I can file it too, when I get a chance. Thanks!

PR #155, opened 2015-10-27
Implement a basic macro-assembler for test cases.

sunfishcode:
While WebAssembly's current AST has some high-level constructs like `if`, it doesn't have other obvious high-level control constructs such as `while` and `for`, making it difficult for humans to code by hand.

To address this concern, this PR introduces the start of a macro-assembler facility which allows tests to be written using high-level control constructs. So far, only `while` is implemented, but the basic idea is demonstrated: there is a fac.mast testcase which implements the common fibonacci function using a `while` loop, and the runtest.py script automatically runs the macro-assembler and generates a fac.generated.wast file, and then runs the test.

The macro-assembler preserves whitespace and comments in non-transformed parts of the code, so the output is fairly readable. With a modest amount of additional effort, it could be made to format the transformed parts in a pretty way as well.


pjuftring:
Awesome! Would it make sense to add a "replace"-macro, which replaces @ (or smth. else) with a list of strings? like this:
```
(replace ("foo" "bar")
  (invoke "@")
)
```
Gets translated to
```
(invoke "foo")
(invoke "bar")
```

sunfishcode:
Perhaps we could call it a "for each" macro or something. In any case, that kind of thing is certainly possible.

pjuftring:
Ok, I've written the basic implementation for this and will clean up the code (perhaps) tomorrow. When that's done, shall I make a PR for the branch? (I'm asking, because this would be a PR for a PR and therefore some kind of PR-ception, don't know, if this is intended use of PRs)

pjuftring:
I noticed a bug: If I rewrite a line in fac.mast from
```
    (while (i64.ne (get_local $i) (i64.const 0))
```
to
```
    (while
      (i64.ne (get_local $i) (i64.const 0))
```
It seems to add an "invisible" operand, which messes up the amount and position of operands, and the "while"-instruction doesn't get translated anymore.

sunfishcode:
Friendly ping.

ml-proto/sexp.py
```diff
@@ -0,0 +1,234 @@
...
+def isspace(c):
+    return c == ' ' or c == '\t' or c == '\n'
+
+def isdigit(c):
+    return c >= '0' and c <= '9'
+
+def isalnum(c):
+    return isdigit(c) or (c >= 'a' and c <= 'z') or (c >= 'A' and c <= 'Z')
+
+def skip_whitespace_and_comments(s, i):
```
pjuftring: This is the only function, in which you are checking if the index exceeds the length of the string. MASM still raises an IndexError, if the index is out-of-bounds in the other functions, but it could be hard to debug mast-files.

ml-proto/sexp.py
```diff
@@ -0,0 +1,234 @@
...
+def parse_number_literal(s, i):
+    j = i
+    while isalnum(s[j]) or s[j] in ['.', '+', '-']:
+        j += 1
+    t = s[i:j]
+    if t.find('.') == -1:
+        x = int(t)
+    elif t.startswith('0x'):
+        x = float(t)
+    else:
```
pjuftring: What's the purpose of this "else" ?

ml-proto/sexp.py
```diff
@@ -0,0 +1,234 @@
...
+
+def skip_whitespace_and_comments(s, i):
+    l = len(s)
+    while i < l:
+        if isspace(s[i]):
+            i += 1
+            # fixme
+            #while i < l and isspace(s[i]):
+                #i += 1
+        elif s[i] == ';' and i+1 < l and s[i+1] == ';':
```
pjuftring: I don't know if anybody uses them, but
```
(* This type of comment *)
```
Isn't handled by this code. They seem to work, but if I don't close a bracket:
```
(* (Like this *)
```
It raises an index error.

titzer: What about landing this in a separate repo? The concern here is to avoid
specifying a particular testing helper language as part of the official
definition of WebAssembly.

ml-proto/sexp.py
sunfishcode: An artifact of an earlier version of the code. I've now removed it.

sunfishcode:
We want to make it as easy as possible to write conformance tests, because a strong testsuite is one of the main tools we have for ensuring that implementations behave consistently. A builtin macro-assembler layer might make it significantly easier to write some kinds of tests.

I'm not super attached to the actual code right now; it's pretty primitive right now. I'm mainly looking for feedback on the idea.

rossberg-chromium:
I'm rather impartial to this, but I fear it might blur the boundary of what is spec'ed for people trying to learn about the details of the language by looking at the examples from tests. Also, do we actually expect to manually write large tests in the future, instead of generating them?

Should we discuss at the meeting?

sunfishcode:
Closing for now, since it's not presently a high priority. This idea, with or without the code here, can be brought back if there is need of it.

PR #157, opened 2015-10-31
Add support for "foreach" and "define" to masm

pjuftring:
This is an implementation for "foreach" and "define" as announced in #155 (Still not sure, if this is the right way to contribute to open PRs)
It works great, the only problem I encountered is, that in some cases (e.g. nesting) artifacts from comments mess up the generated code, but apart from that, everything works as intended.


sunfishcode:
I think the idea of a foreach macro makes sense. I'll take a closer look at this once we know what's happening with #155.

pjuftring:
Closing this PR since #155 is closed, too.

PR #159, opened 2015-11-03
Remove page_size

pjuftring:
Removes every trace of existence of page_size from the source, the
testfiles and README.md
https://github.com/WebAssembly/design/pull/442


ml-proto/host/params.ml
```diff
@@ -1,4 +1,4 @@
-let page_size = 4096L
+let page_size = 0x10000L (* 64 KiB *)
```
rossberg-chromium: Given that this is no longer a parameter, can you instead remove this definition completely? It should also vanish from the `host_params` type in the eval module.

sunfishcode:
Keeping page_size as a named variable improves the "specyness" of the code, and makes it easier to find all the places in the spec that case about page sizes. I would prefer to leave it in.

rossberg-chromium:
Fair enough. The only remaining occurrence is in the evaluation of GrowMemory. I suggest exporting page_size as a constant from the memory module and using that there.

sunfishcode:
lgtm

rossberg-chromium:
LGTM

PR #162, opened 2015-11-05
Renamed multiple definitions of exported function names

jcbeyler:


sunfishcode:
I realize that I've neglected #141 as of late, but I still think that as long as we have a multiple-modules-per-file setup, we should introduce new tests which specifically test that one can have multiple modules with functions with the same name when we fix the accidental occurrences of this.

jcbeyler:
Are you talking about multiple functions of the same name but not exported? Or are you talking about multiple functions that are then mangled because they come from different modules?

I' m open to doing whatever you like and helping out but I'd like to get this (or a form of) merged because it blocks my testing of the two test files :)

sunfishcode:
Ah. Sorry for being slow in #141; I had somehow thought that you were going to implement per-module name mangling or something and that it wasn't urgent. I've now updated and merged #141.

That said, consider yourself encouraged to make your implementation pass the new testcase too :-).

jcbeyler:
I see #141 so I was able to do: https://github.com/WebAssembly/wasm-to-llvm-prototype/commit/53392c3275e7df2f4136b688eae821ea998de4a6

But two things that are weird: names.wast seems not to have made in the mirror testsuite repo or is it me?

Second, it seems that I still have the issue for the exports.wast file that I fixed here too.

(Thanks for doing this :))

jcbeyler:
Though to be fair: because of your names.wast; I have to mangle the names a bit more (otherwise LLVM won't be happy). I'll probably end up mangling the names and adding the support to handle the exports.wast case; so we can probably go and do more interesting things :)

sunfishcode:
The testsuite repo update script missed the new file. This is fixed now.

lukewagner:
With #141 fixed, can we close this PR?  It seems like the changes in this PR to exports.wast shouldn't be needed since the reused names are in separate modules.

jcbeyler:
Yes I should be fixing it on my side with my name mangling, if I do have issues, I might come back but we can close this now

jcbeyler:
Just FYI: done now here: https://github.com/WebAssembly/wasm-to-llvm-prototype/commit/94f1fc6fa958a9735b16164cc13955962b8e3aae

Supporting names.wast and exports.wast :); next on my list is forward.wast!

PR #153, opened 2015-10-26
Implement tableswitch and flexible control flow

rossberg-chromium:
~~Bikeshedding:~~

~~- Currently still named `switch`, but we can rename it to be super-explicit if preferred.~~
~~- For consistency, new jumps are named `if_break`, `case_break`, `default_break`, but we can change that to `*_branch` or something if we decide to rename the existing `break` operator as well.~~

PR #163, opened 2015-11-05
Implement changes to control flow operators

rossberg-chromium:
Okay, here is a new implementation of the control operators (as an alternative to #153) that approximates the current description in the design doc, including the more verbose tableswitch. I managed to keep most of the complexity out of the kernel. The price to pay is that the desugaring no longer just consists of local expansions, it actually has to rewrite the case list to handle label targets. That is not ideal but probably bearable. The rest of the complexity is in the parser, where case labels get their own namespace with somewhat subtle scoping.

The abstract syntax of the jump table is more heavyweight than you envisioned, but don't say I didn't tell you so. ;)


rossberg-chromium:
Btw, here is an idea for slightly simplifying the notion of default: instead of making it a target separate from the jump table, why don't we simply say that the index operand is clamped to the size of the table? That would be equivalent, since you could then just make the default the last element of the table, but it removes one minor concept. WDYT?

rossberg-chromium:
See https://github.com/WebAssembly/spec/compare/control...control.clamp for the latter suggestion.

titzer:
On Thu, Nov 5, 2015 at 10:35 AM, rossberg-chromium <notifications@github.com
> wrote:

> Btw, here is an idea for slightly simplifying the notion of default:
> instead of making it a target separate from the jump table, why don't we
> simply say that the index operand is clamped to the size of the table? That
> would be equivalent, since you could then just make the default the last
> element of the table, but it removes one minor concept. WDYT?
>
Making it the last target of the table sounds fine to me, but I'd avoid
wording about "clamping" and just say directly that the last entry is the
default target.

ml-proto/README.md
```diff
@@ -157,9 +145,15 @@ expr:
   ( memory_size )
   ( grow_memory <expr> )
 
+switch:
+  ( table <target>* )
```
sunfishcode: Part of me was hoping we could use a LISP-style array literal, `#( <target>* )` or `[ <target>* ]` depending on the dialect, but I'm ok if we can't because of generic S-expression parsers.

ml-proto/spec/check.ml
```diff
@@ -39,6 +39,12 @@ let import c x = lookup "import" c.imports x
 let local c x = lookup "local" c.locals x
 let label c x = lookup "label" c.labels x
 
+module CaseSet = Set.Make(
+  struct
+    type t = I32.t option
+    let compare = Lib.Option.compare I32.compare_u
+  end)
```
sunfishcode: CaseSet appears to be unused now.

ml-proto/spec/int.ml
```diff
@@ -194,4 +197,7 @@ struct
   let to_string = Rep.to_string
 
   let of_int = Rep.of_int
+
+  let compare_s = compare
+  let compare_u x y = compare (Rep.add x Rep.min_int) (Rep.add y Rep.min_int)
```
sunfishcode: With CaseSet unused, these compare functions are unused too.

ml-proto/spec/int.ml
```diff
@@ -64,6 +64,9 @@ sig
   val ge_s : t -> t -> bool
   val ge_u : t -> t -> bool
 
+  val compare_s : t -> t -> int
+  val compare_u : t -> t -> int
```
sunfishcode: And these compare declarations.

ml-proto/given/lib.ml
```diff
@@ -50,6 +50,13 @@ struct
   let app f = function
     | Some x -> f x
     | None -> ()
+
+  let compare cmp_a o1 o2 =
+    match o1, o2 with
+    | None, None -> 0
+    | None, Some _ -> -1
+    | Some _, None -> 1
+    | Some x1, Some x2 -> cmp_a x1 x2
```
sunfishcode: This compare function appears unused as well.

ml-proto/given/lib.mli
```diff
@@ -20,6 +20,7 @@ sig
   val get : 'a option -> 'a -> 'a
   val map : ('a -> 'b) -> 'a option -> 'b option
   val app : ('a -> unit) -> 'a option -> unit
+  val compare : ('a -> 'a -> int) -> 'a option -> 'a option -> int
```
sunfishcode: And its type.

ml-proto/host/parser.mly
```diff
@@ -49,6 +49,11 @@ let literal s t =
     | Failure msg -> error s.at ("constant out of range: " ^ msg)
     | _ -> error s.at "constant out of range"
 
+let int32 s =
+  try I32.of_string s.it with
+    | Failure msg -> error s.at ("constant out of range: " ^ msg)
+    | _ -> error s.at "constant out of range"
```
sunfishcode: This int32 function appears unused now too.

ml-proto/test/fac.wast
```diff
@@ -27,14 +27,15 @@
     (set_local 2 (i64.const 1))
     (label
       (loop
-        (if
+        (if_else
           (i64.eq (get_local 1) (i64.const 0))
-          (break 0)
+          (br 1)
```
sunfishcode: Why did this change from 0 to 1?

sunfishcode:
This looks good to me at an initial readthrough, with some minor comments above. We can iterate from here.

I'm not partial to putting the default in the table. In my mental model, `tableswitch` desugars into first testing whether the index is out of bounds, and performing the jump table operation otherwise, so it feels more natural to keep the default separate. However, I agree that the semantics can work that way.

On that note, part of me would like to get rid of the default altogether and have `tableswitch` just trap if the index is out of bounds. It'd be great when one can prove no default is needed (eg. in a language with a strong type system). The main downside though is that if an application needs a default and does its own range check, and the tableswitch has its own builtin range check under the hood, we'd get two range checks. A clever JIT could optimize away the second by noticing it's covered by the first, or use virtual memory to make the second check "free", or other magic. However on today's hardware with a simple JIT, we'd get two range checks. I'm wondering if the part of me that wants this can convince the part that's concerned about simple JITs that it's worth doing though. If other people like it, perhaps we could explore the idea. What do you think?

ml-proto/test/fac.wast
rossberg-chromium: Because `loop` now became a jump target for its backward edge, whereas `label` used to be the only one.

rossberg-chromium:
@sunfishcode, removed dead code, thanks for finding!

rossberg-chromium:
Once this is landed, I'll create a separate PR for the "clamping" default, so we can have the discussion there.

As for a trapping switch, I'd like that even better. @titzer, do you think the scenario @sunfishcode mentions would be easy enough to optimise for something like v8-proto?

titzer:
Not sure that trapping tableswitch is a good idea. As @sunfishcode
mentions, that means an application does its own clamping to the default or
end of the table. The tableswitch basically looks like "switch(a >= size ?
size -1 : a)", i.e. a branch on the phi of a branch. Then the VM adds its
check before the switch, which is the output of the applications' check.
The only way to know the second check is in range is to do range analysis
(or a silly pattern match). If you want trap-on-out-of-bounds (which I
think will be rarer), it's easier to just make the default case have an
unconditional trap. You get the basically the same code for
trap-on-out-of-bounds and avoid needing the extra application-level
clamping on default-on-out-of-bounds.

On Thu, Nov 5, 2015 at 11:25 PM, rossberg-chromium <notifications@github.com

sunfishcode:
@titzer I don't think producers would use a clamp. LLVM, for example, will do something like this:

```
   if (a < size) {
      switch (a) {
      ...
      }
   } else {
      default_stuff_here();
   }
```

because this avoids routing the default path through the indirect branch, which is quite nice if the default path is hot, which it sometimes is in C/C++. (This is also related to why I'm not fond of thinking about the default being inside the table.)
 
If the JIT also lowers to this form instead of a clamp, we could end up with code effectively like this:

```
    if (a < size) {
       if (a < size) {
          switch (a) {
          ...
          }
       } else {
          trap();
       }
    } else {
       default_stuff_here();
    }
```

This way it's just a matter of recognizing that `a < size` is true on the true side of an `a < size` check.

rossberg-chromium:
So anybody not okay with landing this PR?

titzer:
lgtm

sunfishcode:
lgtm

dschuff:
Fixed up the conflicts with unreachable, merging based on lgtms and IRC with @sunfishcode

rossberg-chromium:
Thanks!

PR #167, opened 2015-11-07
No return value check for $print in address.wast

Teemperor:


rossberg-chromium:
This is a weaker and less complete check. The original form verifies that the function does indeed _not_ return anything (a.k. returns the empty value). That was intentional.

Teemperor:
Oh, I assumed the checks were added by accident because not even imports.wast tests for return values of imported functions. Would you accept a patch that moves/expands the tests to imports.wast? It makes IMHO more sense if an unexpected return value in a imported function would cause imports.wast to fail rather than address.wast. Also, imports.wast is already calling all existing importable functions so it's no big change to do the checks there.

rossberg-chromium:
Fine by me.

rossberg-chromium:
Oops, just rediscovered this one (hint: please ping after updating a PR). @Teemperor, there are conflicts now, would you mind resolving?

PR #170, opened 2015-11-10
Reword from "core language" to "internal kernel language".

sunfishcode:
This rewrites a paragraph of ml-proto's README.md to clear up a common confusion about the relationship between what before this patch is called the "core language" and the officially specified language.

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
 See `wasm -h` for (the few) options.
 
 
-## Core Language vs External Language
+## Official Language Specification vs Internal Kernel Language
 
-The implementation tries to separate the concern of what is the language (and its semantics) from what is its external encoding. In that spirit, the actual AST is regular and minimal, while certain abbreviations are considered "syntactic sugar" of an external representation optimised for compactness.
+The implementation serves as a language specification, so it aims to document the official language. At the same time, to simplify its internals, it lowers some constructs from the official language to produce a reduced "kernel" language which is simpler to validate (check.ml) and evaluate (eval.ml) in this implementation's chosen style. This is purely an implementation detail, and does not reflect any semantic property of the official language itself.
 
-For example, `if` always has an else-branch in the AST, but in the external format an else-less conditional is allowed as an abbreviation for one with `nop`. Similarly, blocks can sometimes be left implicit in sub-expressions.
+For example, `if` in the official language does not have an else arm, but in the internal kernel language it is given an else arm containing a `nop`. Similarly, blocks can sometimes be left implicit in sub-expressions in the official language, and are made explicit in the internal kernel language.
 
-Here, the external format is S-expressions, but similar considerations would apply to a binary encoding. That is, there would be codes for certain abbreviations, but these are just a matter of the encoding.
+Here, the specified language is represented as S-expressions, but similar considerations would apply to a binary encoding. That is, there would be codes which this specification would internally lower into other forms to simplify its own code.
 
 
-## Internal Syntax
+## Internal Kernel Language
 
-The core language is defined in [ast.ml](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/ast.ml).
+The internal kernel language is defined in [ast.ml](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/ast.ml).
 
-## External Syntax
+## Official S-expression Syntax
 
 The S-expression syntax is defined in [parser.mly](https://github.com/WebAssembly/spec/blob/master/ml-proto/host/parser.mly), the opcodes in [lexer.mll](https://github.com/WebAssembly/spec/blob/master/ml-proto/host/lexer.mll).
 
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -82,11 +82,11 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
 See `wasm -h` for (the few) options.
 
 
-## Official Language Specification vs Internal Kernel Language
+## WebAssembly Language Specification vs Internal Kernel Language
 
-The implementation serves as a language specification, so it aims to document the official language. At the same time, to simplify its internals, it lowers some constructs from the official language to produce a reduced "kernel" language which is simpler to validate (check.ml) and evaluate (eval.ml) in this implementation's chosen style. This is purely an implementation detail, and does not reflect any semantic property of the official language itself.
+This implementation serves as a specification and documentation of the WebAssembly language. At the same time, to simplify its internals, it internally lowers some WebAssembly language constructs to produce a reduced "kernel" language (ast.ml) which is simpler to validate (check.ml) and evaluate (eval.ml) in this implementation's chosen style. The distinction between WebAssembly and the internal kernel language is purely an implementation detail, and does not reflect any semantic property of WebAssembly itself.
 
-For example, `if` in the official language does not have an else arm, but in the internal kernel language it is given an else arm containing a `nop`. Similarly, blocks can sometimes be left implicit in sub-expressions in the official language, and are made explicit in the internal kernel language.
+For example, WebAssembly has separate `if` and `if_else` opcodes, but in the internal kernel language one construct handles both, using a `nop` to indicate the absence of an else arm. Similarly, the internal kernel language sometimes requires explicit blocks in places where they aren't required in WebAssembly itself.
 
 Here, the specified language is represented as S-expressions, but similar considerations would apply to a binary encoding. That is, there would be codes which this specification would internally lower into other forms to simplify its own code.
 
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
Can we simply call it "full" vs "kernel" language? That clarifies the crucial property that the kernel is a proper subset and not actually a different language, which is very much by design. It also avoids extra adjectives like "official" and "internal".

ml-proto/README.md
```diff
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
 See `wasm -h` for (the few) options.
 
 
-## Core Language vs External Language
+## Official Language Specification vs Internal Kernel Language
```
rossberg-chromium: How about "Full Language vs Kernel Language"? (I'm not sure how to contrast the "official language specification" with what is described here. Is there another language specification?)

ml-proto/README.md
```diff
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
...
 
-The implementation tries to separate the concern of what is the language (and its semantics) from what is its external encoding. In that spirit, the actual AST is regular and minimal, while certain abbreviations are considered "syntactic sugar" of an external representation optimised for compactness.
+The implementation serves as a language specification, so it aims to document the official language. At the same time, to simplify its internals, it lowers some constructs from the official language to produce a reduced "kernel" language which is simpler to validate (check.ml) and evaluate (eval.ml) in this implementation's chosen style. This is purely an implementation detail, and does not reflect any semantic property of the official language itself.
```
rossberg-chromium: Drop "in this implementation's chosen style", which sounds as if it was just an incidental consequence.
Also, I'm not sure what the last half-sentence is supposed to convey -- AFAICT it does "reflect semantic properties", which is the whole point?

ml-proto/README.md
```diff
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
...
 
-For example, `if` always has an else-branch in the AST, but in the external format an else-less conditional is allowed as an abbreviation for one with `nop`. Similarly, blocks can sometimes be left implicit in sub-expressions.
+For example, `if` in the official language does not have an else arm, but in the internal kernel language it is given an else arm containing a `nop`. Similarly, blocks can sometimes be left implicit in sub-expressions in the official language, and are made explicit in the internal kernel language.
```
rossberg-chromium: In the full language, we have both forms of conditional, so this is a bit confusing. How about:
"For example, conditionals can be expressed without an else-arm in the full language, but in the kernel language these are represented as an else containing a `nop`."

ml-proto/README.md
```diff
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
...
 
-Here, the external format is S-expressions, but similar considerations would apply to a binary encoding. That is, there would be codes for certain abbreviations, but these are just a matter of the encoding.
+Here, the specified language is represented as S-expressions, but similar considerations would apply to a binary encoding. That is, there would be codes which this specification would internally lower into other forms to simplify its own code.
 
 
-## Internal Syntax
+## Internal Kernel Language
```
rossberg-chromium: Just "Kernel Language"

ml-proto/README.md
```diff
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
...
 
-The core language is defined in [ast.ml](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/ast.ml).
+The internal kernel language is defined in [ast.ml](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/ast.ml).
```
rossberg-chromium: Dito.

ml-proto/README.md
```diff
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
...
 
-## External Syntax
+## Official S-expression Syntax
```
rossberg-chromium: Drop the "official" -- you most likely don't want to speak of "official S-expression syntax". ;)

ml-proto/README.md
```diff
@@ -82,20 +82,20 @@ Note however that the REPL currently is too dumb to allow multi-line input. :)
...
 
-For example, `if` always has an else-branch in the AST, but in the external format an else-less conditional is allowed as an abbreviation for one with `nop`. Similarly, blocks can sometimes be left implicit in sub-expressions.
+For example, `if` in the official language does not have an else arm, but in the internal kernel language it is given an else arm containing a `nop`. Similarly, blocks can sometimes be left implicit in sub-expressions in the official language, and are made explicit in the internal kernel language.
 
-Here, the external format is S-expressions, but similar considerations would apply to a binary encoding. That is, there would be codes for certain abbreviations, but these are just a matter of the encoding.
+Here, the specified language is represented as S-expressions, but similar considerations would apply to a binary encoding. That is, there would be codes which this specification would internally lower into other forms to simplify its own code.
```
rossberg-chromium: Can we drop the "to simplify its own code"?

sunfishcode:
I'll reword some of these to say "official" a little less.

However, my understanding is that the internal kernel language *is* an incidental consequence, with the exclusive purpose of simplifying this implementation's own code. It happens to be part of how this specification implements the language semantics, but it's not the only way the same semantics could be implemented. Other implementations don't need to know what's in the internal kernel language and what's not to be conforming.

rossberg-chromium:
Thanks. I don't disagree with the above. The comment about "incidental" was specific to the formulation "simpler ... in this implementation's chosen style", where of course that very style is not an accident but carefully chosen to enable such simplifications in the first place.

JSStats:
Why not just note that the Specification might describe one operation in terms of other operations, and that the reference implementation is exploring this by implementing some operations in terms of other operations. I don't think you even have a Specification at this point, and I don't think the reference implementation qualifies - it can't be advanced as wasm at the w3c. Has there ever been a web specification published as a reference implementation, and would it even cut it from a legal pointer of view?

kg:
Factoring into two layers like this makes sense from an engineering perspective, but as long as we have two layers mixed in the ```spec/``` directory it's going to be hard to maintain a clean divide between the two and it will be difficult for people to tell them apart. No amount of text we write will completely address this.

> The distinction between WebAssembly and the internal kernel language is purely an implementation detail, and does not reflect any semantic property of WebAssembly itself.

This is a very difficult sentence for me to interpret meaningfully. Both layers must be read and examined as a whole to understand the behavior of a given operation (like ```if_else```). The ```spec/``` directory can't really contain non-normative 'implementation details' because It's The Specification.

Describing this idea in text is a good start but if we're going to push forward with two layers, one of the layers needs to be split out of the spec directory (as a host-provided implementation of APIs, much like an external cstdlib), or we should spec both layers in detail. Doesn't the ES spec do this by defining lower-level operations like ```GetProperty``` or ```ReturnIfAbrupt``` and then defining higher-level language elements in terms of them? That seems to be what we are trying to do here, so we should just do it.

rossberg-chromium:
@kg, on Luke's request, I'm currently preparing a PR to make the full language more explicit in the spec dir, by introducing a separate type for the full language AST that then gets desugared into the smaller kernel AST. That should increase clarity about how full and kernel language relate.

That said, I agree that the quoted sentence leaves room for improvement (I still don't know what the part about "semantic properties" means). I'm in favour of the intended clarifications, but the wording generally sounds a bit like implying a false dichotomy and brushing off the kernel language as a random implementation artifact that has little to do with "WebAssembly" when it is in fact part of it.

@sunfishcode, I wouldn't mind if you tweaked the wording a little to make it less so (in particular, considering my suggestion of contrasting "kernel" and "full" language, instead of "kernel" and "WebAssembly"). But I can also do that myself in a follow-up, as long we are in agreement that it should be done. ;)

sunfishcode:
@rossberg-chromium In my background, there is an exuberance to distinguish between a language's spec and the language itself. As long as an implementation behaves such that by all observable accounts it is *as if* it were doing what's literally described in a language's spec, it is said to implement the language. In this tradition, out of an abundance of such exuberance, the factoring of concepts used within the language spec itself isn't considered to be part of the language, and an internal kernel language such as is used here is indeed be considered a random implementation artifact.

I pushed an update which rewrites the "semantic property" sentence to attempt to reflect the fact that the internal kernel language does indeed play a role in the spec, through its behavior.

I also don't find the kernel language to be a proper subset. It has factorings (Unary, Binary, etc.) which are its own, opcodes (Label, perhaps Nop) that are its own, and type abstractions (in Load, Select, etc.) that are its own. Consequently, I disagree that "kernel" versus "full" is an accurate portrayal.

rossberg-chromium:
Neither the kernel nor the full AST are "the language itself", both are abstract spec notions, aren't they?

Yes, the kernel AST decomposes the structure of the language a bit further. In particular, it decomposes the structure of operators themselves. The liberty to explicate structure is the reason for resorting to ASTs, though! The kernel AST still renders (a subset of) the same concrete language, and there is an immediate congruent bijection. That is, it's a different (more abstract) _AST_, but I very much disagree that it represents a different _language_, as I have been very careful not to make it so.

As for grouping operators, consider what you would do in a paper spec: you would use notation like

  expr_1 \odot expr_2

to range over various binary operations in one schema. The kernel AST is the natural rendering of that into a datatype. It's the kind of benign abstraction you do in every "spec" (and likewise in real implementations). It's not introducing "new opcodes", unless you somehow equate datatype constructors with opcodes.

But you have a point re labels, and in fact, as part of my full-AST patch I have been thinking about ways of getting rid of this node, for the very reason you mention. So, thanks for bringing it up.

I hope that explains why I am somewhat unhappy with the current wording in this patch, some of which, to my subjective taste, almost verges on being dismissive (which your later update hardly improved).

PR #172, opened 2015-11-11
Do not accept a too-large offset in memory accesses

### LINKED ISSUE #161: address.wast testcase constants
That testcase has

````
    (func $bad1 (param $i i32) (i32.load offset=4294967296 (get_local $i)))
````

and it tests that that traps on loads of 0 and 1.

1. The offset seems odd - that isn't a 32-bit unsigned. Is it internally represented as a larger number?
2. If it's a 32-bit unsigned, shouldn't that  first turn into a 0, and so there is no fault?

### PR DESCRIPTION
**Pull request by kripken**
They should be a uint32 value.

This fixes #161.

### DIFF
--- a/ml-proto/spec/check.ml
+++ b/ml-proto/spec/check.ml
@@ -259,6 +259,7 @@ and check_has_memory c at =
 
 and check_memop memop at =
   require (memop.offset >= 0L) at "negative offset";
+  require (memop.offset <= 0xffffffffL) at "offset too large";
   Lib.Option.app
     (fun a -> require (Lib.Int.is_power_of_two a) at "non-power-of-two alignment")
     memop.align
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ml-proto/spec/check.ml [hunk 1]
```diff
@@ -259,6 +259,7 @@ and check_has_memory c at =
 
 and check_memop memop at =
   require (memop.offset >= 0L) at "negative offset";
+  require (memop.offset <= 4294967295L) at "offset too large";
```
sunfishcode: This constant could use hex form to make it more obvious what value it is: 0xffffffffL.

sunfishcode:
lgtm, with minor nit above.

ml-proto/spec/check.ml
kripken: Done.

rossberg-chromium:
LGTM

PR #173, opened 2015-11-11
Remove the "nop" operator.

sunfishcode:
The design repo has no "nop", and with the separation of `if_else` from `if`, it no longer serves an obvious purpose.

### DIFF
--- a/ml-proto/host/lexer.mll
+++ b/ml-proto/host/lexer.mll
@@ -131,7 +131,6 @@ rule token = parse
   | "f32" { VALUE_TYPE Types.Float32Type }
   | "f64" { VALUE_TYPE Types.Float64Type }
 
-  | "nop" { NOP }
   | "block" { BLOCK }
   | "if" { IF }
   | "if_else" { IF_ELSE }
--- a/ml-proto/host/parser.mly
+++ b/ml-proto/host/parser.mly
@@ -159,7 +159,7 @@ let implicit_decl c t at =
 %}
 
 %token INT FLOAT TEXT VAR VALUE_TYPE LPAR RPAR
-%token NOP BLOCK IF IF_ELSE LOOP LABEL BR BR_IF TABLESWITCH CASE
+%token BLOCK IF IF_ELSE LOOP LABEL BR BR_IF TABLESWITCH CASE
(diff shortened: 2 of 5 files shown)


rossberg-chromium:
LGTM. I'll create follow-up to remove it from the kernel as well.

@titzer, I originally copied this from v8-proto, which still seems to have Nop.

titzer:
Nop is extremely useful, e.g. because br and br_if take a value.

On Wed, Nov 11, 2015 at 9:39 AM, rossberg-chromium <notifications@github.com

sunfishcode:
I was assuming that either:

  - `br` and `br_if` have some way of knowing how many result values they have (and thus how many operands they have), either from information embedded in themselves or inferred from where they're branching to

or

 - If the value isn't used at destination, it doesn't matter what value is returned, so we can just use the value of whatever statement would have preceded the `br` or `br_if`.

depending on how we represent statements=expressions.

titzer:
On Wed, Nov 11, 2015 at 9:59 AM, Dan Gohman <notifications@github.com>
wrote:

> I was assuming that either:
>
>    - br and br_if have some way of knowing how many result values they
>    have (and thus how many operands they have), either from information
>    embedded in themselves or inferred from where they're branching to
>
> That would only be possible if blocks declare their types, or at least
their arity, up front.

>
>
> or
>
>    - If the value isn't used at destination, it doesn't matter what value
>    is returned, so we can just use the value of whatever statement would have
>    preceded the br or br_if.
>
> In v8-native, br and br_if take an additional operand which represents the
value to produce for the labeled block or loop when breaking. That makes
sense for breaking out of a block, but doesn't make sense for continuing
back to a loop. Implicitly returning a value from the previous statement
doesn't really work.

kripken:
A minor nice thing about nop is that it makes it easy to write wasm=>wasm transformation passes, as it gives an easy way to remove code.

kg:
```nop``` is an absolute requirement for writing wasm->wasm transforms, debugging filters, etc.

sunfishcode:
I can see how nop might be an absolute requirement and/or a nice thing if the binary encoding is designed in a certain way. However, that's not the only possible way. It's now clear that anything that is done here will have to be done in conjunction with a specific binary encoding proposal.

PR #174, opened 2015-11-13
Introduce explicit type for full AST

rossberg-chromium:
Step 1 towards explicitly representing full operator syntax as given in AstSemantics.

This mainly renames some modules and introduces a new type for the full AST, currently copying the structure of the kernel one. It does not yet flatten it into individual opcodes -- I split that into a follow-up patch, since it involves massive changes to lexer and parser.


ml-proto/spec/ast.ml
```diff
@@ -2,148 +2,67 @@
...
+  | Load of Kernel.memop * expr
+  | Store of Kernel.memop * expr * expr
+  | Load_extend of Kernel.extop * expr
+  | Store_wrap of Kernel.wrapop * expr * expr
+  | Const of Kernel.literal
+  | Unary of Kernel.unop * expr
+  | Binary of Kernel.binop * expr * expr
+  | Select of Kernel.selop * expr * expr * expr
+  | Compare of Kernel.relop * expr * expr
+  | Convert of Kernel.cvt * expr
```
sunfishcode: Would it be possible to move memop, literal, unop, binop, etc. out of Kernel and into Ast, and have Kernel import then from Ast instead? Since the Ast module is the thing that has the 1-1 correspondence with the design repo, it would be nice to have all the fields described in the design repo here, in one place.

ml-proto/spec/kernel.ml
```diff
@@ -0,0 +1,149 @@
+(*
+ * (c) 2015 Andreas Rossberg
+ *)
```
sunfishcode: One extra thing for #175.

ml-proto/spec/ast.ml
rossberg-chromium: This dependency will disappear with the flattened AST in the follow-up patch anyway, so is it worth worrying about?

ml-proto/spec/kernel.ml
rossberg-chromium: Done.

ml-proto/spec/ast.ml
sunfishcode: Ah, this sounds fine then.

sunfishcode:
lgtm, as a first step with others to follow.

PR #176, opened 2015-11-13
Remove the AUTHORS file and revert to a standard LICENSE.

sunfishcode:
The AUTHORS file claims to be a list of people and their organizations who have contributed to the repository, but it isn't complete in that it doesn't list many of the people who have contributed, and it doesn't list the organizations of the people it does list. 

And, the LICENSE file was modified from the original Apache text in an apparent attempt to connect it with the AUTHORS file, however the change was made in the appendix section, which only contains suggested boilerplate text for use with the license, and we aren't currently using the boilerplate.

This PR proposes to rectify these issues by reverting the LICENSE to the original text and removing the AUTHORS file for now. If there is a future need for an AUTHORS file, we can start a new one, in a deliberate way.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,8 +1,0 @@
-# Below is a list of people and organizations that have contributed
-# to the project.  Names should be added to the list like so:
-#
-#   Name/Organization <email address>
-
-Andreas Rossberg <rossberg@chromium.org>
-Katelyn Gadd <kg@luminance.org>
-
--- a/LICENSE
+++ b/LICENSE
@@ -186,8 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright (c) 2015 repository authors, see AUTHORS file.
-
+   Copyright {yyyy} {name of copyright owner}
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.


lukewagner:
Yes, as a spec it's weird to have any notion of copyright associated with it.  Looking at what the WHATWG does, it seems the license is all CC0 1.0 Universal (when there is a LICENSE at all) [[1](https://github.com/whatwg/compat/blob/master/LICENSE), [2](https://github.com/whatwg/mediasession/blob/master/LICENSE), [3](https://github.com/whatwg/storage/blob/master/LICENSE)].  That avoids the awkwardness of the copyright line in the current LICENSE file.  Can we just do that?

jfbastien:
Eek, yes.

Along the lines of this and #175, the source code should follow what the LICENSE file says, and contain:

> Copyright {yyyy} {name of copyright owner}

rossberg-chromium:
Complete license noob here, so fine with anything. FWIW, I just copied this from https://github.com/WebAssembly/polyfill-prototype-1/blob/master/LICENSE, so somebody may need to fix that as well. :)

sunfishcode:
Merging, as this seems like an uncontroversial first step.

@jfbastien With respect to the LICENSE strategy, another option is to do [what Luke suggested](https://github.com/WebAssembly/spec/pull/176#issuecomment-156506618). Either way, I think it's a topic for a separate issue/PR.

PR #178, opened 2015-11-17
Removed white spaces at end of lines

jcbeyler:
Not sure we care about this but when I try to remove some from my repo, the test suite has a lot of them and then it makes it more tricky to find. So thought I would offer this in case we do care (other projects I work on do care :))


pjuftring:
Woops, didn't see them (for obvious reasons ;) )

rossberg-chromium:
LGTM, thanks!

Issue #179, opened 2015-11-18
Break expression handling


jcbeyler:
Dear all,

As I was starting to handle more complex labeling, I was looking at the labels.wast case:

```
  (func $loop2 (result i32)
    (local $i i32)
    (set_local $i (i32.const 0))
    (loop $exit $cont
      (set_local $i (i32.add (get_local $i) (i32.const 1)))
      (if (i32.eq (get_local $i) (i32.const 5))
        (br $cont (i32.const -1))
      )
      (if (i32.eq (get_local $i) (i32.const 8))
        (br $exit (get_local $i))
      )
      (set_local $i (i32.add (get_local $i) (i32.const 1)))
      (br $cont)
    )
  )
```

I am especially interested in this break:
        (br $cont (i32.const -1))

Is the idea behind this to say:
  - If I do branch back to the loop, execute the code of the expression before jumping (well here it would get DCE'd anyway but still)
     - So for example, if it was (unreachable) we would trap there instead of DCE-ing it

  - Even though I know no one will consume it afterwards (compared to a branch to an exit for example)

I read the design and it wasn't clear if this was or not the behavior that we were going for and I'd prefer checking.

Thanks!
Jc

kg:
> execute the code of the expression before jumping

This brings to mind a related question: Is ```(br $label (br $otherlabel (i32.const -1)))``` valid? If the child expression is evaluated before the jump, then it would jump to otherlabel and then maybe jump to label. If it's evaluated after, it would jump to label and then maybe jump to otherlabel. Whether this actually works would depend on the nesting, and the evaluation order would make this either correct or incorrect.

```(br $label1 (block (br $label2 (i32.const 0)) (br $label3 (i32.const 1)))```
???

We should make sure to clearly specify the behavior of br and br_if in detail so that it's unambiguous how all this works. I don't see much detail in the design repo.

**Comment by kg** [comment]
Should've thought of this before hitting comment:
Break targets are encoded as nesting distances, so if break-inside-break is evaluated in the wrong order it could actually end up jumping to an unexpected location.
I can also imagine an implementation of break-within-break where the inner break simply causes the outer break operation to never execute.

**Comment by rossberg-chromium** [comment]
@jcbeyler: Hah, I think you just found a bug. The extra argument should neither be necessary nor allowed. I'll look into it.

@kg: yes, that is legal, though obviously not a particularly useful example. Conceptually, it's no different from something like
```
throw (() => { throw -1 ))()
```
[...]

PR #180, opened 2015-11-18
Tighter typing of Break

rossberg-chromium:
Disallows redundant arguments to a break targeting a label of type `None` (addresses #179).

Conceptually, there is no specific harm in allowing this, but it also is not useful. Rather be consistent with the checking for calls.

### DIFF
--- a/ml-proto/test/labels.wast
+++ b/ml-proto/test/labels.wast
@@ -78,3 +78,4 @@
 (assert_return (invoke "return" (i32.const 1)) (i32.const 2))
 (assert_return (invoke "return" (i32.const 2)) (i32.const 2))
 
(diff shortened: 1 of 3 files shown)


jcbeyler:
Looks good to me :)

ml-proto/test/labels.wast
```diff
@@ -24,7 +24,7 @@
     (loop $exit $cont
       (set_local $i (i32.add (get_local $i) (i32.const 1)))
       (if (i32.eq (get_local $i) (i32.const 5))
-        (br $cont (i32.const -1))
+        (br $cont)
```
lukewagner: Perhaps also add an `assert_invalid` test?

rossberg-chromium: Done.

lukewagner:
lgtm

PR #181, opened 2015-11-18
Obfuscating Loop1

jcbeyler:
The loop1 method is too simple for LLVM, which then makes it return just 5.
Though this proves that the code is correct, we probably want to have the
  code generation all the same.

This patch makes the code a bit different but at least the compiler can no
  longer optimize it all away.

### DIFF
--- a/ml-proto/test/labels.wast
+++ b/ml-proto/test/labels.wast
@@ -78,3 +78,4 @@
 (assert_return (invoke "return" (i32.const 1)) (i32.const 2))
 (assert_return (invoke "return" (i32.const 2)) (i32.const 2))
 
+(assert_invalid (module (func (loop $l (br $l (i32.const 0))))) "arity mismatch")
--- a/ml-proto/test/labels.wast
+++ b/ml-proto/test/labels.wast
@@ -64,14 +64,14 @@
     (block $outer
       (block $inner
         (br_if (i32.const 0) $inner)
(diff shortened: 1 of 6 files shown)


ml-proto/test/labels.wast
```diff
@@ -6,12 +6,12 @@
...
 
-  (func $loop1 (result i32)
+  (func $loop1 (param $max i32) (result i32)
     (local $i i32)
-    (set_local $i (i32.const 0))
+    (set_local $i (i32.const 1))
     (loop $exit $cont
-      (set_local $i (i32.add (get_local $i) (i32.const 1)))
-      (if (i32.eq (get_local $i) (i32.const 5))
+      (set_local $i (i32.add (get_local $i) (get_local $i)))
```
rossberg-chromium: This is doubling $i on each iteration, so the expected result would be something like 2^8, no?

rossberg-chromium: Ah, sorry, misread break condition.

rossberg-chromium:
LGTM

sunfishcode:
What would you think about just adding the new more complex loop as a new test, rather than replacing the existing test? I think there's value in both.

jcbeyler:
Done, I added a loop3 then

jcbeyler:
let me do this again...

PR #184, opened 2015-11-19
Allow optional argument to br_if

rossberg-chromium:
Matching up with v8-prototype and https://github.com/WebAssembly/design/pull/464 .

Also, add simple test for br_if.


titzer:
lgtm

AndrewScheidecker:
The way this is implemented, the argument must not be evaluated unless the condition is true. AFAICT that means it must be implemented with two hardware branches, contrary to the motivation for `br_if`. It seems strange, but maybe the argument should always be evaluated?

titzer:
In v8-native, the argument is indeed always evaluated.

rossberg-chromium:
@AndrewScheidecker: good point, refined accordingly in #188.

PR #189, opened 2015-11-21
Creating an obfuscated Loop1

jcbeyler:
The loop1 method is too simple for LLVM, which then makes it return just 5.
Though this proves that the code is correct, we probably want to have the
  code generation all the same.

This patch makes a loop3 method based on loop1 so that the code ais  bit
  different but at least the compiler can no longer optimize it all away.

We've left loop1 to show the difference in compiler technologies

### DIFF
--- a/ml-proto/test/labels.wast
+++ b/ml-proto/test/labels.wast
@@ -46,6 +46,18 @@
     )
   )
 
+  (func $loop4 (param $max i32) (result i32)
+    (local $i i32)
+    (set_local $i (i32.const 1))
+    (loop $exit $cont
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
LGTM

sunfishcode:
lgtm

sunfishcode:
This has LGTMs, and is now just waiting on a merge conflict resolution.

jcbeyler:
Fixed the merge conflict

sunfishcode:
Great! Carrying forward lgtms and merging.

PR #190, opened 2015-11-21
Make label indexing independent of whether loop and switch have explicit label names

AndrewScheidecker:
Without this change, the label indices are different depending on whether the `loop` and `switch` nodes have named labels (and how many labels for `loop`).

Since the label names are syntax sugar, I assume the this change is closer to the desired binary format semantics, as well as being generally easier to understand.

### DIFF
--- a/ml-proto/host/parser.mly
+++ b/ml-proto/host/parser.mly
@@ -229,8 +229,8 @@ bind_var :
 ;
 
 labeling :
-  | /* empty */ %prec LOW { let at = at () in fun c -> c, Unlabelled @@ at }
-  | bind_var { let at = at () in fun c -> bind_label c $1, Labelled @@ at }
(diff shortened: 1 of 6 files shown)


rossberg-chromium:
I suspect there are some problems with this patch, e.g. the label tracking for `loop` seems suspicious. You also shouldn't need to keep around the `labeling` type.

But more importantly, the change is subsumed by a larger patch for flattening out the AST that I am about to upload, and which removes `label` altogether. Can you hold on for a little while? :)

rossberg-chromium:
See #193.

AndrewScheidecker:
Ok, I'll leave this PR open, but I concur that it's subsumed by #193, and should be closed if that PR is merged.

AndrewScheidecker:
Closing since #193 was merged.

PR #191, opened 2015-11-21
Wrap the tableswitch default target in a (default ...) node

AndrewScheidecker:
Wrap the tableswitch default target in a (default ...) node so that something like (default (case x)) stands out from its (case ...) siblings that define targets rather than references to targets.

Before:
```
(tableswitch (i32.const 1)
    (table (case $0))
    (case $default)
    (case $default ...)
    (case $0 ...)
)
```

After:
```
(tableswitch (i32.const 1)
    (table (case $0))
    (default (case $default))
    (case $default ...)
    (case $0 ...)
)
```


kg:
It would be much better if this were a part of the case node instead of a new imaginary (i.e. not actually a functional part of the AST or encoded into binary) s-expression node that adds extra nesting.

We have something like ```align=N``` for loads and stores, so how about ```(case default $label)``` vs the regular ```(case $label)```? Alternately, you introduce a ```(default-case $label)``` opcode.

AndrewScheidecker:
I went with `(default ...)` to match the `(table ...)` node preceding it in the `tableswitch` syntax.

> Alternately, you introduce a (default-case $label) opcode.

Then you'd also need `default-br`, since `tableswitch` supports a non-case label as the default, distinguished by `(default (br ...))` (using this commit's syntax). Maybe finding a way to avoid that distinction would be best, though.

rossberg-chromium:
I agree the syntax of the default target can be slightly irritating, but is it really a problem serious enough to justify adding extra syntactic baggage?

If we think so, then I'd rather revive the earlier suggestion of simply including the default in the table, and define dispatch by clamping the index value. See https://github.com/WebAssembly/spec/pull/163#issuecomment-154147226 and subsequent discussion. At least that would make the syntax simpler, not more complicated, though I also understand Dan's reservations.

Another option would be to use a different keyword for the `case` in the table/default, but I don't have a good suggestion.

sunfishcode:
Another alternative is to remove the default from `tableswitch` altogether. A proposal to do that is now at https://github.com/WebAssembly/design/pull/482 .

rossberg-chromium:
Obsolete with br_table, closing.

PR #192, opened 2015-11-22
Add a test for a loop without a trailing br

kripken:
All loops in the test suite have a trailing `br`, so we didn't have a test yet for loops not automatically looping. I noticed this on binaryen which had this wrong.


rossberg-chromium:
LGTM

PR #193, opened 2015-11-23
Flatten AST into concrete operators

rossberg-chromium:
This patch flattens out the representation of the full AST into individual operators with names as given in the design doc. It also removes the administrative `label` node in both full and kernel AST. Also various updates to the README, which had gotten out of date lately.

This flattening is quite messy tbh, and not something you'd want to do in a real implementation. Essentially, the lexer first recognises the operators with their inner structure, then the parser throws it away when building the flat AST, just to be reconstructed in the next pass when desugaring into the kernel. As a consequence, there is an awful amount of repetition and code duplication now.

I guess that's because the boundaries between concrete and abstract syntax get a bit fuzzy here. Arguably, the full AST is fairly concrete in its prescription of operators being designated by flat unstructured names that we want to match concretely. :)


lukewagner:
I haven't had time to look it over carefully, but at a first glance this looks great and a good basis for defining a `to_binary` function.  Thanks!

JSStats:
It looks problematic having operators with optional arguments follows but other optional arguments. For example `block`, `loop`, etc. I don't expect the binary encoding would follow this, rather just support a fixed number of arguments with an optional a variable length tail. So might it be better to split the operator names in the s-exp format to make this clear, or use a filler for unused arguments.

rossberg-chromium:
@lukewagner, well, the flattening is not necessary for defining `to_binary`, it merely makes it slightly more immediate (while increasing complexity everywhere else). As you can tell from the patch description, I'm still having my doubts that this is the right direction; it doesn't mirror what you'd do in a real implementation _or_ in a paper spec (where you'd employ notational conventions that mirror roughly what the previous AST rendered).

@JSStats, re optional arguments, are you referring to label bindings? Then note that these are just concrete syntax convenience of the text format that is not part of the actual AST or the binary format. The "raw" AST has no explicit label binders to name nodes, and all label references are simply relative numeric indices. I added explicit notes about this to the README.

lukewagner:
@rossberg-chromium I'm happy to sacrifice some boilerplate to be able to simply point to an AST which has an obvious correspondence to AstSemantics.md.  That being said, I was thinking that a reasonable alternative that achieves this goal is to treat what is currently the `Kernel` AST as the "representation type", and instead have something like:
```
type Expr = { op : Op; rep : Rep } and
type Op = { I32_add, F32_add, ... } and
type Rep = UnaryOp of Expr | BinOp of Expr * Expr | ...
```
Such that there was this clear delineation of (AstSemantics.md-defined) node vs. shared representation for the benefit code reuse.  So this would mean:
* No kernel vs. concrete, `desugar.ml` goes away. \o/
* We'd still want a 1:1 correspondence between `Expr`s and AstSemantics.md-defined nodes, so we'd still want the parts of this PR that remove wholly-kernel nodes like `Label`.
* Even though it'd be a bit more boilerplate, I think, for the purpose of specifying the precise `op`:`rep` correspondence, it'd be clarifying if `check.ml` `match`ed on `op` and then funneled each case down to a `check_*` for each `Rep` type (failing if the `Rep` mismatched), rather than simply `match`ing on `rep`.

WDYT?

rossberg-chromium:
@lukewagner, wouldn't this defeat most of the advantage of having a typed AST? You'd no longer ensure by construction that each op has the right number of operands, and thus would allow a lot of random garbage into the representation. To me that sounds like a big step backwards. (You could play tricks with GADTs, but that seems unjustified, given that there is a simpler, more canonical way to do it.)

Also, why would that eliminate the full vs kernel distinction? AFAICT, the kernel would still have a smaller operator type in this representation.

lukewagner:
@rossberg-chromium Maybe I didn't explain well since it sounds like you're thinking about something different.  In what I described above, the only extra thing that you can get wrong is having `op` not match `rep`; once these match (as checked by `check.ml`) everything else would be equivalently statically-typed.

rossberg-chromium:
@lukewagner, maybe I still misunderstand, but it sounds like you are suggesting to extend type checking with syntax checking? Because the types you sketch above would allow representing _syntactically_ invalid programs. Those cannot possibly be the result of decoding a binary, however, so it would be sad if the validator had to worry about them.

lukewagner:
@rossberg-chromium Ok, fair enough, just a thought.  I realized another thing I like about the ast.ml in this patch is that it makes it clear what is a "real" immediate and what's derived from the operator name.  That, no synthesized kernel nodes, and completely explicit enumeration all seem worth it to me.

ml-proto/spec/kernel.ml
```diff
@@ -50,16 +50,16 @@ struct
...
 end
 
-module Int32Op = IntOp ()
-module Int64Op = IntOp ()
-module Float32Op = FloatOp ()
-module Float64Op = FloatOp ()
+module I32Op = IntOp ()
+module I64Op = IntOp ()
+module F32Op = FloatOp ()
+module F64Op = FloatOp ()
```
sunfishcode: I'm having difficulty understanding how this code works. How would you describe what `I32Op` is here?

rossberg-chromium: Well, it's a module defining the set of i32 operators. It's created from a generic module defining integer operators. But there is nothing new here, I just renamed them to match the Wasm type names better.

rossberg-chromium:
Ping. Anybody feels like reviewing this?

ml-proto/host/lexer.mll
```diff
@@ -145,98 +154,181 @@ rule token = parse
...
+    { BINARY (fun (e1, e2) ->
+        floatop t (F32_copysign (e1, e2)) (F64_copysign (e1, e2))) }
+
+  | (nxx as t)".select"
+    { SELECT (fun (e1, e2, e3) ->
+        numop t (I32_select (e1, e2, e3)) (I64_select (e1, e2, e3))
+                (F32_select (e1, e2, e3)) (F64_select (e1, e2, e3))) }
+
+  | (ixx as t)".eq"
+    { COMPARE (fun (e1, e2) -> intop t (I32_eq (e1, e2)) (I64_eq (e1, e2))) }
```
kg: This is really noisy, is there any way you could factor it out? The actual opcode name is obscured in the noise.

i.e. ```COMPARE intop_2arg I32_ne I64_ne (ixx as t)".lt_s"``` or something better than that

kg:
Can you split the removal of labels out into a separate PR? We need to discuss that specifically, I think.

rossberg-chromium:
@kg, do you mind discussing labels here? I probably could separate them out, but it would be quite a bit of (mostly redundant) work at this point.

ml-proto/host/lexer.mll
rossberg-chromium: I'd love to make all this niftier, but I don't know how. It would be nicer if we had curried constructors in OCaml (then it'd be just `COMPARE (intop t I32_eq I64_eq)`), but we don't.

kg:
> @kg, do you mind discussing labels here? I probably could separate them out, but it would be quite a bit of (mostly redundant) work at this point.

So, it's not clear to me at this point: What purpose do labels serve, specifically? As I understand it, br and other opcodes all target an enclosing block by nesting depth, not a specific label by name. So is that why labels are gone entirely?

How does the removal of labels interact with tableswitch? Do all of the individual switch targets end up being enclosing scopes with br?

Can you make the new parser/lexer code that handles labels clearer? It's basically indecipherable to me, and this seems like an important behavioral detail that isn't just grammar.

rossberg-chromium:
>
> So, it's not clear to me at this point: What purpose do labels serve,
> specifically? As I understand it, br and other opcodes all target an
> enclosing block by nesting depth, not a specific label by name. So is that
> why labels are gone entirely?
>

Yes. (More precisely, targets are block, loop, or tableswitch.)


> How does the removal of labels interact with tableswitch? Do all of the
> individual switch targets end up being enclosing scopes with br?
>

Hm, nothing should have changed regarding tableswitch. The only observable
change of this patch is that the `label` pseudo operator is gone.


> Can you make the new parser/lexer code that handles labels clearer? It's
> basically indecipherable to me, and this seems like an important behavioral
> detail that isn't just grammar.
>

The only way it could be made clearer, I suppose, is by introducing yet
another (full) intermediate representation with explicit symbolic labels,
and another translation step on the host side. I would like to avoid that
complication if possible, since all the symbolic naming is just syntactic
convenience anyway, not part of the "official" AST.

I would argue that the current code isn't all that complicated. The main
trick to notice is that each semantic action is parameterised over a
context `c`, which contains the label environment (and others), and which
is passed on to children and updated in the right places. That's why you
have all the `fun c -> ...$1 c...$2 c...`, and a couple of let-bindings for
updating `c` to `c'` locally.

sunfishcode:
lgtm

rossberg-chromium:
Thanks! Landing with Dan's LGTM above and Luke's off-line consent.

PR #195, opened 2015-11-25
Generate a dumb batch file for Cygwin-less build on Windows

rossberg-chromium:
Include `winmake.bat` (generated from ocamlbuild's output), which can be run in a Windows shell to build the interpreter without requiring Cygwin.

Also, update the build instructions in the README. I have tested both the Cygwin-less and Cygwin-powered instructions on a bare Windows 10 and a bare Windows Vista machine, and they work without problems.


kg:
Can you be clearer about the 'If you want to build a native code executable, or want to hack on the interpreter' part? It's unclear to me why I need cygwin to hack on the interpreter but don't need it to compile. Do the new instructions work without cygwin because we switched to non-opt builds? If we switch back or add opt builds to some part of our test suite, will people need to install cygwin retroactively?

Also, I assume ```runtests.py``` works sans-cygwin on Windows with these instructions, but did you check? If not, I can test that out next week.

rossberg-chromium:
@kg, without Cygwin you have neither `make` nor `ocamlbuild`, so no incremental compilation (and unfortunately, bare Windows has no other build system either). I added a clarifying parenthesis. There is no problem with invoking the (byte code) compiler directly, though, which is what the batch file does.

Native code compilation requires Cygwin for its assembler and linker tool chain. The default target picked in the Makefile is unrelated to what the new batch file does, however.

Regarding the test runner, there already is a sentence saying that you need Python installed for that -- that is orthogonal to how you install OCaml or build the interpreter (up to using the --build option, of course, which calls back to `make`). I added an additional sentence pointing out that you can invoke `runtests.py` directly.

rossberg-chromium:
Test runner works sans cygwin, but I had to drop the `.exe` extension from the wasm executable to make the Python happy.

rossberg-chromium:
Ping. Anybody wants to LGTM?

ml-proto/Makefile
```diff
@@ -10,7 +10,6 @@ NAME_UNOPT =    $(NAME)
 DIRS =		given host spec
 LIBS =		str bigarray
 
-OCB_FLAGS += 	-quiet
```
kg: Why did -quiet go away?

ml-proto/Makefile
```diff
@@ -21,7 +20,7 @@ OCB =		ocamlbuild $(OCB_FLAGS)
 .PHONY:		all opt unopt
 unopt:		$(NAME_UNOPT)
 opt:		$(NAME_OPT)
-all:		opt unopt test
+all:		winmake.bat opt unopt test
```
kg: What constrains this to only invoking winmake on Windows? I don't see anything that does

ml-proto/winmake.bat
```diff
@@ -0,0 +1,68 @@
+rem Auto-generated from Makefile!
```
kg: If this is auto-generated but we're checking it in, we should make sure to clearly documented that people should rerun make and check in the .bat file every time

ml-proto/Makefile
rossberg-chromium: Was moved to the individual targets below, since it must not be used for the run creating the batch file.

ml-proto/Makefile
rossberg-chromium: And it shouldn't! Any patch needs to regenerate the batch file as necessary, regardless of dev platform.

ml-proto/winmake.bat
rossberg-chromium: Yes, good point. Added.

rossberg-chromium:
Friendly ping. Anybody opposed to landing this? It's just an infrastructure update for Windows users.

kg:
Sorry, lgtm

rossberg-chromium:
Thanks!

PR #196, opened 2015-11-27
Simplify builtins module

rossberg-chromium:
A temporary measure, since the LLVM producer can only generate empty names currently.


sunfishcode:
It seems like it'd be better to put something in wasmate.py to add module names, since it is effectively a linking function. Also, that way this issue stays relatively localized, rather than spread around to all s-expr consumers.

rossberg-chromium:
Yeah, I don't have a specific opinion on this. This change was requested as a stop-gap measure by several people. Note that producers who use the module name will also continue to work just fine, so not necessarily a need for anybody else to adjust.

lukewagner:
+1 on solving this in wasmate.py, since it can simply and unconditionally splat in "stdio", right?

kg:
Yeah, based on my experience with ilwasm I think wasmate should just put in stdio.

rossberg-chromium:
Okay, sounds good. I removed the actual ignoring part from this patch. Since it also contained some clean-up, I'dd like to land the rest anyway. Retitled accordingly. PTAL.

sunfishcode:
The cleanup part lgtm.

sunfishcode:
I've now submitted https://github.com/WebAssembly/experimental/pull/48 to wasmate implementing the wasmate suggestion here.

PR #197, opened 2015-11-27
Do streaming lexing

### LINKED ISSUE #156: String.create error in interpreter
I tried to parse the Poppler PDF library compiled to `.wast` in the interpreter. It's 39MB, and the interpreter errors on

```
[..]/spec/ml-proto/wasm: uncaught exception Invalid_argument("String.create")
Raised by primitive operation at file "host/main.ml", line 12, characters 2-206
Called from file "host/main.ml", line 53, characters 15-24
Called from file "list.ml", line 73, characters 12-15
Called from file "host/main.ml", line 86, characters 4-33
```

I'm not that familiar with ocaml, so I'm not sure what to do to investigate this further.

I can attach the file if that's useful (too big for github though, I think).

rossberg-chromium:
Parse from an input stream instead of a string. Two benefits:

1. Can now parse large files (should fix #156, although I have no 32 bit system around to test).

2. When invoked as a REPL, the interpreter now accepts input stretching multiple lines, and recognises automatically when a command is complete.


hhugo:
using opam, you can install a 32bit compiler on a 64bit host with
`opam switch 4.02.3+32bit`

rossberg-chromium:
Ping. Anybody wants to LGTM this?

kg:
LGTM*

* I don't know ocaml very well but I trust your judgement, it passes tests and it looks OK

PR #202, opened 2015-12-10
Address test

jcbeyler:
Seems like we never had an expected output for this test


jcbeyler:
sorry, let me do this again

PR #204, opened 2015-12-11
Remove has_feature

rossberg-chromium:
AFAIU, this has been dropped from the MVP, so removing it from spec.


kg:
We dropped feature detection from the MVP? When?

lukewagner:
There was some initial discussion around the fact that we don't really have the experience to determine whether `has_feature` is the right primitive to support feature-testing polyfills and that, rather than guessing, we should temporarily remove it, work on a real feature-testing polyfill (using post-MVP features like threads and SIMD as use cases) and then reconsider what is the right primitive.

For example, maybe `has_feature` is ultimately best expressed as a [layer 1](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md) primitive, where it could serve as a condition in the macro layer (analogous to `defined`).  Hard to be sure at this point.

Anyhow, we haven't written up the design repo PR to remove `has_feature` yet; perhaps we should do that first.

rossberg-chromium:
Fair enough. Mainly, this was on my "list of things to clean up before vacation", in prep for teaching ml-proto the binary format from v8-proto. But there is no real hurry.

jcbeyler:
Man, and I just put that in wasm-to-llvm-prototype last week ;-) (it took like... let me think... 5 minutes ? :))

rossberg-chromium:
Merging, following offline resolution today.

PR #205, opened 2016-01-05
align and offset can be specified in either order

binji:


kg:
LGTM

jfbastien:
Great, thanks!

For context, this addresses https://github.com/WebAssembly/sexpr-wasm-prototype/issues/17

sunfishcode:
Wouldn't it be just as easy to change s2wasm to emit these in the spec order? Requiring them to be in a consistent order makes code reduces gratuitous variation, which should make code easier for humans to read in general.

jfbastien:
@sunfishcode yes it's easy to fix, but IMO a human-oriented format shouldn't mandate an order for named parameters so I figured it was easier to fix it officially than let someone else stumble on it later.

sunfishcode:
Fixed order: easier for humans to read
Flexible order: easier for humans to write

I think optimizing for readability is more important than writeability.

kripken:
Not to get philosophical, but is fixed order always easier for humans to read? I don't think it is for me - I would be surprised if the order of these particular things is fixed, i.e., I would not properly reject an invalid input when reading it - and I think I'm human (at least enough to fool captchas).

But I don't feel strongly. Happy to update s2wasm if that's what people want.

sunfishcode:
Perhaps it doesn't for all humans, but I've fooled a few captchas myself ;-), and I find code easier to read when there is less meaningless variation.

sunfishcode:
With https://github.com/WebAssembly/binaryen/issues/66 merged, the immediate concern here is now resolved, so I think we can close this. If people would like to discuss it futher, feel free to re-open or file a new issue.

PR #207, opened 2016-01-12
Update assert_return_nan and tests for the new NaN signbit rules

sunfishcode:
This updates the host environment and tests for the new NaN signbit rules, as discussed in https://github.com/WebAssembly/design/issues/477 .


ml-proto/spec/f32.ml
```diff
@@ -1,6 +1,7 @@
 include Float.Make(struct
                      include Int32
                      let default_nan = 0x7fc00000l
+                     let alternate_nan = 0xffc00000l
```
rossberg-chromium: Nit: Is there a reason to favour one over the other? If not, why not simply call them `pos_nan` and `neg_nan`?

rossberg-chromium:
LGTM

sunfishcode:
Renamed default_nan and alternate_nan to pos_nan and neg_nan per comments; merging with LGTM above.

Issue #208, opened 2016-01-18
Change memory sizes to be in units of the page size.


JSStats:
Could we change the memory sizes, in the memory section for the min and max and in grow-memory to be in units of the pages size which is 64k?

Other sizes for the min and max can not even be encoded in the binary format, and it's just a burden for consumers and producers to have to be rounding these off and scaling them.

This will avoid the need for page size error checking in grow-memory.

kg:
Locking the unit to 64k would prevent us from changing the page size constraints in the future (there are some mentions of this in the design repo). So I think it probably needs to be in bytes.

JSStats:
It was already decided to have just one page size and go with that. There is/was consensus to change the v8 encoding to units of the pages size, and less significant bits are just noise in the wast and a challenge to define how they map to the binary encoding? Happy to close this as it is just a wast suggestion and the deployment encoding is going to page sized units.

JSStats:
Give up on this, keeping them as bytes. There are spec tests imported into other tools such as sexpr-wasm-prototype that would require coordination.

lukewagner:
The point of the wasm page size constant is to not change in the future (after v.1).  We have discussed that in the future modules could optionally declare a bigger (multiple) of page size in the memory section and then the page-count would, I assume, be interpreted as a count of the declared page size.  This would all be backwards-compatible, though, since the bigger page size was opt-in.

So since the binary format is being changed to use units of pages, I'd recommend doing the same for the text format.

JSStats:
Could this be closed now?

lukewagner:
It seems to still be the case that `memory_size` and `grow_memory` return units of bytes.

rossberg-chromium:
Fixed a while ago.

PR #209, opened 2016-01-19
Use page aligned memory sizes in the tests.

JSStats:
Otherwise the tests will not be consistent when run on an implementation that rounds up the memory sizes in bytes to the page size units, and impossible for the runtime to check given that the binary encoding does not store the memory sizes in bytes.

### DIFF
--- a/ml-proto/spec/check.ml
+++ b/ml-proto/spec/check.ml
@@ -299,22 +299,22 @@ let check_start c start =
       "start function must not return anything";
   ) start
 
-let check_segment size prev_end seg =
+let check_segment pages prev_end seg =
   let seg_len = Int64.of_int (String.length seg.it.Memory.data) in
   let seg_end = Int64.add seg.it.Memory.addr seg_len in
   require (seg.it.Memory.addr >= prev_end) seg.at
     "data segment not disjoint and ordered";
-  require (size >= seg_end) seg.at
+  require (Int64.mul pages Memory.page_size >= seg_end) seg.at
     "data segment does not fit memory";
   seg_end
 
 let check_memory memory =
   let mem = memory.it in
   require (mem.initial <= mem.max) memory.at
-    "initial memory size must be less than maximum";
-  require (mem.max <= 4294967296L) memory.at
-    "linear memory size must be less or equal to 4GiB";
-  ignore (List.fold_left (check_segment mem.initial) Int64.zero mem.segments)
+    "initial memory pages must be less than or equal to the maximum";
+  require (mem.max <= 65535L) memory.at
+    "linear memory pages must be less or equal to 65535 (4GiB)";
+  ignore (List.fold_left (check_segment mem.initial) 0L mem.segments)
 
 let check_module m =
   let {memory; types; funcs; start; imports; exports; table} = m.it in
--- a/ml-proto/spec/eval.ml
+++ b/ml-proto/spec/eval.ml
@@ -289,7 +289,7 @@ and eval_hostop c hostop vs at =
      * Since we currently only support i32, just test that. *)
     if I64.gt_u new_size (Int64.of_int32 Int32.max_int) then
       Trap.error at "memory size exceeds implementation limit";
-    Memory.grow mem delta;
+    Memory.grow mem (Int64.div delta Memory.page_size);
     Some (Int32 (Int64.to_int32 old_size))
 
   | _, _ ->
--- a/ml-proto/spec/memory.ml
+++ b/ml-proto/spec/memory.ml
@@ -46,7 +46,7 @@ let host_index_of_int64 a n =
 
 
 let create' n =
-  let sz = host_size_of_int64 n in
+  let sz = host_size_of_int64 (Int64.mul n page_size) in
   let mem = Array1.create Int8_unsigned C_layout sz in
   Array1.fill mem 0;
   mem
@@ -68,9 +68,9 @@ let init mem segs =
 let size mem =
   int64_of_host_size (Array1.dim !mem)
 
-let grow mem n =
-  let old_size = size mem in
-  let new_size = Int64.add old_size n in
+let grow mem pages =
+  let old_size = Int64.div (size mem) page_size in
+  let new_size = Int64.add old_size pages in
   if I64.gt_u old_size new_size then raise SizeOverflow else
   let after = create' new_size in
   let host_old_size = host_size_of_int64 old_size in
--- a/ml-proto/TestingTodo.md
+++ b/ml-proto/TestingTodo.md
@@ -12,7 +12,6 @@ Misc semantics:
  - test that too-big linear memory initial allocation fails
  - test that function addresses are monotonic indices, and not actual addresses.
  - ~~test that non-pagesize `grow_memory` fails~~
- - test that non-pagesize initial linear memory allocation fails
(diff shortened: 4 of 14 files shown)


sunfishcode:
If we're going to make it a requirement that memory sizes be specified in multiples of page sizes, we should make the spec repo enforce it.

JSStats:
@sunfishcode I did not get consensus to change the wast to specify the memory min and max in units of the page size, which was my first attempt. But there appeared to be consensus to change the v8 encoding to specify them in units of the page size. I really don't care how the wast specifies them, and I would be happy to signal an error if they are not page aligned? Changing these tests is an incremental change though, a first step in landing some of the other changes? Are you objecting to this as a first step? Do you need a change to the spec reference implementation before landing this?

sunfishcode:
A change to all the tests to use page-size multiples the spec should be accompanied by a change to enforce page-size multiples in the spec. I believe consensus for such a change is plausible, since this is one of the advantages of having a fixed known page size, though it'll require a clear and concise proposal.

JSStats:
I thought there was already consensus here. The spec test changes are blocking this as sexp-wasm depends on them.

https://codereview.chromium.org/1597163002/
Issue 1597163002: wasm: change the module memory size to be multiples of the page size, 64k.
'lgtm' status

https://github.com/WebAssembly/sexpr-wasm-prototype/pull/21
Change the memory sizes to be in units of pages rather than powers of two

https://github.com/WebAssembly/binaryen/pull/121
Memory sizes are in pages.

sunfishcode:
If there's a consensus already, then changing the spec should be uncontroversial.

JSStats:
Updated the spec to assert that the memory sizes are page aligned, fixed a few tests this caught, and added a few more tests.

rossberg-chromium:
@sunfishcode, @lukewagner, deferring to you

lukewagner:
I think the recent PR to specify memory sizes (`initial` and `max`) in terms of the wasm page size (64KiB) obviates these changes.

JSStats:
@lukewagner The change to the binary encoded in the design repo has landed, and does specify these in pages, but updating the tests is blocking landing any implementation or tool changes because tests will fail.

The original patch also used units of pages in the wast and in the tests, but there were strong objections so units of bytes was retained in the wast.

This create an issue of rounding them off to pages. There was a request to simply fail when parsing or checking wast sizes that are not page aligned.

Perhaps I have missed some other patch that landed by this still seems to be blocking progress on this issue.

lukewagner:
I see, thanks for explaining.  So lgtm on rounding byte sizes up to page sizes as a quick way to make progress.  Were the objections to changing text format from bytes to pages made before the binary encoding change?  Either way, it makes sense to me to have the text format reflect the binary format in this regard, but that could be a separate PR.

JSStats:
@lukewagner Thank you. See issue https://github.com/WebAssembly/spec/issues/208 for the objection to using units of pages in the wast. If people have changed their position then lets fix it now before landing anything to save work later as there are patch to binaryen and sexp-wasm and v8 pending.

@sunfishcode requested the page aligned sized enforced, see the second comment in this PR 'If we're going to make it a requirement that memory sizes be specified in multiples of page sizes, we should make the spec repo enforce it.' I did find that this help find issues so I support this rather than having consumers rounding up the sizes.

JSStats:
There have been no current objections in https://github.com/WebAssembly/spec/issues/208 to using units of pages in the wast memory section for the initial and max memory so as requested the patch has been updated. ptal

lukewagner:
lgtm!  Given page units in text and binary, seems like we should store in units of pages in the AST too so we can drop the never-hit page-size checks in check.ml.   @rossberg-chromium agreed?

rossberg-chromium:
Yup, agreed with @lukewagner, AST should change.

JSStats:
Think that was a 'yes' to removing the page alignment checks, done. ptal

lukewagner:
What I think we were agreeing to is that parser.mly would store the number of pages directly in the `initial`/`max` fields of the ast and the multplication by `page_size` would instead happen when creating the instance.  Perhaps `Memory.create` would also take units of pages too (since `Memory` defines `page_size`) and throw `Out_of_memory` if the multiplication goes out of range (there's already an `Out_of_range` condition for the `Int64` being bigger than `max_int`).

JSStats:
Sorry, misunderstood the request. Updated. ptal

lukewagner:
Excellent, lgtm!  You too @rossberg-chromium?

ml-proto/spec/check.ml
```diff
@@ -1,6 +1,7 @@
 open Kernel
 open Source
 open Types
+open Memory
```
rossberg-chromium: Only open modules if there are plenty of uses.

ml-proto/spec/check.ml
```diff
@@ -300,21 +301,21 @@ let check_start c start =
...
 let check_memory memory =
   let mem = memory.it in
   require (mem.initial <= mem.max) memory.at
-    "initial memory size must be less than maximum";
-  require (mem.max <= 4294967296L) memory.at
-    "linear memory size must be less or equal to 4GiB";
+    "initial memory pages must be less than or equal to the maximum";
+  require (mem.max <= 65535L) memory.at
+    "linear memory pages must be less or equal to 65535 (4GiB)";
   ignore (List.fold_left (check_segment mem.initial) Int64.zero mem.segments)
```
rossberg-chromium: While you're here, I wouldn't mind you replacing `Int64.zero` with `0L`.

ml-proto/spec/memory.ml
```diff
@@ -72,7 +72,7 @@ let grow mem n =
   let old_size = size mem in
   let new_size = Int64.add old_size n in
```
rossberg-chromium: For consistency with `create`, please change `grow` to also take a number of pages now.

JSStats:
Thank you. Updated. ptal

ml-proto/spec/check.ml
```diff
@@ -299,22 +299,22 @@ let check_start c start =
...
   ) start
 
-let check_segment size prev_end seg =
+let check_segment pages prev_end seg =
   let seg_len = Int64.of_int (String.length seg.it.Memory.data) in
   let seg_end = Int64.add seg.it.Memory.addr seg_len in
   require (seg.it.Memory.addr >= prev_end) seg.at
     "data segment not disjoint and ordered";
-  require (size >= seg_end) seg.at
+  require ((Int64.mul pages Memory.page_size) >= seg_end) seg.at
```
rossberg-chromium: Nit: redundant parens

ml-proto/spec/eval.ml
```diff
@@ -289,7 +289,7 @@ and eval_hostop c hostop vs at =
      * Since we currently only support i32, just test that. *)
     if I64.gt_u new_size (Int64.of_int32 Int32.max_int) then
       Trap.error at "memory size exceeds implementation limit";
-    Memory.grow mem delta;
+    Memory.grow mem (Int64.div delta Memory.page_size);
```
rossberg-chromium: Hm, I wonder if we should adjust the grow_memory operator as well. But probably for a separate patch.

rossberg-chromium:
LGTM modulo nit

JSStats:
Fixed nit, thank you. ptal.

PR #214, opened 2016-01-22
Fix tableswitch syntax.

mbebenita:
Not quite sure what the `<switch>` non-terminal is for, but it looks wrong.


ml-proto/README.md
```diff
@@ -98,7 +98,7 @@ expr:
   ( loop <name1>? <name2>? <expr>* )             ;; = (block <name1>? (loop <name2>? (block <expr>*)))
   ( br <var> <expr>? )
   ( return <expr>? )                             ;; = (br <current_depth> <expr>?)
-  ( tableswitch <name>? <expr> <switch> ( table <target>* ) <case>* )
+  ( tableswitch <name>? <expr> ( table <target>* ) <case>* )
```
binji: you're missing the default `<target>` after the table too

rossberg-chromium:
LGTM, thanks!

PR #219, opened 2016-01-28
Add a test for a memory operator with an 'align' argument and without an offset argument.

JSStats:
This case seems to be permitted but there was no test for this.

I would also have liked to add a 'bad' test for the case in which the `align` argument precedes the `offset` argument. It generates a `syntax error` and I can't see any other examples of testing for bad syntax?

When working on an operator encoding table, the order of these was noted as an issue. The table might want to allow multiple memory arguments in an operator, and if the order is not fixed then the reverse order creates an ambiguity as to which memory access these options apply to. So I would like to fix the order and test for it so that other tools can check this too.

Tests for bad multiple `align` and `offset` arguments might also be useful.


rossberg-chromium:
Deferring to @lukewagner for this one.

JSStats:
I note that there seems to be consensus to have a fixed ordering of align and offset now. What I missed was that it is still ambiguous placing two memory option arguments one after the other because of the case in which both the align and offset are missing. I needed some help on how to write a test for syntax errors in ml-proto to test for multiple align or offset arguments or out-of-order align and offset?

lukewagner:
The patch lgtm.  I'm not sure what the best way to test a syntax error is since iirc the entire .wast is parsed up front.  Perhaps we could add some `assert_syntax_error` that embeds a whole `(module ...)` as a string literal and then the parser is invoked when that command is executed?

JSStats:
@lukewagner Could you land this one, just a positive test. I don't expect to be adding new testing infrastructure to support testing syntax errors.

PR #221, opened 2016-01-31
Reorder select's operands to put the condition last.

sunfishcode:
This implements the select portion of
https://github.com/WebAssembly/design/pull/489

br_if operands are more involved, so I'll submit a separate PR for them.


ml-proto/spec/eval.ml
```diff
@@ -234,9 +234,9 @@ let rec eval_expr (c : config) (e : expr) =
       with exn -> arithmetic_error e.at e1.at e2.at exn)
 
   | Select (selop, e1, e2, e3) ->
-    let cond = int32 (eval_expr c e1) e1.at in
-    let v1 = some (eval_expr c e2) e2.at in
-    let v2 = some (eval_expr c e3) e3.at in
+    let cond = int32 (eval_expr c e3) e3.at in
```
rossberg-chromium: It seems you also need to move this line after the other calls.

sunfishcode:
Fixed the evaluation order now that 76553cfeb6c29c813c949487322401ebf1af82f6 has landed which has the test for operand evaluation order.

rossberg-chromium:
lgtm

PR #226, opened 2016-02-01
Allow trailing dot before exponent in float literal

binji:
See https://github.com/WebAssembly/sexpr-wasm-prototype/issues/29

### DIFF
--- a/ml-proto/host/lexer.mll
+++ b/ml-proto/host/lexer.mll
@@ -93,8 +93,8 @@ let num = sign digit+
 let hexnum = sign "0x" hexdigit+
 let int = num | hexnum
 let float =
-    (num '.' digit+)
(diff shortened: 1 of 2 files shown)


rossberg-chromium:
LGTM, although I wonder why anybody would want this.

binji:
Not sure, but @jfbastien was surprised it didn't work in sexpr-wasm. I guess it's commonly supported in C-like languages.

jfbastien:
In C it's common to write `1.` to get a `double` and `1.f` to get a `float`. Without the `.` it would be an integral value. That naturally extends to floating-point values with exponents e.g. `1.e300`.

With wasm it's not clearly as useful because you can't mix types and get conversions as C gives you. I therefore agree it's not useful for wasm per-se, but it's useful for people who are used to their trailing dot :-)

PR #232, opened 2016-02-03
Remove parsing support for `(; ... ;)` comments.

sunfishcode:
Accompanying https://github.com/WebAssembly/spec/pull/228 removing the tests for `(; ... ;)` comments, this is a PR removing lexing support for them from the interpreter.

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -149,7 +149,6 @@ Comments can be written in one of two ways:
 ```
 comment:
   ;; <character>* <eol>
-  (; (<character> | <comment>)* ;)
 ```
 
 In particular, comments of the latter form nest properly.
--- a/ml-proto/host/lexer.mll
+++ b/ml-proto/host/lexer.mll
@@ -357,15 +357,12 @@ rule token = parse
 
   | ";;"[^'\n']*eof { EOF }
   | ";;"[^'\n']*'\n' { Lexing.new_line lexbuf; token lexbuf }
-  | "(;" { comment (Lexing.lexeme_start_p lexbuf) lexbuf; token lexbuf }
   | space { token lexbuf }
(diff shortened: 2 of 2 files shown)


binji:
I actually found these to be pretty convenient, and use them for many tests in the sexpr-wasm repo as part of the test format. Seems like we should have some alternative for a block comment if these are removed.

sunfishcode:
I'm ok either way; I just want the spec and testsuite to remain in sync. If we keep this syntax, we should revert #228 or add new tests.

binji:
@JSStats @rossberg-chromium Is there a preferable sexpr block comment syntax?

haxxpop:
How about `;- comment -;` ? This is quite analogous to the C style (`//` and `/* comment */`), but, in sexpr, we use `;;` and `;- comment -;`

JSStats:
Lisp uses a `;` to start a comment, and it runs to the end of the line. Then it becomes a matter of personal preference what conventions you use after this. For example: `;;;;` for large section comments, `;;;` for function header comments, `;;` for comments between statements in a line of their own, and `;` to start line end comments.

JSStats:
@binji I see how this is used in the sexp-wasm d8 tests. Common Lisp has `#| ...|#` for this but there seems little reason to use that here. If prefixing every comment line with `;` is a burden then perhaps keep the block support. But could the use in the spec tests at least be confined to one file that tests this alone, just to help general s-exp parsers a little which could just skip the file. Not a big issue either way, it would just need some extra parsing support for a general s-exp parse.

rossberg-chromium:
I also find block comments pretty useful soemtimes, and wouldn't want to lose them. Agreed that there should be a test for them.

rossberg-chromium:
Closing. We now have explicit tests.

PR #233, opened 2016-02-04
Add tests for block comments, in a separate file.

JSStats:
https://github.com/WebAssembly/spec/pull/232

### DIFF
--- a/ml-proto/test/block_comments.wast
+++ b/ml-proto/test/block_comments.wast
@@ -1,0 +1,13 @@
+(;;)
+(;(((((((((( ;)
+(;)))))))))));)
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
LGTM, but you might also want to test proper nesting, e.g.
```
(; a (; b ;) c ;)
(; ;; bla ;)
(; ;; bla
;)
```
Also, why not just name the test `comments`?

JSStats:
Thank you, added the nesting tests. A simple s-exp parser might not bother with the block comments, and just skip this test, but might still implement the `;;` comments which are more common and could be used in the other tests.

Issue #227, opened 2016-02-02
Questionable 'arity mismatch' - re-submitted.


JSStats:
Re-submitting https://github.com/WebAssembly/spec/issues/223 as I believe it was incorrectly closed.

This example fails with a 'arity mismatch' in the ml-proto. It is accepted by sexp-wasm.

```
(module
 (func $f1 (result i32)
   (block $l0
     (br_if (i32.const 1) $l0
       (block $l1
         (br $l1 (i32.const 1))))
     (i32.const 1))))
```

@rossberg-chromium wrote 'This is an inverse-duplicate of #179. Please reopen that if you disagree with the resolution.'

The above case appears to be significantly different to that in #179. The expected type for `block $l1` is the expected type of `block $l0` which is `i32`. The expected type and the actual type do appear to match here?  Sorry if I have missed something?

rossberg-chromium:
It actually is the very issue. The `br_if` desugars into an expression that contains
```
(block (block $l1 (br $l1 (i32.const 1))) (nop))
```
which is rejected because the inner block has type `()`.

Now, one could argue that this is primarily a problem with the specific desugaring (which will have to change anyway to reflect the new argument order). But I need to think about this a bit. The ()-subtyping is designed to give implementations (and type checkers) some leeway where they drop the values, because the difference is not observable. But that doesn't extend to label types under the current rules. With the original multi-values semantics, the above worked -- that would be an argument in favour of your request to relax break typing to what it was before. Alternatively, a more targeted change is to allow the label's type to be a subtype of its block's type. However, that would make top-down typing non-deteministic.

JSStats:
@rossberg-chromium I'm sure you know this code well, and I can barely read the Ocaml, but I would note that `br_if` seems to desugar to `if` (and then `if_else`) but when doing this manually the error no longer occurs, even adding the extra block with a fall-through nop as in your example above.

```
(module
 (func $f1 (result i32)
   (block $l0
     (if (i32.const 1)
       (br $l0
         (block $l1
           (br $l1 (i32.const 1)))))
     (i32.const 1))))
```

rossberg-chromium:
No, the actual desugaring is
```
(module
  (func $f1 (result i32)
    (block $l0
      (if_else (i32.const 1)
        (br $l0 (block $l1 (br $l1 (i32.const 1))))
        (block (block $l1 (br $l1 (i32.const 1))) (nop))
      )
      (i32.const 1)
    )
  )
)
```
[...]

PR #234, opened 2016-02-04
Put br_if conditions last.

sunfishcode:
This implements the second half of https://github.com/WebAssembly/design/pull/489.

Desugaring br_if to if+break was fairly awkward; adding br_if to the kernel
language was straightforward and avoided the awkwardness.

This change happened to expose an aspect of `br_if`'s type checking which
differed from my expectations, so I've taken the liberty of fixing that
too. See the deleted assert_invalid in test/labels.wast and the added
$br_if1 testcase. `br_if` either sends its value to the destination label
or to its own result.

### DIFF
--- a/ml-proto/host/parser.mly
+++ b/ml-proto/host/parser.mly
@@ -233,7 +233,8 @@ expr1 :
   | LOOP labeling1 labeling1 expr_list
     { fun c -> let c' = $2 c in let c'' = $3 c' in Loop ($4 c'') }
   | BR var expr_opt { fun c -> Br ($2 c label, $3 c) }
-  | BR_IF expr var expr_opt { fun c -> Br_if ($2 c, $3 c label, $4 c) }
+  | BR_IF expr var { fun c -> Br_if (None, $2 c, $3 c label) }
+  | BR_IF expr expr var { fun c -> Br_if (Some ($2 c), $3 c, $4 c label) }
   | RETURN expr_opt
     { let at1 = ati 1 in
       fun c -> Return (label c ("return" @@ at1) @@ at1, $2 c) }
--- a/ml-proto/host/parser.mly
+++ b/ml-proto/host/parser.mly
@@ -233,8 +233,8 @@ expr1 :
   | LOOP labeling1 labeling1 expr_list
     { fun c -> let c' = $2 c in let c'' = $3 c' in Loop ($4 c'') }
   | BR var expr_opt { fun c -> Br ($2 c label, $3 c) }
-  | BR_IF expr var { fun c -> Br_if (None, $2 c, $3 c label) }
-  | BR_IF expr expr var { fun c -> Br_if (Some ($2 c), $3 c, $4 c label) }
+  | BR_IF var expr { fun c -> Br_if ($2 c label, None, $3 c) }
+  | BR_IF var expr expr { fun c -> Br_if ($2 c label, Some ($3 c), $4 c) }
   | RETURN expr_opt
     { let at1 = ati 1 in
       fun c -> Return (label c ("return" @@ at1) @@ at1, $2 c) }
--- a/ml-proto/spec/ast.ml
+++ b/ml-proto/spec/ast.ml
@@ -19,7 +19,7 @@ and expr' =
   | Block of expr list
   | Loop of expr list
   | Br of var * expr option
-  | Br_if of expr option * expr * var
+  | Br_if of var * expr option * expr
   | Return of var * expr option
   | If of expr * expr
   | If_else of expr * expr * expr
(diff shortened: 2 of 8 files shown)


JSStats:
The prior type check appears to be because `br_if` returned the empty-values type, the same type that `nop` returns, and it was in the fall-through position of the `block $l` which expected a single `f32` value.

Your changes to the type checking appear to be consistent with `br_if` now returning the result of evaluating the expression and is that the case?  So when the result expression is omitted it would be equivalent to a `nop` and return the same as a `nop. When the expression is a `nop` this would be equivalent?

With the change, is there still the duplication noted in https://github.com/WebAssembly/spec/issues/227 ?  Perhaps it resolves that issue too.

This might have some interaction with https://github.com/WebAssembly/spec/pull/215

Any opinion on https://github.com/WebAssembly/design/issues/529 while your working in this area?

sunfishcode:
I don't wish to connect this issue to #215 or #227. Accordingly, I've now removed that part of the change. This PR now just contains the operand order change.

I would also like to keep this issue focused and avoid discussions of WebAssembly/design#529.

JSStats:
@sunfishcode Are you saying you don't understand the type system issues and can not answer these questions? Does it change the existing behaviour as noted in https://github.com/WebAssembly/spec/issues/227 ? Does it change the 'arity' checking as noted in https://github.com/WebAssembly/spec/pull/215 ? Or are you asking for help, do you want other people to check this and give feedback?

This does move the arguments around, and it would look better to me if the label were first, and the v8 encoding has the label depth first, it is an immediate, and having it last is also not consistent with `br` - would not block the issue on this but still your are moving them and I think we can expect you to champion this change and engage in discussion?

sunfishcode:
I'm saying I don't want the part of this PR I'm primarily interested in get lost in tangential discussions.

JSStats:
This does change the behaviour of the issue noted in https://github.com/WebAssembly/spec/issues/227 For example this now passes:

```
(module
 (func $f1 (result i32)
   (block $l0
     (br_if (block $l1
              (br $l1 (i32.const 1)))
            (i32.const 1) $l0)
     (i32.const 1))))
```

This seems a good thing to me, but there have been different opinions!

JSStats:
Fwiw is seems to be neutral wrt https://github.com/WebAssembly/spec/pull/215 The following still gives an 'arity mismatch' error:

```
(module
 (func $f1
   (block $l0
     (br_if (nop) (i32.const 1) $l0))))
```

sunfishcode:
So it does; I've now added the testcase.

ml-proto/host/parser.mly
```diff
@@ -233,7 +233,8 @@ expr1 :
   | LOOP labeling1 labeling1 expr_list
     { fun c -> let c' = $2 c in let c'' = $3 c' in Loop ($4 c'') }
   | BR var expr_opt { fun c -> Br ($2 c label, $3 c) }
-  | BR_IF expr var expr_opt { fun c -> Br_if ($2 c, $3 c label, $4 c) }
+  | BR_IF expr var { fun c -> Br_if (None, $2 c, $3 c label) }
+  | BR_IF expr expr var { fun c -> Br_if (Some ($2 c), $3 c, $4 c label) }
```
rossberg-chromium: That syntax is really odd. The argument should still go after the label, to be consistent with plain `br` (and calls). So I think this should be `BR_IF var expr expr` instead.

ml-proto/spec/ast.ml
```diff
@@ -19,7 +19,7 @@ and expr' =
   | Block of expr list
   | Loop of expr list
   | Br of var * expr option
-  | Br_if of expr * var * expr option
+  | Br_if of expr option * expr * var
```
rossberg-chromium: Similarly here: `Br_if of var * expr option * expr`.

ml-proto/spec/kernel.ml
```diff
@@ -82,6 +82,7 @@ and expr' =
   | Block of expr list                      (* execute in sequence *)
   | Loop of expr                            (* loop header *)
   | Break of var * expr option              (* break to n-th surrounding label *)
+  | Br_if of expr option * expr * var       (* conditional break *)
```
rossberg-chromium: And here.

rossberg-chromium:
Looks good modulo the argument order. It's really unfortunate that the kernel now loses orthogonality and has to duplicate the break logic, but I have no better answer either (other than removing `br_if` :) ).

sunfishcode:
@rossberg-chromium Thanks! I added a patch to put the argument after the label.

rossberg-chromium:
lgtm

sunfishcode:
Thanks!

PR #235, opened 2016-02-05
Use 0 instead of None for alignment

lukewagner:
I realized that since `0` is already an invalid alignment (currently fails the "is a power of 2" check) we don't need alignment to be an `int option`, we can simply use an `int`; no reason to have two sentinel cases.  This also makes it an epsilon simpler when talking about binary encoding.

### DIFF
--- a/ml-proto/host/lexer.mll
+++ b/ml-proto/host/lexer.mll
@@ -157,34 +157,44 @@ rule token = parse
 
   | (nxx as t)".load"
     { LOAD (fun (o, a, e) ->
-        numop t (I32_load (o, a, e)) (I64_load (o, a, e))
-                (F32_load (o, a, e)) (F64_load (o, a, e))) }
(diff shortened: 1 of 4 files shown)


sunfishcode:
Another approach would be to make alignment non-optional, and then optimize the binary encoding using the idea of including immediate values in the opcode table. That way we wouldn't need any sentinel values. What do you think?

lukewagner:
Oh right, that's even simpler and all of the encoding schemes we're talking about shouldn't benefit from having `0` vs. `4`.

lukewagner:
Updated. So 0 is back to invalid and the parser fills in natural alignment if you leave it off.  For testing purposes, explicitly setting `alignment=0` will continue to emit 0 and cause a validation error.  On a side note, this was much easier than last time I mucked loads/stores; nice job @rossberg-chromium!

sunfishcode:
lgtm

rossberg-chromium:
LGTM, I like this better as well

PR #236, opened 2016-02-05
 Support start, add testcase, and can be executed properly

haxxpop:
The following is the list of what I have done to add start functionality

- I have changed the Ast module and Kernel module to include the start function index.
- Have the start function run as soon as the module is loaded.
- In the print_module function in host/print.ml, I have added the start function to be the output to the screen
- Add test/start.wast to test the start function

Please consider if we should add this feature to the main repo

### DIFF
--- a/ml-proto/host/print.ml
+++ b/ml-proto/host/print.ml
@@ -30,7 +30,7 @@ let print_table_elem i x =
 let print_start start =
   match start with
   | Some x -> printf "start = func %d\n" x.it
-  | None -> printf "there is no start\n"
+  | None -> ()
 
 (* Ast *)
 
--- a/ml-proto/host/print.ml
+++ b/ml-proto/host/print.ml
@@ -28,9 +28,7 @@ let print_table_elem i x =
   printf "table [%d] = func %d\n" i x.it
 
 let print_start start =
-  match start with
-  | Some x -> printf "start = func %d\n" x.it
-  | None -> ()
+  Lib.Option.app (fun x -> printf "start = func %d\n" x.it) start
 
 (* Ast *)
 
--- a/ml-proto/host/print.ml
+++ b/ml-proto/host/print.ml
@@ -27,6 +27,10 @@ let print_export_sig m prefix n f =
 let print_table_elem i x =
   printf "table [%d] = func %d\n" i x.it
 
+let print_start start =
+  match start with
+  | Some x -> printf "start = func %d\n" x.it
+  | None -> printf "there is no start\n"
 
 (* Ast *)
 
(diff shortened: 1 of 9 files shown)


ml-proto/host/print.ml
```diff
@@ -27,6 +27,10 @@ let print_export_sig m prefix n f =
 let print_table_elem i x =
   printf "table [%d] = func %d\n" i x.it
 
+let print_start start =
+  match start with
+  | Some x -> printf "start = func %d\n" x.it
+  | None -> printf "there is no start\n"
```
rossberg-chromium: Nit: just don't print anything in this case.

ml-proto/spec/eval.ml
```diff
@@ -313,12 +313,18 @@ let add_export funcs ex =
...
-   host}
+  let {memory; funcs; exports; start; _} = m.it in
+  let instance =
+    {module_ = m;
+     imports;
+     exports = List.fold_right (add_export funcs) exports ExportMap.empty;
+     memory = Lib.Option.map init_memory memory;
+     host}
+  in
+  ignore (match start with
```
rossberg-chromium: Nit: Apply `ignore` more locally to the `eval_func` call.

haxxpop: What does "more locally" mean ?
Does it mean something like this ?

  (match start with 
    | Some x -> ignore (eval_func instance (lookup "function" funcs x) []) 
    | None -> ());

rossberg-chromium: Yes. In fact, you can also simplify the whole thing to
```
Lib.Option.app (fun x -> ignore (eval_func instance (lookup "function" funcs x) []) start;
```

haxxpop:
@rossberg-chromium 
I have simplified both print_start and eval_func. You can check it in the latest commit.

rossberg-chromium:
Thanks, LGTM.

jfbastien:
Will this assert out if `$foo` isn't nullary in `(start $foo)`?

haxxpop:
@jfbastien Oh, thank you. I have recently modified it to assert that it's nullary before calling eval_func.

jfbastien:
Are there tests in the spec to validate this? Or do we just not do negative tests at the moment?

haxxpop:
No, not yet. I will do it soon :)

haxxpop:
@jfbastien I added it already. There are three negative tests in test/start.wast now.

haxxpop:
@rossberg-chromium fixed :)

rossberg-chromium:
LGTM, thanks!

lukewagner:
Nice!

PR #237, opened 2016-02-09
Make br_if return the value of its value operand.

sunfishcode:
This corresponds to https://github.com/WebAssembly/design/pull/539.

### DIFF
--- a/ml-proto/spec/check.ml
+++ b/ml-proto/spec/check.ml
@@ -231,7 +231,7 @@ and check_exprs c ts es =
 
 and check_expr_opt c et eo at =
   match et, eo with
-  | _, Some e -> check_expr c et e
+  | Some t, Some e -> check_expr c et e
   | None, None -> ()
   | _ -> error at "arity mismatch"
 
--- a/ml-proto/spec/check.ml
+++ b/ml-proto/spec/check.ml
@@ -134,9 +134,9 @@ let rec check_expr c et e =
     check_expr_opt c (label c x) eo e.at
 
   | Br_if (x, eo, e) ->
-    check_expr_opt c (label c x) eo e.at;
+    check_expr_opt c et eo e.at;
     check_expr c (Some Int32Type) e;
-    check_type None et e.at
+    check_type (label c x) et e.at
 
   | If (e1, e2, e3) ->
     check_expr c (Some Int32Type) e1;
@@ -231,7 +231,7 @@ and check_exprs c ts es =
 
 and check_expr_opt c et eo at =
   match et, eo with
-  | Some t, Some e -> check_expr c et e
+  | _, Some e -> check_expr c et e
   | None, None -> ()
   | _ -> error at "arity mismatch"
 
--- a/ml-proto/spec/check.ml
+++ b/ml-proto/spec/check.ml
@@ -135,8 +135,8 @@ let rec check_expr c et e =
 
   | Br_if (x, eo, e) ->
     check_expr_opt c et eo e.at;
-    check_expr c (Some Int32Type) e;
-    check_type (label c x) et e.at
+    check_expr_opt c (label c x) eo e.at;
+    check_expr c (Some Int32Type) e
 
   | If (e1, e2, e3) ->
     check_expr c (Some Int32Type) e1;
--- a/ml-proto/spec/eval.ml
+++ b/ml-proto/spec/eval.ml
@@ -157,7 +157,7 @@ let rec eval_expr (c : config) (e : expr) =
   | Br_if (x, eo, e) ->
     let v = eval_expr_opt c eo in
     let i = int32 (eval_expr c e) e.at in
-    if i <> 0l then raise (label c x v) else None
+    if i <> 0l then raise (label c x v) else v
 
   | If (e1, e2, e3) ->
     let i = int32 (eval_expr c e1) e1.at in
--- a/ml-proto/test/labels.wast
+++ b/ml-proto/test/labels.wast
@@ -141,10 +141,12 @@
           (br_if $l (f32.const 3) (get_local 0))))))
 
   (func $br_if6
-    (loop $l0 $l1 (br $l0 (i32.const 0))))
+    (loop $l0 $l1
+      (br $l0 (i32.const 0))))
 
   (func $br_if7
-    (block $l0 (br_if $l0 (nop) (i32.const 1))))
+    (block $l0
+      (br_if $l0 (nop) (i32.const 1))))
 
   (func $br_if8 (result i32)
     (block $l0
@@ -153,6 +155,10 @@
         (block (block $l1 (br $l1 (i32.const 1))) (nop))
       )
     (i32.const 1)))
+
+  (func $br_if9
+    (block $l
+      (f32.neg (br_if $l (f32.const 0) (i32.const 1)))))
 
   (func $misc1 (result i32)
    (block $l1 (i32.xor (br $l1 (i32.const 1)) (i32.const 2)))
@@ -179,6 +185,7 @@
   (export "br_if6" $br_if6)
   (export "br_if7" $br_if7)
   (export "br_if8" $br_if8)
+  (export "br_if9" $br_if9)
   (export "misc1" $misc1)
   (export "misc2" $misc2)
 )
@@ -208,13 +215,13 @@
 (assert_return (invoke "br_if6"))
 (assert_return (invoke "br_if7"))
 (assert_return (invoke "br_if8") (i32.const 1))
+(assert_return (invoke "br_if9"))
 (assert_return (invoke "misc1") (i32.const 1))
 (assert_return (invoke "misc2") (i32.const 1))
 
 (assert_invalid (module (func (block $l (f32.neg (br_if $l (i32.const 1))) (nop)))) "arity mismatch")
 
 (assert_invalid (module (func (result i32) (block $l (br_if $l (f32.const 0) (i32.const 1))))) "type mismatch")
-(assert_invalid (module (func (block $l (f32.neg (br_if $l (f32.const 0) (i32.const 1)))))) "type mismatch")
 (assert_invalid (module (func (param i32) (result i32) (block $l (f32.neg (br_if $l (f32.const 0) (get_local 0)))))) "type mismatch")
 
 ;; br_if's own result type doesn't match the result type of the block it's in.
@@ -251,4 +258,12 @@
       (f32.neg
         (block $i
           (br_if $l (f32.const 3) (get_local 0)))))))
+  "type mismatch")
+
+;; br_if result not used, but value's type doesn't match branched-to block.
+(assert_invalid (module
+  (func (param i32) (result i32)
+    (block $l
+      (br_if $l (f32.const 3) (get_local 0))
+      (i32.const 0))))
   "type mismatch")
--- a/ml-proto/test/labels.wast
+++ b/ml-proto/test/labels.wast
@@ -90,28 +90,6 @@
     )
   )
 
-  (func $br_if0 (result i32)
-    (local $i i32)
-    (set_local $i (i32.const 0))
-    (block $outer
-      (block $inner
-        (br_if $inner (i32.const 0))
-        (set_local $i (i32.or (get_local $i) (i32.const 0x1)))
-        (br_if $inner (i32.const 1))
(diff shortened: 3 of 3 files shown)


ml-proto/spec/check.ml
```diff
@@ -134,9 +134,9 @@ let rec check_expr c et e =
     check_expr_opt c (label c x) eo e.at
 
   | Br_if (x, eo, e) ->
+    check_expr_opt c et eo e.at;
```
rossberg-chromium: This doesn't look right, it breaks the implicit result dropping subsumption that is allowed for all other expressions. For example, you could no longer use a `br_if` with argument in a block, as your test cases already demonstrate:
```
(block (br_if $l (arg) (cond)) (other))
```
becomes a type error. That violates a basic design principle of the type system, and doesn't make sense from a practical perspective.

However, the previous version of this PR changed the typing of `br` and `return`. As @JSStats pointed out, it effectively reverted PR #180, which addressed issue #179. We could work around that with more special casing, but that would be fairly ugly, and I'm not convinced it's worth it.

I think the least inconsistent way to implement this change is to reach agreement that we undo #180. @jcbeyler, @lukewagner, would you be okay with that? It's not ideal either, but at this point, I don't have a particularly strong opinion about that one. It's still much better than introducing special case rules for optional arguments, which I'd very much like to avoid.

sunfishcode: From a purely practical perspective, there's no problem with making

```
    (block (br_if $l (arg) (cond)) (other))
```

a type error, because can achieve a similar effect writing this code as

```
   (block (arg) (br_if $l (cond)) (other))
```

The practical restriction is that one can't use `br_if` to pass a value to one successor and drop it in the other. However, this restriction is consistent with how I expect to use `br_if`, and it's consistent with `br_if`'s operand being strict. It's evaluated for both values of the condition, so it makes sense to type-check it for both values of the condition.

lukewagner: @rossberg-chromium Rather than revert, could we do `br0`/br0_if`?  Then we keep the syntactic arity match (which is what postorder needs) and then this PR would switch to `check_type` of an `expr` and the result-dropping works.

rossberg-chromium: @sunfishcode, that's not correct, the two expressions are completely different. The latter won't transfer `arg` to `$l`, and thus won't type-check if `$l` actually requires a value. There is no trivial way to rewrite the former under the current PR. You'll essentially need the encoding of a "drop" operator, avoiding which was a core design goal of the type system. (The simplest such encoding I can see would be `(block (if (i32.const 1) (br_if $l (arg) (cond)) (other))`.)

@lukewagner, special-casing typing on label arity would paint us into a corner regarding multiple values. I think we need a more future-proof solution. Reverting #180 is the one I can see.

I'm starting to wish we had a let-like construct, then we could easily desugar `(br_if $l (arg) (cond))` as follows and get the desired typing for free:
```
(let $x (arg) (block (if (cond) (br $l (get_value $x))) (get_value $x)))
```
:)

lukewagner: @rossberg-chromium If we decoupled construction of tuples of values from `br` (such that you'd write `(br_if $1 (tuple a b) (cond))` or just `(br_if $1 (call foo_returning_tuple) (cond))`, then is there a problem?  In the model of this I have in my head, a tuple would just be 1 entry on the expr stack (so that means each entry on the expr stack is a `Vector<Pair<Def,Type>>` where `Def` is your compiler SSA definition) so you still want the binary `br0`/`br` split.

rossberg-chromium: @lukewagner, with tuples, the nullary case would be covered by the nullary tuple, and `br0` should then behave like a shorthand for `(br (tuple))`, me thinks. That only works if you generalise from syntactic to semantic arity matching, which is useful for other reasons. Also, I'm not sure you actually want to treat tuples as single entries on the expression stack, but rather as multiple entries.

But either way, right now my main concern is not to lock us in.

lukewagner: At the AST/semantic level, yes, but in the binary encoding the difference would be that `br0` never had a child subtree while `br` always did.  It's tempting to just have this be a "binary encoding detail", but, as with ast.ml, I think it's really useful to have a direct correspondence between the AST and binary encoding rather than the binary encoding introducing something "interesting" (such as multiple opcodes for a single operator or an additional hidden arity immediate).  I appreciate that there's more to discuss here, but it seems like the only bad option is reverting #180 since that breaks post-order/bottom-up.  It seems like `br0`/`br` is a conservative way to start that doesn't preclude tuples but is also something we could change later (before v.1, but after more significant tuple experimentation) if we decide we want 1 n-ary `br` instead of `br0`/`br`.

JSStats: @lukewagner Reverting https://github.com/WebAssembly/spec/pull/180 is orthogonal to your pre/post order concerns. To keep the AST clean for both pre/post order it is simply necessary that the encoding of each operator stands on its own and does not depend on context either way - I would support a PR to make this a design principle. For example, we can revert https://github.com/WebAssembly/spec/pull/180 and use the v8 encoding for breaks and fill with the nop, and this works pre/post order. It was only the high interest in the explicit encoding efficiency at the AST level that demanded br0.

Expression results need to be one entry on the expression stack, otherwise you have a stack machine!  This is needed to be able to discard them as a set and to consume only the first-n, and to pass them through the 'parameterized' operators. For example: to use `if_else` and `select` to choose between sets of values, and to be able to pass a set from a block, and with a good `br_if` design to be able to direct a set of values.

@sunfishcode It affects usability of the AST if the consumers of `br_if` values can not independently decide to discard values - as you seem to have already experienced.

lukewagner: @rossberg-chromium Actually, I went back and read your comments again and checked out the original multi-value support again and I see what you're getting at.  Yes, reverting #180 makes sense assuming that we pretend that `expr option` is actually an `expr list` and then encode the list length as an immediate in the binary encoding.  This would be enough for a post-order/bottom-up algo.  To better stake out the ground for multi-values after the MVP, I wonder if we should also go back to having `expr_type` be an `expr list` again (whose lengths were necessarily either 0 or 1) so that we didn't have to wave our hands with the `expr option`...

JSStats: @lukewagner Encoding the argument list length in the operator encoding would solve your pre/post order concerns, but it adds an extra byte to all break operator encodings even those with one expression, and is an even worse encoding efficiency outcome than the current v8 encoding.

Further, I fail to see why the break operators should be the only places to construct a tuple/values. It's just as useful for the arguments to `if_else`, `select`, the `block` fall-through expression, etc. Why should we need `(block $l (br $l v1 v2 v3))` to construct a multi-value expression result - it just frustrates producers and is technically less efficient to encode.

Can anyone find fault with the br0/br proposal and with having separate tuple/values constructors in future for mv support? Fail-early-please.

JSStats: @sunfishcode Here might be one example for which the consumers expect different numbers of result values. 

```
(func (param $save i32)
  (block $top
    (i32.store (addr) (br_if $top (call $i32_side_effect_fn) (get_local $save)))))
```

I do wonder if we are overdoing the expressions optimizations. There might remain a lot of cases that still require writing state to locals between blocks and the stores/loads might still need to be optimized away well by the runtime compiler, so might the overall design be simpler without a value expression for br_if at all. There is a seemingly efficient wasm SSA pass, and from there it doesn't seem too hard to optimize inter-block copy operations away. I hope there is enough support to land this change, with more flexible checks, and perhaps this could all be re-visited later and compared to a minimalist design. Shall be interesting to see how useful this is in practice.

rossberg-chromium: @lukewagner, all sounds very good to me. I've always thought of these options as "lists with a length limit" (and the design doc actually describes them as such). Reverting them to actual lists in ml-proto would perhaps make that clearer, and as you say, would match the binary encoding that we may be converging on.

@JSStats, once we add multi-value support post-MVP, I agree that we should add a `tuple` or `values` operator. Re stack machine: yes indeed, and that would be one possible way to describe the semantics. There is no problem with that, because the language is typed, so this stack is guaranteed to always be manipulated consistently. Basically, the benefit of making tuples/multi-values second-class is that both interpretations/implementation strategies are interchangeable, because a program cannot observe the difference. In fact, maintaining that equivalence is a basic litmus test for getting the multi-value semantics right.

lukewagner: @rossberg-chromium Ok, so I guess that PR would subsume these others.  The original proposal didn't have `br_if`, but the reduce would have `br_if` return its value, right?

rossberg-chromium: @lukewagner, I kinda lost track of which PR(s) you are referring to right now, but I suppose the answer is: yes, it would or could. :)

lukewagner: By any lucky chance are you so inclined to write such a PR :) ?

sunfishcode:
Closing this for now, so we can work out the multiple result design first.

ml-proto/spec/check.ml
JSStats: @rossberg-chromium Re: stack machine: yes, there are certainly CL implementations that are stack machines and these support multiple value expressions, but they keep a count of values on the stack at some points or a frame pointer as a reference position, but I think you are right that wasm would not need to do this because the runtime compiler always knows the number of actual and expected values at compile time, so even if wasm allows a mismatch between expected values of `br_if` consumers it would still know to pop any unused values. The only remaining concern might be if people expected to execute the AST literally as a byte code and then might need to keep track of the number of expression values at runtime (as opposed to the runtime compiler which can account for them at compile time).

I notice that `if` desugars to `if_else` in the ml-proto and that it can return a different number of values depending on the condition, and it might need to be the responsibility of the operators to return a consistent number of values here so `if` would want to return the empty-values irrespective of condition. Also `if_else` and `selection` might want to be defined to return a consistent number of values irrespective of the condition.

rossberg-chromium: @JSStats, under the tuple interpretation used in ml-proto's eval, it is neither necessary nor desirable to do that. Only a stack machine would require adjustments. But the approach you suggest would be the wrong one. Instead, you'd need to adjust whenever the type system uses subsumption (i.e., "forgets" a non-empty value in `check_type`). That occurs in more places than just branches, e.g. `(block (i32.const 0) (i32.const 1))` requires dropping the first value. Branches do not need any special treatment.

JSStats: @rossberg-chromium Sorry, it's not clear to me what you are referring to as 'neither necessary nor desirable to do that'. As I understand your position, the 'tuple' proposal, it would only be possible to either discard the entire tuple or consume all the values, whereas for my 'values' proposal it would be possible to consume any length head of the values which includes both none and all as limiting cases and thus subsumes your tuple proposal and I believe can be demonstrate to be practically more useful and efficiently at encoding the lowered wasm for at least some languages. I believe these two proposals can be judged on their efficiency at encoding lowered code, and that a 'right' or 'wrong' subjective judgement is not necessary, and perhaps we could stop using such words and I apologise if I have framed some aspects of this issue in those terms.

In any case I still fail to see how tracking the number of values can be avoided. As noted, the runtime compiler seems to know the number of values at compile time so can optimize away tracking the number of values, but I fail to see how an interpreter of the literal AST could avoid tracking the number of values, be it the number of values in a tuple or the number of values in a values list? Even if the interpreter boxed the tuple, the tuple object would be encoding the number of values. Sorry if I am missing some obvious optimization that the tuple strategy supports, but perhaps you could share this if there is one?

I certainly agree that accounting for differences between the number of expected and actual values is a general operation and not specific to branches, and I just gave this as an example because this is the topic of this issue. However `bf_if` does seem to be a unique operation in wasm as it is the only operation that has multiple consumers of its result values and this might make it uniquely useful for discussing some aspects of the type system. Every other case of multiple references seems to go through variables. I support each consumer being handled separately for `br_if` so that one consumer can drop the values and the other consume them, and for the reason that this is more flexible and can thus increase the efficiency of coding some patterns.

rossberg-chromium: @JSStats, you are now talking about a language extension again, I wasn't. But even with more general subtyping like you propose, you wouldn't need or want to adjust tuples in the operational semantics. In a tuple-based implementation, subtyping just works, without coercing, because there is no operation that can ever observe excess values. They just get ignored when decomposing a tuple. A real-world implementation might of course want to implement that differently, e.g. a stack-based one has to.

PR #239, opened 2016-02-10
Add failing test case.

mbebenita:
The first three tableswitches appear to be working, however the last one fails. It should follow the default branch target at `index 1` which should exit the outer block. Strangely enough, if you change `(table (br 0)) (br 1)` to `(table (br 1)) (br 1)` then it works.

### DIFF
--- a/ml-proto/test/switch.wast
+++ b/ml-proto/test/switch.wast
@@ -80,10 +80,35 @@
     (get_local $x)
   )
 
+  ;; Index Break
+  (func $index_break (result i32)
+    (local $x i32)
+    (tableswitch (i32.const 0)
+      (table (case 0)) (case 0)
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
Nice find! See #240 for the fix (the test case there probably subsumes this patch).

mbebenita:
@rossberg-chromium great, thanks! 
It would still be nice to add a test case where a `tableswitch` has several cases, yet it breaks to an outer block. I ran into this bug because Binaryen was treating `(br 0)` as a `(case 0)` when the `tableswitch` had at least one case.

rossberg-chromium:
Sure, feel free to adjust this patch once #240 has landed.

rossberg-chromium:
Closing this, since we are dropping tableswitch anyway.

PR #241, opened 2016-02-13
Test that the start function index space does not include imports.

JSStats:
The v8 encoding appears to be using it's imports-and-functions index
space for the start function index encoding, so these tests check that
tools are mapping it correctly.


haxxpop:
Should we also add the `expected-output/start.wast.log` file ?

JSStats:
The other checks should detect an index space error, the start function can not have operands and the import does.

PR #244, opened 2016-02-17
Implement env.exit and env.abort

rossberg-chromium:
Should make the waterfall significantly happier.


jfbastien:
Awesome, thanks! Once this is in the tests won't actually execute because I'm not generating the .wast files with `(start ...)` yet. I'll wait for your change to make it in, and then I'll add start :-)

sunfishcode:
"Env" sounds like a fairly non-temporary name. What are the future plans for this module?

Env.abort: Is the existing `unreachable` instruction insufficient?

Env.exit: I haven't thought about this deeply, but should we make this an instruction too? In a JS context, one can just call out to JS, have it throw an exception, and unwind all the wasm frames. However in a non-JS context, there isn't a way to say "unwind all the stack frames and exit" short of trapping, but that has the additional effect of indicating a failure of some kind.

jfbastien:
FWIW when we get a nice wasm libc that we all agree on (and our tests link to it) these will change names and be __syscall_abort and __syscall_exit or something like that. It'll be a simple rename from what @rossberg-chromium adds here.

The libc won't be Standard since it'll all be user-space, but it'l be what we end up agreeing to use.

sunfishcode:
If we know the names we want, let's just use those from the start. If we don't and want to defer making that decision, let's use sufficiently temporary-looking names so that we don't get confused.

Env.abort: Is the existing unreachable instruction insufficient?

Env.exit: I haven't thought about this deeply, but should we make this an instruction too? In a JS context, one can just call out to JS, have it throw an exception, and unwind all the wasm frames. However in a non-JS context, there isn't a way to say "unwind all the stack frames and exit" short of trapping, but that has the additional effect of indicating a failure of some kind.

rossberg-chromium:
@sunfishcode, the point of this change isn't expressiveness, but to be able to run various tests from the waterfall, many of which try to import one of these functions. I don't care about the module name, although I actually agree that `env` is overly generic. @jfbastien, can't we call it `libc` or `libcbase` or something?

sunfishcode:
@rossberg-chromium So this is a hack to underpin another hack. There may be a motivation for this, but I will at least request that hacks be clearly labeled for what they are, because it's very confusing to have so many hacks mixed in with the real design when we're trying to work on the real design at the same time.

`libc` or most any name including `libc` would be similarly confusing for me. `waterfall_scaffolding` would be less problematic.

lukewagner:
We already have `spectest` for our grab bag of stuff that's good for testing/automation; can we just put them there?

rossberg-chromium:
@sunfishcode, @lukewagner, "builtins" mock a useful execution environment, they are not part of the spec. Naming needs to match the other relevant testing/execution environments, otherwise it's useless. The `spectest` module does not exist anywhere else (and rightfully so, I would argue).

Eventually we should find policies for allocating names in the global namespace of modules, but is that a relevant concern right now? Should development block on it? We can easily adjust tests later.

lukewagner:
Agreed on it not being part of the spec... but how is that different than `spectest`?  that is, why do we need two grabbags?

rossberg-chromium:
It's not a grab bag, it's mocking (a subset of) a concrete module that the libc tests on the waterfall import. It does not have any relation to `spectest`.

lukewagner:
Ah, I see the distinction now.  Seems fine then.

kripken:
Putting it on "env" would collide with existing code, that sends exit or abort there, as e.g. asm2wasm does.

I think `waterfall_scaffolding` makes sense, or `waterfalltest` (to parallel `spectest`). If this is for the waterfall specifically, then that seems like it should be in the name.

jfbastien:
"env" comes from binaryen:
https://github.com/WebAssembly/binaryen/blob/master/src/s2wasm.h#L757

I don't really care what it is, but I think we're putting too much thought into something pretty simple.

kripken:
The env there is for real imports arriving from the outside, though. And exit and abort will in fact be provided in a lot of existing code. Whereas this adds env.exit and abort to the actual spec, if I understand correctly? In other words, this moves a userspace issue into something to be standardized? That doesn't feel right.

jfbastien:
I'm fine with changing the name to whatever is decided upon, and adding support for it in binaryen. Just pick a name.

rossberg-chromium:
@kripken, "builtins" are not part of the spec. Maybe the name is misleading. They are part of the `host` directory of ml-proto, used to mock an embedding environment, i.e., real imports from the outside as you say. In ml-proto, it's the only way to provide external modules.

rossberg-chromium:
Okay, I refactored the handling of imports to make it (a) more modular and configurable, (b) clearer that the actual modules provided by the host are custom. Renamed `builtins` to `imports`, and there now is a new directory `host/import` where host-provided modules live. They can be registered in `main`.

kripken:
Perhaps it's worth documenting the distinction that the stuff in `host/` is not part of the spec? As this is the spec repo, my assumption was that everything here is part of the spec.

rossberg-chromium:
@kripken, done.

PTAL

kripken:
lgtm, thanks.

jfbastien:
lgtm as well.

ml-proto/host/import/env.ml
```diff
@@ -0,0 +1,39 @@
+open Values
+open Types
+
```
sunfishcode: The name `env` is confusing to me. The name `waterfalltest` was suggested in this PR; I'd be ok with that.

sunfishcode: Also, please add a top-level comment to this file explaining what its purpose is. The text in the README.md is a good introduction, but we should be clear about what these various builtin modules are for.

ml-proto/host/import/spectest.ml
```diff
@@ -0,0 +1,12 @@
+open Types
+
+
```
sunfishcode: A top-level comment would be useful here as well. Something like "The spectest module provides functions that support the specialized testsuite execution environment."

ml-proto/host/import/env.ml
jfbastien: This isn't just for the waterfall.

sunfishcode: What is it for?

jfbastien: Running compiled code as a stopgap until we have agreement on what libc should look like. When we have an agreed libc we'll likely do something like syscalls, and abort/exit will be simple renames. libc still won't be "standard", but it'll be an ad-hoc thing we end up using.

I'll be testing this on the waterfall, but something that only works on the waterfall is a silly idea so I want to avoid silly.

sunfishcode: Please put this in a comment in the file then, so that we can all know what it is we're dealing with here.

Also, please rename the module to `stopgap` or something similarly descriptive.

rossberg-chromium: Added comments to `spectest` and `env` modules explaining their purpose. PTAL

@sunfishcode, as explained before, the `env` module emulates the respective module imported by Binaryen code. Renaming it here would defeat the purpose. Binaryen would have to decide that.

kripken:
Thinking about this some more, I'm not sure of the scope plans here. This adds `exit` and `abort`, what if next we see some tests that would run if we also added `memcpy`, `printf`, `sbrk`, and so forth?

If there's a clear line that explains why `exit` and `abort` but not the rest of libc, then I think it's worth mentioning in the README.

jfbastien:
@kripken same reasoning as for:
https://github.com/WebAssembly/binaryen/blob/ed567a4f49d3296dcf91bdb966c0deb077dacff7/src/binaryen-shell.cpp#L130
The torture tests are self-checking, and call abort / exit on success / failure. Until we have a from-the-ground-up wasm libc which isn't contended we need a simple way to run tests (more than the torture tests IMO) and this is a basic first step.

We're *waayyyyy* over thinking things right now.

kripken:
I agree let's not overthink this. Just seems worth documenting "this is stuff specifically for the torture tests, and not intended to include any significant parts of libc", or something along those lines. I added some docs to binaryen now for that one hack we have.

rossberg-chromium:
Okay, is anybody still opposed to landing this?

sunfishcode:
"A subset of the `env` module as used by Binaryen." isn't a clear explanation of what this module is for. Would it be possible to put JF's description, "Running compiled code as a stopgap until we have agreement on what libc should look like." in the comment in the file?

Also, binaryen is changeable. We can rename things. "env" is a very generic name. I think we should rename this module, and teach Binaryen about the new name, to help avoid ending up making something which outlives its intended use, such after all this is, as JF says, a stopgap.

rossberg-chromium:
Tweaked comment to express the purpose and merging. Landing, but happy to keep this in sync with any name change Binaryen chooses to implement.

sunfishcode:
Why are you so unwilling to put what you're actually doing here in a comment in the file?

PR #245, opened 2016-02-18
Add a few more floating-point tests.

sunfishcode:
* Test that demote rounds small values near zero to the appropriately
   signed zero.
 * Reorganize the floating-point literal tests and cover both hexadecimal and
   decimal versions of all the major boundary numbers.


rossberg-chromium:
lgtm

PR #250, opened 2016-02-24
Capture JF's description of env.ml in a comment.

sunfishcode:
This just adds a comment, using wording that JF used to explain what this file is for, in [this comment](https://github.com/WebAssembly/spec/pull/244#discussion_r53517759).


rossberg-chromium:
Note that pretty much _any_ host module we are ever going to include in ml-proto is going to be an emulation stopgap in the same way. That's the sole purpose of the mechanism. But if spelling that out for each one individually is important to you, LGTM.

kripken:
Hopefully though we won't need any more host modules in the spec repo, and this will be the only stopgap?

lgtm

rossberg-chromium:
Suppose you want to emulate pieces of the browser environment, in order to run tests that interact with the DOM or WebGL.

PR #251, opened 2016-02-24
Allow empty blocks

rossberg-chromium:
At the same time, make it explicit in the data type that blocks are non-empty in the kernel.

Also, add a couple of basic tests for blocks.


lukewagner:
lgtm

PR #255, opened 2016-02-25
Moved return-value checks of imported functions to imports.wast

Teemperor:
Cleaner version of #167


rossberg-chromium:
LGTM, thanks.

PR #256, opened 2016-02-25
Fix `return` representation in Ast

rossberg-chromium:
`Ast.Return` improperly had a label argument to support its desugaring. That is now removed, and the logic properly encapsulated in desugaring.


lukewagner:
Ah, nice, I hadn't even noticed that `var` straggler.

rossberg-chromium:
@lukewagner, is that an LGTM? :)

Btw, I screwed up my branching, so that #257 is included in this branch. Sorry about that, just look at the first commit.

lukewagner:
oops, yes :)

rossberg-chromium:
Thanks :)

ml-proto/test/memory.wast
```diff
@@ -95,7 +95,7 @@
     (loop
       (if
         (i32.eq (get_local 0) (i32.const 0))
-        (br 1)
+        (br 2)
```
kripken: This pull is for `return` - why did these `br`s change?

And where does 2 go? the only `block`s or `loops` above this `br` are the one `loop`, which would take indexes 0 and 1?

binji: if has an implicit label now too. Change was here: 05688003a4d5abf31781141f9b3fde588e454414. Not sure why it didn't break then.

kripken: thanks! now i see

rossberg-chromium: Sorry for this, I messed up the branching, and accidentally merged the `if` changes into the `return` branch instead of master. This line is part of the `if` changes (`if` now has an implicit block).

kripken: is it the if that has an implicit block, or the then and else branches that do? because I see they can have a label name on them now? must it be the same for both of them if there is?

rossberg-chromium: The branches have implicit blocks, so the labels go there (also, that makes it clear that they are not in scope in the condition). They can be named differently, see the examples in test/labels.wast.

kripken: got it, thanks!

PR #257, opened 2016-02-25
Adjust `if` to match post-order format.

rossberg-chromium:
The post-order binary format induces minor changes to `if`:

- There is only one opcode for `if`, with and without else-branch.
- The branches can have sequences of expressions (they are effectively blocks).

This PR reflects these generalisations in the AST and the S-expression format. Unfortunately, this required tagging the branch blocks in the S-expression syntax, so that the both lists can be distinguished. The tags can be omitted if the sequence consists of exactly one expression, thereby keeping the previous concrete syntax valid. See README for details.

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -97,10 +97,10 @@ cvtop: trunc_s | trunc_u | extend_s | extend_u | ...
 
 expr:
   ( nop )
-  ( block <name>? <expr>+ )
-  ( if_else <expr> <expr> <expr> )
-  ( if <expr> <expr> )                           ;; = (if_else <expr> <expr> (nop))
-  ( br_if <expr> <var> <expr>?)                  ;; = (if_else <expr> (br <var> <expr>?) (block <expr>? (nop)))
+  ( block <name>? <expr>* )
+  ( if <expr> ( then <name>? <expr>* ) ( else <name>? <expr>* )? )
+  ( if <expr1> <expr2> <expr3>? )                ;; = (if <expr1> (then <expr2>) (else <expr3>))
+  ( br_if <expr> <var> <expr>? )
   ( loop <name1>? <name2>? <expr>* )             ;; = (block <name1>? (loop <name2>? (block <expr>*)))
   ( br <var> <expr>? )
   ( return <expr>? )                             ;; = (br <current_depth> <expr>?)
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -99,7 +99,7 @@ expr:
   ( nop )
   ( block <name>? <expr>* )
   ( if <expr> ( then <name>? <expr>* ) ( else <name>? <expr>* )? )
-  ( if <expr1> <expr2> <expr3>? )                ;; = (if <expr1> (then <expr2>) (else <expr3>))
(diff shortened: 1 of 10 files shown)


JSStats:
My understand of an advantage of the post-order encoding is that it reduces the need for the decoder to maintain a production stack of nodes still pending reduction. For some operations such as `if` and `if_else` the decoder needs to make ordered decisions. For example: to split or merge SSA environments; or to take different control flow paths. There appears to be a strong support to do this in one pass.

This patch appears to be in support of moving state that the decoder will need to track anyway into explicit opcodes. For example to encode it as: `<cond> if <exp1> then <exp2> else`. A decoder would appear to still need to maintain a stack here, if only to validate the pattern, but probably also to manage the SSA state etc.

If using a post-order encoding then could you please consider a mixed post/pre order encoding that keeps the pre-order encoding where pre-order information needs to be maintained by the decoder anyway, and for the reason of avoiding explicit redundant information in the binary encoding.

If using a post-order encoding for `if` and `if_else then could you please consider at least taking advantage of the opportunities and accepting a sequence of expressions in the `then` and `else` operations (implicit blocks with a block depth).

I would note that my experiments show that replacing `br_if` with `if` and `if_else` can usefully reduce the binary file size, so even though they could be avoided they do have utility and could be expected to be frequent. This advantage might be reversed if `if` and `if_else` are less efficient to encode.

If a pre-order encoding wins on the compressed file size and the wasm binary format adopts a post-order encoding then the pipeline may well be permuting the order which I guess would be at least as costly and resource intensive as the decoder tracking the productions stack. A small win for the decoder might not justify a decision on pre/post order when put in perspective.

JSStats:
Also if making this change then could you please consider supporting `(if <cond> (else <expr>*))` which also gives us `unless`, as requested in discussions in https://github.com/WebAssembly/design/pull/481

JSStats:
Sorry, I see the patch is already 'taking advantage of the opportunities and accepting a sequence of expressions in the `then` and `else` operations' along with names for the blocks. 

In terms of the README syntax could you please also consider:
`( if <expr> ( then <name>? <expr>* )? ( else <name>? <expr>* )? )`

rossberg-chromium:
@JSStats, the binary format and this patch already allow sequences of expressions in the branches, please reread the patch description. As for dropping the then-branch, the binary format will not be able to support that, it effectively requires an empty sequence. Could allow it as a shorthand in the text format, but not sure it's worth it.

JSStats:
@rossberg-chromium Yes, thanks, missed that on first reading, sorry again. Shall be interesting to see how supporting block here helps - I do see the pattern and think it will help.

It's a bit hard to comment without knowing what people actually have in mind for the post-fix binary encoding for `if`, but I had assumed it would be post-fix: `<cond> if <exp1> <exp2> ... then <exp3> <exp4> ... else` in which case it might be possible to cleanly allow either the `then` or `else` operators to be missing. Do you have something else in mind?

JSStats:
Sorry again, obviously it would not be possible to allow the final closing operator to be missing, so one more operator might be needed. For example:

```
<cond> if <exp1> <exp2> ... then_endif
```
```
<cond> if <exp2> <exp2> ... else_endif
```
```
<cond> if <exp1> <exp2> ... then <exp3> <exp4> ... else_endif
```

lukewagner:
lgtm

PR #261, opened 2016-03-12
Binary encoder + various cleanup

rossberg-chromium:
First step towards handling the binary format in the spec. The patch was getting large, so I decided to split it up. This part includes the encoder, and several additions to make it useful. For example, you can now invoke the interpreter like
```
wasm module.wast -o module.wasm
```
to convert from text to binary. In a while, the same will also be possible in the inverse direction.

I also extended the script language with `(input <file>)` and `(output <file>)` commands. Both are supposed to be able to handle both wast and wasm eventually. The former allows including other scripts or binary modules, the latter allows conversion as part of a script.

Finally, the command line now supports an `-e <script>` option, which enables to give commands directly. This is useful, for example, when intermingled with binary module arguments, e.g., to invoke exports:

```
wasm module.wasm -e '(invoke "foo")'
```

Caveat: the binary format implemented in this patch already uses post-order. There currently is no compatible consumer (the decoder for the follow-up patch isn't quite ready yet), so this is completely untested. Uploading it anyway, to factor reviewing a little bit. Tests will follow once the decoder is ready.

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -81,7 +81,7 @@ wasm module.wasm -o module.wast
 Finally, the option `-e` allows to provide arbitrary script commands directly on the command line. For example:
 
 ```
-wasm module.wasm -e "(invoke \"foo\")"
+wasm module.wasm -e '(invoke "foo")'
 ```
 
 If neither a file nor any of the previous options is given, you'll land in the REPL and can enter script commands interactively. You can also get into the REPL by explicitly passing `-` as a file name. You can do that in combination to giving a module file, so that you can then invoke its exports interactively, e.g.:
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -11,7 +11,7 @@ Currently, it can
 * *encode* the binary format,
 * *prettyprint* the S-expression format (work in progress).
 
-The S-expression format is a (very dumb) form of *script* that cannot just define a module, but also batch a sequence of invocations, assertions, and conversions. As such it is a superset of the binary format, with the additional functionality purely intended as testing infrastructure.
+The S-expression format is a (very dumb) form of *script* that cannot just define a module, but in fact a sequence of them, and a batch of invocations, assertions, and conversions to each one. As such it is a superset of the binary format, with the additional functionality purely intended as testing infrastructure. (See [below][#scripts] for details.)
 
 The interpreter can also be run as a REPL, allowing to enter pieces of scripts interactively.
 
@@ -187,13 +187,13 @@ cmd:
   ( assert_return_nan (invoke <name> <expr>* ))        ;; assert return with floating point nan result of invocation
   ( assert_trap (invoke <name> <expr>* ) <failure> )   ;; assert invocation traps with given failure string
   ( assert_invalid <module> <failure> )                ;; assert invalid module with given failure string
-  ( input <string> )                                   ;; read script from file
+  ( input <string> )                                   ;; read script or module from file
   ( output <string> )                                  ;; output module to file
 ```
 
-Invocation, most assertions, and output is only possible after a module has been defined.
+Commands are executed in sequence. Invocation, assertions, and output apply to the most recently defined module (the _current_ module), and are only possible after a module has been defined. Note that there only ever is one current module, the different module definitions cannot interact.
 
-The input and output commands determine the requested file format from the file name extension. They can handle both `.wast` and `.wasm` files.
+The input and output commands determine the requested file format from the file name extension. They can handle both `.wast` and `.wasm` files. In the case of input, a `.wast` script will be recursively executed.
 
 Again, this is only a meta-level for testing, and not a part of the language proper.
 
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -11,7 +11,7 @@ Currently, it can
 * *encode* the binary format,
 * *prettyprint* the S-expression format (work in progress).
 
-The S-expression format is a (very dumb) form of *script* that cannot just define a module, but in fact a sequence of them, and a batch of invocations, assertions, and conversions to each one. As such it is a superset of the binary format, with the additional functionality purely intended as testing infrastructure. (See [below][#scripts] for details.)
+The S-expression format is a (very dumb) form of *script* that cannot just define a module, but in fact a sequence of them, and a batch of invocations, assertions, and conversions to each one. As such it is different from the binary format, with the additional functionality purely intended as testing infrastructure. (See [below](#scripts) for details.)
 
 The interpreter can also be run as a REPL, allowing to enter pieces of scripts interactively.
 
@@ -76,7 +76,7 @@ wasm module.wast -o module.wasm
 wasm module.wasm -o module.wast
 ```
 
-(The second direction is work in progress.)
+In the second case, the produced script contains exactly one module definition (work in progress).
 
 Finally, the option `-e` allows to provide arbitrary script commands directly on the command line. For example:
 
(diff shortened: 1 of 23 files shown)


ml-proto/spec/kernel.ml
```diff
@@ -119,7 +119,7 @@ and func' =
 type memory = memory' Source.phrase
 and memory' =
 {
-  initial : Memory.size;
+  min : Memory.size;
```
sunfishcode: It's perhaps subtle, but why is "initial" changing to "min" here and elsewhere? Apparently BinaryFormat.md calls it "minimum", but AstSemantics.md still calls it "initial".

ml-proto/README.md
```diff
@@ -5,10 +5,13 @@ This repository implements a prototypical reference interpreter for WebAssembly.
...
 * *parse* a simple S-expression format,
+* *decode* the binary format (work in progress),
 * *validate* modules defined in it,
-* *execute* invocations to functions exported by a module.
+* *execute* invocations to functions exported by a module,
+* *encode* the binary format,
+* *prettyprint* the S-expression format (work in progress).
 
-The file format is a (very dumb) form of *script* that cannot just define a module, but also batch a sequence of invocations.
+The S-expression format is a (very dumb) form of *script* that cannot just define a module, but also batch a sequence of invocations, assertions, and conversions. As such it is a superset of the binary format, with the additional functionality purely intended as testing infrastructure.
```
sunfishcode: Another useful thing to point out here is that .wast files can define a sequence of *modules* too, each with a sequence of tests following them.

Also, a lot of people (myself included) have been confused about .wast files. For me, they made sense when I realized that .wast files are not "module" files per se, or even containers for modules. If they define multiple modules, the modules don't interact; each module just replaces the previous one to be the "current" module -- semantics that would be surprising for any purpose other than testing. I'd suggest dropping the "superset" language, and saying something like this:

"As it processes its command-line arguments, it keeps track of a *current* module. When it encounters a module file, it decodes the module and assigns it as the current. When it encounters a test script file, it executes the test script. The `module` directive in the test script language creates a module and assigns it as current. Scripts may also contain invocations, assertions, and conversions, which perform tests on the current module."

ml-proto/README.md
```diff
@@ -61,17 +64,33 @@ Either way, in order to run the test suite you'll need to have Python installed.
...
-See `wasm -h` for (the few) options.
+(The second direction is work in progress.)
+
+Finally, the option `-e` allows to provide arbitrary script commands directly on the command line. For example:
+
+```
+wasm module.wasm -e "(invoke \"foo\")"
+```
+
+If neither a file nor any of the previous options is given, you'll land in the REPL and can enter script commands interactively. You can also get into the REPL by explicitly passing `-` as a file name. You can do that in combination to giving a module file, so that you can then invoke its exports interactively, e.g.:
```
sunfishcode: And if you make the "current" module concept explicit above, you can describe the behavior of `-` when it follows a script in terms of it: it just launches a REPL on the current module, regardless of how it was defined.

ml-proto/host/encode.ml
```diff
@@ -0,0 +1,417 @@
+(* Version *)
+
+let version = 11
```
sunfishcode: Elsewhere, the project documents the version as hex, so this would be more clear as `0xb`.

ml-proto/README.md
```diff
@@ -168,9 +187,13 @@ cmd:
...
   ( assert_trap (invoke <name> <expr>* ) <failure> )   ;; assert invocation traps with given failure string
   ( assert_invalid <module> <failure> )                ;; assert invalid module with given failure string
+  ( input <string> )                                   ;; read script from file
+  ( output <string> )                                  ;; output module to file
 ```
 
-Invocation is only possible after a module has been defined.
+Invocation, most assertions, and output is only possible after a module has been defined.
+
+The input and output commands determine the requested file format from the file name extension. They can handle both `.wast` and `.wasm` files.
```
sunfishcode: From this description, I might guess that when `input` gets a .wast file, it just executes the whole script, right?

ml-proto/README.md
```diff
@@ -168,9 +187,13 @@ cmd:
   ( assert_return_nan (invoke <name> <expr>* ))        ;; assert return with floating point nan result of invocation
   ( assert_trap (invoke <name> <expr>* ) <failure> )   ;; assert invocation traps with given failure string
   ( assert_invalid <module> <failure> )                ;; assert invalid module with given failure string
+  ( input <string> )                                   ;; read script from file
```
sunfishcode: If I understand correctly, the comment should say something like "read script or module from file".

ml-proto/README.md
rossberg-chromium: Good point. Added most of that to the section on scripts, and only a short remark and a forward reference here.

ml-proto/README.md
rossberg-chromium: Yep, done.

ml-proto/README.md
rossberg-chromium: Yes, added a clarifying sentence.

ml-proto/host/encode.ml
rossberg-chromium: Done.

ml-proto/spec/kernel.ml
rossberg-chromium: Yes, fair point. The naming in BinaryFormat is more symmetric, so I chose that. We should resolve it either way, happy to change back if that's preferred.

ml-proto/README.md
```diff
@@ -5,10 +5,13 @@ This repository implements a prototypical reference interpreter for WebAssembly.
...
 * *parse* a simple S-expression format,
+* *decode* the binary format (work in progress),
 * *validate* modules defined in it,
-* *execute* invocations to functions exported by a module.
+* *execute* invocations to functions exported by a module,
+* *encode* the binary format,
+* *prettyprint* the S-expression format (work in progress).
 
-The file format is a (very dumb) form of *script* that cannot just define a module, but also batch a sequence of invocations.
+The S-expression format is a (very dumb) form of *script* that cannot just define a module, but in fact a sequence of them, and a batch of invocations, assertions, and conversions to each one. As such it is a superset of the binary format, with the additional functionality purely intended as testing infrastructure. (See [below][#scripts] for details.)
```
sunfishcode: I'll make one more argument against calling .wast a "superset" of a .wasm:

From a user perspective, as opposed to a ml-proto implementation perspective, .wast files are not modules; they are scripts that have a way to define modules inline. They don't satisfy the "isa" relationship, except via an awkward path where all modules but one are silently discarded, behavior which isn't useful outside of artificial testcases.

ml-proto/README.md
```diff
@@ -61,17 +64,33 @@ Either way, in order to run the test suite you'll need to have Python installed.
...
 
-where `file` is a script file (see below) to be run. If no file is given, you'll get into the REPL and can enter script commands interactively. You can also get into the REPL by explicitly passing `-` as a file name. You can do that in combination to giving a module file, so that you can then invoke its exports interactively, e.g.:
+where `file`, depending on its extension, either should be an S-expression script file (see below) to be run, or a binary module file to be loaded.
+
+A file prefixed by `-o` is taken to be an output file. Depending on its extension, this will write out the preceding module definition in either S-expression or binary format. This option can be used to convert between the two in both directions, e.g.:
 
 ```
-./wasm module.wast -
+wasm module.wast -o module.wasm
+wasm module.wasm -o module.wast
```
sunfishcode: I would find it useful to have a clarification stating that when the output is a .wast file, it will contain a script which contains a command defining the module, to help clarify that .wast files are not module files.

rossberg-chromium: Done.

ml-proto/README.md
rossberg-chromium: Avoiding the word "superset" now.

rossberg-chromium:
Superseded by #268.

PR #264, opened 2016-03-19
Add br_if and br_table to left-to-right.wast

pjuftring:
This tests if br_if and br_table arguments are evaluated left-to-right.
Until #227 is solved, a small hack circumvents any "arity mismatch" error
messages.


rossberg-chromium:
LGTM

pjuftring:
I removed the hack, because with #271 it is no longer needed.

rossberg-chromium:
Cool, thanks.

PR #265, opened 2016-03-21
Add support for unsigned decimals

### LINKED ISSUE #225: ml-proto does not accept the full range of unsigned i32 or i64 constants
The ml-proto does not appear to support unsigned i32 or i64 constants with the high bit set, rather requires them to be signed numbers. Both wasm-as and sexp-wasm accept either signed or unsigned.

pjuftring:
This fixes #225

### DIFF
--- a/ml-proto/spec/int.ml
+++ b/ml-proto/spec/int.ml
@@ -208,7 +208,19 @@ struct
   let ge_s x y = x >= y
   let ge_u x y = cmp_u x (>=) y
 
-  let of_string = Rep.of_string
+  (* to_string supports unsigned decimals *)
+  let of_string x =
+    let unsigned_decimal x =
+      let l = String.length x in
+      Rep.add
+        (Rep.mul
+          (Rep.of_string (String.sub x 0 (l - 1)))
+          (Rep.of_int 10))
+        (Rep.of_string (String.sub x (l - 1) 1))
+    in
+    try Rep.of_string x with
+      Failure _ -> unsigned_decimal x;;
+      
   let to_string = Rep.to_string
 
   let of_int = Rep.of_int
--- a/ml-proto/spec/int.ml
+++ b/ml-proto/spec/int.ml
@@ -211,15 +211,31 @@ struct
   (* to_string supports unsigned decimals *)
   let of_string x =
     let unsigned_decimal x =
+      let ten = Rep.of_int 10 in
       let l = String.length x in
+      (* First overflow-check via String.length *)
+      if l > (if Rep.bitwidth = 32 then 10 else 20)
+        then raise (Failure "int_to_string");
+      let firstPart = Rep.of_string (String.sub x 0 (l - 1)) in
+      let secondPart = Rep.of_string (String.sub x (l - 1) 1) in
+      (* Negativity-check *)
+      if firstPart < Rep.zero
+        then raise (Failure "int_to_string");
+      (* Second overflow-check *)
+      if firstPart > (div_u Rep.minus_one ten)
+      || (firstPart = (div_u Rep.minus_one ten) && secondPart > (rem_u Rep.minus_one ten))
+        then raise (Failure "int_to_string");
       Rep.add
         (Rep.mul
-          (Rep.of_string (String.sub x 0 (l - 1)))
+          firstPart
           (Rep.of_int 10))
-        (Rep.of_string (String.sub x (l - 1) 1))
+        secondPart
     in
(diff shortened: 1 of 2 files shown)


ml-proto/spec/int.ml
```diff
@@ -208,7 +208,19 @@ struct
...
-  let of_string = Rep.of_string
+  (* to_string supports unsigned decimals *)
+  let of_string x =
+    let unsigned_decimal x =
+      let l = String.length x in
+      Rep.add
+        (Rep.mul
+          (Rep.of_string (String.sub x 0 (l - 1)))
+          (Rep.of_int 10))
+        (Rep.of_string (String.sub x (l - 1) 1))
```
sunfishcode: This code silently wraps on overflow (unless it runs out of stack space first ;-)). I think the spec should explicitly diagnose overflow.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,19 @@ struct
...
+    in
+    try Rep.of_string x with
+      Failure _ -> unsigned_decimal x;;
```
sunfishcode: My reading of host/lexer.mll suggests that this code code encounter a leading '-' sign, which doesn't appear to be expected in the unsigned_decimal code. Would it make sense to explicitly test for it here?

ml-proto/spec/int.ml
pjuftring: Didn't think of the overflow, thanks!

ml-proto/spec/int.ml
pjuftring: I think I'll prohibit negative numbers in unsigned_decimal. But I'm encountering some strange stuff: lexer.mll allows '+' in front of numbers, OCaml doesn't (atleast Rep.of_string doesn't), OCaml allows '_' within numbers, lexer.mll doesn't. (This would brake my current design.) Am I missing something, or is this some not intended behavior?

pjuftring:
I added some overflow-tests and added the possibility to have a plus-sign in front of numbers, which was defined in host/lexer.mll, but is not allowed by the builtin of_string function.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,35 @@ struct
...
   let ge_u x y = cmp_u x (>=) y
 
-  let of_string = Rep.of_string
+  (* to_string supports unsigned decimals *)
+  let of_string x =
+    let unsigned_decimal x =
+      let ten = Rep.of_int 10 in
+      let l = String.length x in
+      (* First overflow-check via String.length *)
+      if l > (if Rep.bitwidth = 32 then 10 else 20)
```
rossberg-chromium: Hm, this effectively hard-codes that we only instantiate with 32 and 64 bit. It would be more modular to have a generic criterium, e.g. `bitwidth / 3`.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,35 @@ struct
...
+        then raise (Failure "int_to_string");
+      let firstPart = Rep.of_string (String.sub x 0 (l - 1)) in
```
rossberg-chromium: Style nit: Ocaml prefers snake_casing

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,35 @@ struct
...
+      let secondPart = Rep.of_string (String.sub x (l - 1) 1) in
+      (* Negativity-check *)
+      if firstPart < Rep.zero
+        then raise (Failure "int_to_string");
+      (* Second overflow-check *)
+      if firstPart > (div_u Rep.minus_one ten)
```
rossberg-chromium: Nit: redundant parens around the div_u (same below)

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,35 @@ struct
...
+      || (firstPart = (div_u Rep.minus_one ten) && secondPart > (rem_u Rep.minus_one ten))
+        then raise (Failure "int_to_string");
+      Rep.add
```
rossberg-chromium: Nit: can we put this on one line? Also, you can shorten to `Rep.(add (mul ...) ...)`

rossberg-chromium:
(Sorry for the delay, just rediscovered this one). I think this is good to have, but the implementation is a bit more ad-hoc than I'd hope. Wouldn't it be simpler to just reimplement the normal conversion from scratch?

pjuftring:
I think that would be a good alternative, this code may work, but its probably not a nice solution.

PR #268, opened 2016-03-24
Binary encoder & decoder

rossberg-chromium:
Add working (post-order) encoder & decoder. Supersedes PR #261.

Extends the test runner with encoding & decoding of all test files via a wast->wasm transcoding and loading the resulting modules. Works for all current tests.

There are still a number of smaller TODOs left, but I think it makes more sense to address them in follow-ups. Other follow-up work: wasm->wast reverse transcoding, more aggressive tests and testing capabilities.

We'll probably also need to revise the use of different integer types throughout the spec, which is not always consistent with what the binary format supports.

### DIFF
--- a/ml-proto/host/encode.ml
+++ b/ml-proto/host/encode.ml
@@ -156,7 +156,7 @@ let encode m =
       | F64_store (o, a, e1, e2) -> binary e1 e2 0x36; memop o a
 
       | Grow_memory e -> unary e 0x39
-      | Memory_size -> op 0x3b
+      | Current_memory -> op 0x3b
 
       | I32_add (e1, e2) -> binary e1 e2 0x40
       | I32_sub (e1, e2) -> binary e1 e2 0x41
--- a/ml-proto/host/encode.ml
+++ b/ml-proto/host/encode.ml
@@ -47,8 +47,8 @@ let rec vs64 s i =
 let vu32 s i = vu64 s (Int64.of_int32 i)
 let vs32 s i = vs64 s (Int64.of_int32 i)
 let vu s i = vu64 s (Int64.of_int i)
-let f32 s x = vu32 s (F32.to_bits x)
-let f64 s x = vu64 s (F64.to_bits x)
+let f32 s x = u32 s (F32.to_bits x)
+let f64 s x = u64 s (F64.to_bits x)
 
 let bool s b = u8 s (if b then 1 else 0)
 let string s bs = vu s (String.length bs); put_string s bs
--- a/ml-proto/host/encode.ml
+++ b/ml-proto/host/encode.ml
@@ -304,7 +304,7 @@ let encode m =
 
     (* Type section *)
     let type_section ts =
-      section "signatures" (vec func_type) ts (ts <> [])
+      section "type" (vec func_type) ts (ts <> [])
 
     (* Import section *)
     let import imp =
@@ -312,17 +312,17 @@ let encode m =
       var itype; string module_name; string func_name
 
     let import_section imps =
-      section "import_table" (vec import) imps (imps <> [])
+      section "import" (vec import) imps (imps <> [])
 
     (* Function section *)
     let func f = var f.it.ftype
 
     let func_section fs =
-      section "function_signatures" (vec func) fs (fs <> [])
+      section "function" (vec func) fs (fs <> [])
 
     (* Table section *)
     let table_section tab =
-      section "function_table" (vec var) tab (tab <> [])
+      section "table" (vec var) tab (tab <> [])
 
     (* Memory section *)
     let memory mem =
@@ -343,11 +343,11 @@ let encode m =
     let export_section exps =
       (*TODO: pending resolution*)
       let exps = List.filter (fun exp -> exp.it.kind <> `Memory) exps in
-      section "export_table" (vec export) exps (exps <> [])
+      section "export" (vec export) exps (exps <> [])
 
     (* Start section *)
     let start_section xo =
-      section "start_function" (opt var) xo (xo <> None)
+      section "start" (opt var) xo (xo <> None)
 
     (* Code section *)
     let compress locals =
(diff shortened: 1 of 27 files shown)


JSStats:
Could you squash it into one patch to make it easier to view.

rossberg-chromium:
@JSStats, the github diff view already squashes it, doesn't it?

JSStats:
@rossberg-chromium Yes, thanks. Can checkout the 'decode' branch too.

ml-proto/host/encode.ml
```diff
@@ -0,0 +1,401 @@
...
+    let rec expr e =
+      match e.it with
+      | Nop -> op 0x00
+      | Block es -> op 0x01; list expr es; op 0x17
+      | Loop es -> op 0x02; list expr es; op 0x17
+      | If (e, es1, es2) ->
+        expr e; op 0x03; list expr es1;
+        if es2 <> [] then op 0x04; list expr es2; op 0x17
+      | Select (e1, e2, e3) -> expr e1; expr e2; expr e3; op 0x05
+      | Br (x, eo) -> opt expr eo; op 0x06; arity1 eo; var x
```
JSStats: The expression is not optional in the binary encoding, one expression is expected which may be a `nop`. The arity is a constant 1 and not necessary but include as a concession.

ml-proto/host/encode.ml
```diff
@@ -0,0 +1,401 @@
...
+      | Br_if (x, eo, e) -> opt expr eo; expr e; op 0x07; arity1 eo; var x
```
JSStats: Same here.

ml-proto/host/encode.ml
```diff
@@ -0,0 +1,401 @@
...
+      | Br_table (xs, x, eo, e) ->
+        opt expr eo; expr e; op 0x08; arity1 eo; vec var32 xs; var32 x
+
```
JSStats: The binary encoding does not even support and expression here, this is not the PR to add one. There is a long standing matter here that has not been resolved and is really blocking this issue. I don't want to see all these operators having implicit tuple constructors, it is not necessary, and don't want to see separate operators for the passing of the values of a single expression.

ml-proto/host/encode.ml
rossberg-chromium: See WebAssembly/design#595, which this already implements.

ml-proto/host/encode.ml
rossberg-chromium: We agreed to add this, and the spec AST already has it. (Keep in mind that this PR is meant to implement the binary format with all changes decided last month. Some of them haven't made it into the design repo or other implementations yet, but we're working on it.)

ml-proto/host/encode.ml
```diff
@@ -0,0 +1,401 @@
...
+      let {Memory.addr; data} = seg.it in
+      vu64 addr; string data
+
+    let data_section segs =
+      section "data" (opt (vec segment))
+        segs (segs <> None && segs <> Some [])
+
+    (* End section *)
+    let end_section () =
+      section "end" ignore () true
```
lukewagner: Currently there is no end section in BinaryEncoding.md.  If we want an explicit terminator, perhaps instead we can propose a `section_count : varuint32` that goes right after the version u32; but a dummy section seems like an abuse of what a section is.

ml-proto/spec/kernel.ml
```diff
@@ -119,16 +119,18 @@ and func' =
 type memory = memory' Source.phrase
 and memory' =
 {
-  initial : Memory.size;
+  min : Memory.size;
```
lukewagner: I think `initial` was the better name for this, since `min` doesn't convey that this is also the initial size of memory.  E.g., in the future, maybe we add a `shrink_memory` operator and then requiring declaring a separate `min` size that was <= `initial`.

ml-proto/spec/kernel.ml
```diff
@@ -119,16 +119,18 @@ and func' =
...
 }
 and segment = Memory.segment Source.phrase
 
 type export = export' Source.phrase
 and export' =
-  | ExportFunc of string * var
-  | ExportMemory of string
+{
+  name : string;
+  kind : [`Func of var | `Memory]
```
lukewagner: Ah, thanks goodness, much nicer.  I had wanted to name these as fields in [design/#659](https://github.com/WebAssembly/design/pull/659).

ml-proto/spec/decode.ml
```diff
@@ -0,0 +1,602 @@
...
+
+let rec iterate f s = if f s then iterate f s
+
+let module_ s =
+  let magic = u32 s in
+  require (magic = 0x6d736100l) s 0 "magic header not detected";
+  let version = u32 s in
+  require (version = 0x0bl) s 4 "unknown binary version";
+  iterate unknown_section s;
+  let types = type_section s in
```
lukewagner: See, it wouldn't be so terrible to pass in `types` as another arg to `code_section` ;)

lukewagner:
I haven't read it all line-by-line, but fwiw, I think it's fine to land this and iterate towards precisely matching BinaryEncoding.md at either 0xb or 0xc.  It's nice to have an alternate implementation that first builds the AST and only afterwards performs type validation.  Ultimately, we'll shake out the discrepancies when we have ml-proto encode .wasms that SM/V8 consume and vice versa with sexpr-wasm-proto.

ml-proto/host/encode.ml
rossberg-chromium: Ah right, forgot to remove that. Done.

ml-proto/spec/kernel.ml
rossberg-chromium: I see, good point. Well, I'll change the name to whatever AstSemantics and BinaryEncoding will eventually agree on. :)

rossberg-chromium:
Okay, thanks. I'm taking that as an LGTM and go merge it.

ml-proto/spec/kernel.ml
lukewagner: Ah, I hadn't noticed that BinaryEncoding.md calls it 'min' too; filed a [PR](https://github.com/WebAssembly/design/pull/661) to update BinaryEncoding.md.  ([Modules.md](https://github.com/WebAssembly/design/blob/master/Modules.md#linear-memory-section) already calls it initial.)

PR #271, opened 2016-04-05
Allow subtyping on optional operands

rossberg-chromium:
This reverses #180, and adds a number of corresponding tests.

Motivation: not allowing optional operands to be implicitly dropped (1) reduced the compositionality of the language, (2) made it harder to allow multiple targets (like with a br_if also returning its operand), and (3) put an extra burden on bottom-up type checkers, where the arity might not be known upfront. And with multiple value, we eventually would want to lift this restriction again anyway.

This should address issues #227, and enable WebAssembly/design#539.

@jcbeyler, you filed #179, which lead to #180. Are you okay with reverting to status ante quo?

### DIFF
--- a/ml-proto/spec/check.ml
+++ b/ml-proto/spec/check.ml
@@ -230,10 +230,9 @@ and check_exprs c ts es at =
   with Invalid_argument _ -> error at "arity mismatch"
 
 and check_expr_opt c et eo at =
-  match et, eo with
-  | Some t, Some e -> check_expr c et e
-  | None, None -> ()
-  | _ -> error at "arity mismatch"
+  match eo with
+  | Some e -> check_expr c et e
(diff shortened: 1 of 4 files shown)


lukewagner:
With this change, is the obvious bottom-up algo described [here](https://github.com/WebAssembly/spec/pull/215#issuecomment-202624909) valid?  If not, I'd be interested to know what are the cases which differ.

rossberg-chromium:
@lukewagner, yes, that's the idea: you simply do a join of all branches targeting the block, wrt to the trivial subtyping semi-lattice that has Void as its top type.

JSStats:
Thank you.

lukewagner:
Great, lgtm

rossberg-chromium:
Landing with LGTM and no objections

PR #273, opened 2016-04-07
Make grow_memory and return pages, and replace memory_size with current_memory

rossberg-chromium:
To match WebAssembly/design#629 and WebAssembly/design#648.


lukewagner:
lgtm, yes, we should update/add `memory_size` too

rossberg-chromium:
Okay, I added `memory_size` to this PR, PTAL.

lukewagner:
lgtm.  One question I just remembered, and the reason I didn't simply throw in the `memory_size` update with [design/#629](https://github.com/WebAssembly/design/pull/629), is that "size" doesn't sound quite right if the returned unit is pages.  That could be a separate issue, though.

rossberg-chromium:
Agreed. `current_memory` perhaps?

lukewagner:
Yes, that has a nice symmetry with `grow_memory`.

rossberg-chromium:
Okay, updated with a rename memory_size -> current_memory. Corresponding design PR is WebAssembly/design#648. PTAL

lukewagner:
Great, lgtm (assuming no objections to design PR).

rossberg-chromium:
Landing with LGTM and no objections

PR #274, opened 2016-04-18
More tests!

sunfishcode:
This adds more floating point rounding tests, more expression optimizer
bait tests, more real-world floating point bug tests, and more comments.


rossberg-chromium:
LGTM, though I haven't checked in detail.

sunfishcode:
Merging, with a qualified LGTM, as this is just adding tests and not changing any spec code. And of course, these and all tests continue to be subject to review at any time by anyone interested.

PR #278, opened 2016-04-25
Implement conversion to S-expression

rossberg-chromium:
The interpreter can now arbitrarily convert (and round-trip) between binary and text format, e.g.:
```
wasm file.wasm -o file.wast
```
disassembles a binary into a (pretty-printed) S-expression file, and
```
wasm file.wast -o file2.wasm
```
reencodes it. Similarly, the new `input` and `output` script commands can be used.

The test runner has been extended to round-trip between formats, and rerun every test with each. However, since conversion loses script commands like assertions and invocations, that currently mostly checks that the module still validates -- we should move tests to use a `start` function to get around that.


kripken:
> However, since conversion loses script commands like assertions and invocations, that currently mostly checks that the module still validates -- we should move tests to use a start function to get around that.

Another option is to parse the wast files and split them up. The binaryen check.py script [splits wasts into modules + asserts that follow them](https://github.com/WebAssembly/binaryen/blob/master/check.py#L164), then does binary and back on the modules, then combines that output with the asserts, and runs that.

lukewagner:
I still like the idea of a .wasmt file which is the same as a .wast except with the contents of `(module)` being an ascii-encoded binary module.  That would allow simple 1:1 mapping that doesn't spit out multiple files.

rossberg-chromium:
@kripken, yes, but that would require more infrastructure. Not sure the spec would be the best place for that.

@lukewagner, it could be useful, and I'm thinking about implementing something along these lines. I think the current feature is still very useful nevertheless.

sunfishcode:
@rossberg-chromium I agree that this is a useful step. However, imagine a programming language with an implicit conversion from "type list" to "type" which worked by silently discarding all but one element of the list; that'd be surprising, and that's basically what ".wasmt" fixes :-).

ml-proto/host/format.mli
```diff
@@ -0,0 +1,7 @@
+open Sexpr
+
+val func_type : Types.func_type -> sexpr
+
+val expr : Kernel.expr -> sexpr
+val module_ : Kernel.module_ -> sexpr
```
lukewagner: I wonder if `Format` is the right module name here; "Format" sounds like it generates formatted text output.

lukewagner:
Yes, it makes sense to merge this now, I was just talking about what a better end state could be wrt testing binary.

ml-proto/host/format.mli
rossberg-chromium: Yeah, I was struggling with a good name. I chose `Format` as the inverse to `Parse`, but later factored the actual printing to a separate module. Suggestions welcome.

lukewagner: I'd been thinking it'd make sense just to put it in `Sexpr`, renaming the functions `of_module`/`of_expr`.

rossberg-chromium:
@sunfishcode, well, the command line flag is just a shorthand for the actual `output` command, which applies to the "current" module just like everything else. You can in fact have several, like
```
wasm file1.wast -o file1.wasm file2.wasm -o file2.wast
```
This is really just a shorthand for executing the script
```
(input "file1.wast") (output "file1.wasm")
(input "file2.wasm") (output "file2.wast")
```

So, everything is in terms of script commands, which operate on modules (except `input`, which is a bit more general on `wast` files). There is no meta command layer that can manipulate scripts themselves. I'd be rather hesitant to go there, since the script layer should really only be a smallish convenience to ease tests. I already started regretting that I introduced it. :)

rossberg-chromium:
Okay, merging this one.

PR #282, opened 2016-04-28
More floating point tests!

sunfishcode:
This adds a bunch more tests focused on floating point.

Note that there are some minor inconsistencies in general with how NaNs are handled; I plan to address those separately as part of the work related to https://github.com/WebAssembly/design/pull/660.


sunfishcode:
Ping :-).

kripken:
I skimmed the tests and things look good to me, although, most goes above my head ;)

However, the binaryen interpreter fails on

````
(assert_return (invoke "f32.no_fold_div_0" (f32.const nan:0x200000)) (f32.const nan:0x600000))
````

I'm not sure I understand what is being tested there?

**Comment by jfbastien** [comment]
> I skimmed the tests and things look good to me, although, most goes above my head ;)
> 
> However, the binaryen interpreter fails in float_exprs.wast on
>
> `(assert_return (invoke "f32.no_fold_div_0" (f32.const nan:0x200000)) (f32.const nan:0x600000))`
>
> I'm not sure I understand what is being tested there?

Looks like the bug is this last line:
```
  Literal div(const Literal& other) const {
    switch (type) {
      case WasmType::f32: {
        float lhs = getf32(), rhs = other.getf32();
        float sign = std::signbit(lhs) == std::signbit(rhs) ? 0.f : -0.f;
        switch (std::fpclassify(rhs)) {
          case FP_ZERO:
          switch (std::fpclassify(lhs)) {
            case FP_NAN: return *this;
```
AstSemantics says:
> When the result of any arithmetic operation other than neg, abs, or copysign is a NaN, the sign bit and the fraction field (which does not include the implicit leading digit of the significand) of the NaN are computed as follows:
> *If the operation has any NaN input values, implementations may select any of them to be the result value, but with the most significant bit of the fraction field overwritten to be 1.
> * If the implementation does not choose to use an input NaN as a result value, or if there are no input NaNs, the result value has a nondeterministic sign bit, a fraction field with 1 in the most significant bit and 0 in the remaining bits.

So the requirement is to set 1 as the MSB of the fraction field, rest can be zero or the original value, and sign bit can be anything. The test seems over-constrained, but I also have a bug in binaryen. I'll fix :-)

**Comment by sunfishcode** [comment]
It's testing that a floating-point divide of a signaling NaN operand divided by an immediate zero returns a quiet NaN, per IEEE 754, and that the rest of the NaN payload and sign bit are propagated per wasm's rules, except that it's not yet updated for https://github.com/WebAssembly/design/pull/660. I plan to update it when I update all the tests.

File: ml-proto/test/float_exprs.wast [hunk 1]
```diff
@@ -1219,6 +1271,19 @@
 
 (assert_return (invoke "llvm_pr27153" (i32.const 33554434)) (f32.const 25165824.000000))
 
+;; Test that (float)x + (float)y is not optimzied to (float)(x + y) when unsafe.
```
binji: optimized

ml-proto/test/float_exprs.wast
```diff
@@ -1219,6 +1271,19 @@
...
+;; https://llvm.org/bugs/show_bug.cgi?id=27036
+
+(module
+  (func $llvm_pr27036 (param $x i32) (param $y i32) (result f32)
+    (f32.add (f32.convert_s/i32 (i32.or (get_local $x) (i32.const -25034805)))
+             (f32.convert_s/i32 (i32.and (get_local $y) (i32.const 14942208))))
+  )
+  (export "llvm_pr27936" $llvm_pr27036)
```
binji: 27036

ml-proto/test/float_exprs.wast
sunfishcode: Fixed; thanks!

ml-proto/test/float_exprs.wast
sunfishcode: Good spot! Fixed.

jfbastien:
`NaN / 0` fixed in https://github.com/WebAssembly/binaryen/pull/442. It still seems like the test is over-constrained? For now all the implementations do "the right thing" so maybe just comment that it's over-constrained?

sunfishcode:
Can you clarify in what way it's over-constrained?

jfbastien:
> Can you clarify in what way it's over-constrained?

I'm not sure if I'm misunderstanding, but this seems to give leeway to implementations:

* If the operation has any NaN input values, implementations may select any of them to be the result value, but with the most significant bit of the fraction field overwritten to be 1.
* If the implementation does not choose to use an input NaN as a result value, or if there are no input NaNs, the result value has a nondeterministic sign bit, a fraction field with 1 in the most significant bit and 0 in the remaining bits.

It sounds like the sign can be whatever, and the NaN can propagate to quiet (same payload, but not signaling) or be quiet NaN with zeroes as fraction. Is that correct?

sunfishcode:
That's the new text in WebAssembly/design#660. Many existing tests also need to be updated as well. I plan to do all the updates at once.

jfbastien:
> That's the new text in WebAssembly/design#660. Many existing tests also need to be updated as well. I plan to do all the updates at once.

Ah right, ty! /me unconfused

jfbastien:
High-level lgtm!

sunfishcode:
Merging, with high-level lgtm, since these just add new tests with no spec changes. As always, these and all tests continue to be open to comments and questions.

rossberg-chromium:
Sorry, I was away. Also LGTM

PR #283, opened 2016-04-30
Write integer of_string from scratch

### LINKED ISSUE #225: ml-proto does not accept the full range of unsigned i32 or i64 constants

pjuftring:
While #265 did the job of supporting unsigned integers, it certainly was
no nice implementation. This version adds the same features, i.e.
allowing signed integers (as defined in the lexer) and larger unsigned
values, but is actually written from scratch and doesn't rely on the
builtin of_string anymore. This PR should close #225 if it is accepted.


rossberg-chromium:
Hm, this code seems to cause assertion failures on invalid input. That should never happen. Can you change it to raise normal exceptions, preferably the same as the original?

Additional nit: there are plenty of redundant parens in the code.

pjuftring:
I removed all redundent parentheses, that I could find, and added a "try" to catch my asserts.

rossberg-chromium:
Please do not use assertions like this. Assertions are intended for checking invariants (pre/post-conditions) and thus the internal consistency of code, _not_ for checking the validity of input data. An assertion failure is supposed to signal a program bug.

Also, please do never write catch-all try handlers. That is a guarantee for unintentionally masking unrelated errors. Only match the exceptions you expect to get and intend to handle.

pjuftring:
Oh, sorry, I didn't know that. I hope, that I solved it correctly.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
   let ge_s x y = x >= y
   let ge_u x y = cmp_u x (>=) y
 
-  let of_string = Rep.of_string
+  (* This implementation allows leading signs and unsigned values *)
+  let of_string x =
+    let len = String.length x in
```
rossberg-chromium: Hint: `let open Rep in` allows to omit the repeated module prefix.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+    let power_of_two n = Rep.shift_left Rep.one n in
+    let ten = Rep.of_int 10 in
+    let fail () = failwith "int_of_string" in
```
rossberg-chromium: You could make this more compact by turning it into a `require` function, similar e.g. to that used in `check.ml`.

The string should probably be `"of_string"`.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+    let parse_hexdigit c =
+      int_of_char c -
+        if '0' <= c && '9' >= c then 0x30
+        else if 'a' <= c && 'f' >= c then 0x61 - 10
+        else if 'A' <= c && 'F' >= c then 0x41 - 10
+        else fail ()
+    in
+    let parse_hex offset sign =
+      let num = ref Rep.zero in
```
rossberg-chromium: The code would probably become shorter (and more idiomatic) if you used recursion instead of a loop.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+      for i = offset to len - 1 do
+        if ge_u !num (power_of_two (Rep.bitwidth - 4)) then fail ();
```
rossberg-chromium: I think you can simplify this to `gt_u !num (shr_u minus_zero 4)`

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+        num := Rep.logor (Rep.shift_left !num 4) (Rep.of_int (parse_hexdigit x.[i]));
+      done;
+      if sign && gt_u !num (power_of_two (Rep.bitwidth - 1)) then fail ();
```
rossberg-chromium: Why not simply `sign && lt_s !num zero`?

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+      !num
+    in
+    let parse_dec offset sign =
+      let max_upper, max_lower =
```
rossberg-chromium: These and a couple of other locals are constants, so it's better to move them out of the entire function. The same goes for a couple of helper functions that do not depend on `x`.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+        if sign then
```
rossberg-chromium: You can push the `if` into the first operand.

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+          divrem_u (power_of_two (Rep.bitwidth - 1)) ten
```
rossberg-chromium: Shouldn't this be `divrem_u (shr_u minus_one 1) ten` instead?

ml-proto/spec/int.ml
```diff
@@ -208,7 +208,55 @@ struct
...
+      for i = offset to len - 1 do
+        if '0' > x.[i] || '9' < x.[i] then fail ();
+        let new_digit = Rep.of_int (int_of_char x.[i] - 0x30) in
+        if gt_u !num max_upper || (!num = max_upper && gt_u new_digit max_lower) then fail ();
+        num := Rep.add (Rep.mul !num ten) new_digit
+      done;
+      !num
+    in
+    let parse_int offset sign =
+      if offset + 3 <= len && String.sub x offset 2 = "0x" then
```
rossberg-chromium: Nit: it's cheaper to just compare `x.[offset]` and x.[offset+1]` than allocating a substring.

rossberg-chromium:
Thanks! I left a few comments. Also, it would be good to have a few additional tests, especially for boundary conditions.

ml-proto/spec/int.ml
pjuftring: Don't I have to worry about the stack, if I use recursion here?

rossberg-chromium: Well, it should be tail-recursive, of course. Like so:
```
let rec parse_hex i num =
  if i = len then num
  else if overflow... then fail ()
  else parse_hex (i+1) (logor (shift_left num 4) (of_int (parse_hexdigit x.[i])))
```
Btw, I think you can get rid of the entire sign-specific logic in these functions. For values with negative sign it should suffice to simply check once in the end (at line 257 below) whether the value is negative already.

pjuftring:
I addressed the comments, but there's some stuff I'm not sure about:
* Did I move the right helper functions out of of_int?
* Do I raise the correct exceptions in my helper functions?
* How can I test the bounds of my implementation sufficently, when I can't catch lexer-errors within the test-file?

ml-proto/spec/int.ml
```diff
@@ -208,7 +210,49 @@ struct
...
-  let of_string = Rep.of_string
+  let parse_hexdigit c =
+    int_of_char c -
+      if '0' <= c && '9' >= c then 0x30
+      else if 'a' <= c && 'f' >= c then 0x61 - 10
+      else if 'A' <= c && 'F' >= c then 0x41 - 10
+      else failwith "unexpected digit"
+
+  let neg_safe x =
+    if lt_s (sub x Rep.one) Rep.minus_one then raise Numerics.IntegerOverflow;
```
rossberg-chromium: Since this is called only once, I'd just inline it. And just use `failwith`, too.

ml-proto/spec/int.ml
```diff
@@ -208,7 +210,49 @@ struct
...
+    Rep.neg x
+
+  (* This implementation allows leading signs and unsigned values *)
+  let of_string x =
+    let open Rep in
+    let len = String.length x in
+    let require b = if not b then failwith "of_int" in
```
rossberg-chromium: Did you mean `"of_string"`?

ml-proto/spec/int.ml
```diff
@@ -208,7 +210,49 @@ struct
...
+      else begin
+        require (le_u num (shr_u minus_one (of_int 4)));
+        parse_hex (i + 1) (logor (shift_left num 4) (of_int (parse_hexdigit x.[i])))
+      end
+    in
+    let rec parse_dec i num =
+      if i = len then num
+      else begin
+        require ('0' <= x.[i] && '9' >= x.[i]);
+        let new_digit = of_int (int_of_char x.[i] - 0x30) in
```
rossberg-chromium: Just for symmetry with `parse_hexdigit`, perhaps move this and the previous line to a helper function, too.

rossberg-chromium:
Thanks, looks pretty good now, just a few nits left.

As for testing, you can write test files named `xyz.fail.wast`, which the runner expects to fail. If you also want to check the specific error output then you have to create associated expected-output files (though that's perhaps overkill in this case).

ml-proto/spec/int.ml
```diff
@@ -208,7 +210,49 @@ struct
   let ge_s x y = x >= y
   let ge_u x y = cmp_u x (>=) y
 
-  let of_string = Rep.of_string
+  let parse_hexdigit c =
+    int_of_char c -
+      if '0' <= c && '9' >= c then 0x30
+      else if 'a' <= c && 'f' >= c then 0x61 - 10
+      else if 'A' <= c && 'F' >= c then 0x41 - 10
+      else failwith "unexpected digit"
```
hhugo: fwiw, I find the following easier to read : 
```
let parse_hexdigit = function
  | '0' .. '9' as c ->  Char.code c - Char.code '0'
  | 'a' .. 'f' as c ->  0xa + Char.code c - Char.code 'a' 
  | 'A' .. 'F' as c ->  0xa + Char.code c - Char.code 'A'
  | _                  ->  failwith "unexpected digit"
```

ml-proto/spec/int.ml
pjuftring: But if I would move this function out, I would have to call it with the error message, like your implementation did in change.ml. I wanted to avoid this, shall I still change it?

ml-proto/spec/int.ml
pjuftring: I think, this is a lot prettier than my version @rossberg-chromium shall I copy this?

ml-proto/spec/int.ml
rossberg-chromium: Not sure I follow -- it's not using any local variables, so you should be able to just hoist it as is.

ml-proto/spec/int.ml
rossberg-chromium: Yes, looks good.

ml-proto/spec/int.ml
pjuftring: Everytime I call "require" I want it to be the same error message "of_string", but if I move it out, I can't hardcode this message, since this particular error message does only make sense in the scope of my "of_string"-function. I will have to supply the same error string everytime I call the hoisted "require"-function. This seemed a bit redundant to me. This is no big deal, I guess, I simply thought, that if I only have one error message, my solution would be a bit compact.

rossberg-chromium: It's still only a helper for to_string, so no need to generalise, you can leave the string fixed. (Being a helper to a specific function does not imply that it has to be local to that function. In general, ML code tends to use modules as the unit of encapsulation, and it's usually not worth being more fine-grained about it.)

pjuftring:
I added some tests and hoisted `parse_decdigit` and `require`.

ml-proto/spec/int.ml
```diff
@@ -250,7 +249,12 @@ struct
     in
     match x.[0] with
       | '+' -> parse_int 1
-      | '-' -> neg_safe (parse_int 1)
+      | '-' ->
+        begin
```
rossberg-chromium: Nit: you don't need the begin/end here

ml-proto/spec/int.ml
```diff
@@ -210,22 +210,22 @@ struct
...
-    int_of_char c -
-      if '0' <= c && '9' >= c then 0x30
-      else if 'a' <= c && 'f' >= c then 0x61 - 10
-      else if 'A' <= c && 'F' >= c then 0x41 - 10
-      else failwith "unexpected digit"
+  let parse_hexdigit = function
+    | '0' .. '9' as c ->  Char.code c - Char.code '0'
+    | 'a' .. 'f' as c ->  0xa + Char.code c - Char.code 'a'
+    | 'A' .. 'F' as c ->  0xa + Char.code c - Char.code 'A'
+    | _ ->  failwith "unexpected digit"
```
rossberg-chromium: This should be "of_string", too.

ml-proto/spec/int.ml
```diff
@@ -210,22 +210,22 @@ struct
...
 
-  let neg_safe x =
-    if lt_s (sub x Rep.one) Rep.minus_one then raise Numerics.IntegerOverflow;
-    Rep.neg x
+  let parse_decdigit c =
+    if '0' > c || '9' < c then failwith "unexpected digit";
```
rossberg-chromium: Same here.

rossberg-chromium:
Thanks! LGTM modulo a last couple of nits

PR #284, opened 2016-05-10
Add tests for if/then/else with indices + unknown label;

bnjbvr:
Hi! I'd like to add test cases we ran into in our implementation and which could be useful for others:

- a test for if/then/else where the depths are not labels, but raw integer indices.
- a test for if/then/else where the label target of a `br` in the `else` is the label of the `then` block.

Unfortunately, the second test (at the end of the file) doesn't work: the parser complains that the label `$l` is unknown, which is exactly what we are testing here. The correct behavior would be to silently swallow the error and consider the test as passed. Am I missing something here?

### DIFF
--- a/ml-proto/runtests.py
+++ b/ml-proto/runtests.py
@@ -15,9 +15,14 @@ def auxFile(path):
     return path
 
 class RunTests(unittest.TestCase):
-  def _runCommand(self, command):
-    exitCode = subprocess.call(command, shell=True)
-    self.assertEqual(0, exitCode, "test runner failed with exit code %i" % exitCode)
+  def _runCommand(self, command, logPath = None, expectedExitCode = 0):
+    out = None if logPath is None else file(logPath, 'w+')
+
+    exitCode = subprocess.call(command, shell=True, stdout=out, stderr=subprocess.STDOUT)
+    self.assertEqual(expectedExitCode, exitCode, "test runner failed with exit code %i, expected %i" % (exitCode, expectedExitCode))
+
+    if logPath is not None:
+      out.close()
 
   def _compareFile(self, expectedFile, actualFile):
     with open(expectedFile) as expected:
@@ -33,28 +38,34 @@ class RunTests(unittest.TestCase):
       pass
 
   def _runTestFile(self, shortName, fileName, interpreterPath):
+
+    expectedExitCode = 1 if ".fail." in fileName else 0
+
     # Run original file
     logPath = auxFile(fileName.replace("test/", "test/output/").replace(".wast", ".wast.log"))
-    self._runCommand(("%s %s > %s") % (interpreterPath, fileName, logPath))
+    self._runCommand(("%s %s") % (interpreterPath, fileName), logPath, expectedExitCode)
     self._compareLog(fileName, logPath)
+
+    if expectedExitCode != 0:
+      return
 
     # Convert to binary and run again
     wasmPath = auxFile(fileName.replace("test/", "test/output/").replace(".wast", ".wast.wasm"))
     logPath = auxFile(fileName.replace("test/", "test/output/").replace(".wast", ".wast.wasm.log"))
     self._runCommand(("%s -d %s -o %s") % (interpreterPath, fileName, wasmPath))
-    self._runCommand(("%s %s > %s") % (interpreterPath, wasmPath, logPath))
+    self._runCommand(("%s %s") % (interpreterPath, wasmPath), logPath)
 
     # Convert back to text and run again
     wastPath = auxFile(fileName.replace("test/", "test/output/").replace(".wast", ".wast.wasm.wast"))
     logPath = auxFile(fileName.replace("test/", "test/output/").replace(".wast", ".wast.wasm.wast.log"))
     self._runCommand(("%s -d %s -o %s") % (interpreterPath, wasmPath, wastPath))
-    self._runCommand(("%s %s > %s") % (interpreterPath, wastPath, logPath))
+    self._runCommand(("%s %s ") % (interpreterPath, wastPath), logPath)
 
     #return
     # Convert back to binary once more and compare
     wasm2Path = auxFile(fileName.replace("test/", "test/output/").replace(".wast", ".wast.wasm.wast.wasm"))
     self._runCommand(("%s -d %s -o %s") % (interpreterPath, wastPath, wasm2Path))
-    self._runCommand(("%s %s > %s") % (interpreterPath, wasm2Path, logPath))
+    self._runCommand(("%s %s") % (interpreterPath, wasm2Path), logPath)
     # TODO: Ultimately, the binary should stay the same, but currently desugaring gets in the way.
     # self._compareFile(wasmPath, wasm2Path)
 
--- a/ml-proto/test/expected-output/if_label_scope.fail.wast.log
+++ b/ml-proto/test/expected-output/if_label_scope.fail.wast.log
@@ -1,0 +1,1 @@
+test/if_label_scope.fail.wast:6.19-6.21: unknown label $l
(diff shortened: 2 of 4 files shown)


rossberg-chromium:
The current behaviour is correct: the body of each if-branch is (implicitly) a block, and the labels name these respective blocks. Not sure why you expect a label to be in scope in a sibling block? It's no different from
```
(if (...) (then (block $l ...)) (else (block (br $l))))
```
which likewise is an error.

The other test look good.

kripken:
lgtm, I think the test is valid too.

kripken:
@rossberg-chromium I think it's not meant to be in scope, it should be an error - it's an assert_invalid test.

rossberg-chromium:
Oh, sorry, totally missed that. Yeah, that's definitely a bug, I'll look into it.

rossberg-chromium:
Okay, I now see what's going on. The problem is that labels (and identifiers in general) are resolved in the parser, because the AST doesn't have them. An unbound name hence is a syntax error, rejecting the script _before_ you ever get to run it.

So it's not really a bug but a limitation of scripts -- a script can't check for this kind of error, just like it can't check other syntax errors. You'll need to check externally against expected-output instead.

binji:
@rossberg-chromium I actually ran an issue in sexpr-wasm because of this; prior to having binary modules, I could assume that `assert_invalid` would only be given modules that were syntactically valid. So I moved the checking pass out of the parser, and the parser always returns a module as an AST, even inside `assert_invalid`. But now that binary modules can fail to parse, I need a second path for passing back binary data that is only later parsed and checked for errors.

bnjbvr:
> So it's not really a bug but a limitation of scripts -- a script can't check for this kind of error, just like it can't check other syntax errors. You'll need to check externally against expected-output instead.
Thank you for the explanation, I can do this, however I wonder how https://github.com/WebAssembly/spec/blob/master/ml-proto/test/switch.wast#L155 doesn't trigger the exact same issue?

rossberg-chromium:
@bnjbvr, that test uses a raw numeric index. What's resolved in the parser is the use of symbolic labels, and their translation into numeric ones (because they are purely syntactic sugar, not a part of the Wasm AST).

I certainly won't argue that this difference is nice. :)

bnjbvr:
I've updated the PR. The code wasn't accepting non-zero exit codes, so I had to refactor this, by adding a way to specify that a test is expected to fail. Also, there were issues with redirecting stderr to stdout with a `>` in the command argument, so I've refactored the way output is written so that stdout and stderr goes to the output file, instead. How does this look?

ml-proto/runtests.py
```diff
@@ -32,29 +37,48 @@ def _compareLog(self, fileName, logPath):
...
 
+  def _readHeaders(self, fileName):
+    expected_exit_code = 0
+
+    with file(fileName) as f:
+      headers = f.readline().split(' ')
+
+      if headers[0] == ';;;':
+        for h in headers[1:]:
+          if h.strip() == 'expected-fail':
```
rossberg-chromium: Instead of instrumenting comments in the file, what do you think about simply using a file name convention? Say, `<test>.fail.wast` for tests that ought to fail? That's easier and cheaper to process (including for other test harnesses).

ml-proto/runtests.py
```diff
@@ -32,29 +37,48 @@ def _compareLog(self, fileName, logPath):
     except IOError:
       pass
 
+  def _readHeaders(self, fileName):
+    expected_exit_code = 0
```
rossberg-chromium: Nit: Python prefers camlCasing.

ml-proto/runtests.py
```diff
@@ -32,29 +37,48 @@ def _compareLog(self, fileName, logPath):
...
+
+    with file(fileName) as f:
+      headers = f.readline().split(' ')
+
+      if headers[0] == ';;;':
+        for h in headers[1:]:
+          if h.strip() == 'expected-fail':
+              expected_exit_code = 1
+
+      return (expected_exit_code)
```
rossberg-chromium: Nit: any reason for the parens?

ml-proto/runtests.py
```diff
@@ -32,29 +37,48 @@ def _compareLog(self, fileName, logPath):
...
+
   def _runTestFile(self, shortName, fileName, interpreterPath):
+
+    (expected_exit_code) = self._readHeaders(fileName)
```
rossberg-chromium: Nit: same here

ml-proto/test/expected-output/label_not_found.wast.log
```diff
@@ -0,0 +1 @@
+test/label_not_found.wast:7.19-7.21: unknown label $l
```
rossberg-chromium: Perhaps use a more specific name for this test, e.g. `if-label-scope`

rossberg-chromium:
Thanks! Looks good modulo a few comments.

ml-proto/runtests.py
bnjbvr: Yeah, good idea.

ml-proto/runtests.py
bnjbvr: I have always been told that python was strongly in favor of snake_casing? Staying consistent with the original style, though.

bnjbvr:
Updated the latest patch, thank you for the review!

ml-proto/runtests.py
```diff
@@ -33,28 +38,34 @@ def _compareLog(self, fileName, logPath):
       pass
 
   def _runTestFile(self, shortName, fileName, interpreterPath):
+
+    expectedExitCode = 1 if "fail" in fileName else 0
```
rossberg-chromium: Can we make this a bit narrower? ".fail." should be good.

ml-proto/runtests.py
rossberg-chromium: Maybe. To be honest, I'm no Python expert. :)

rossberg-chromium:
Hm, Travis still barfs at the .fail test, although I can't immediately say why.

ml-proto/runtests.py
```diff
@@ -15,9 +15,14 @@ def auxFile(path):
...
-    exitCode = subprocess.call(command, shell=True)
-    self.assertEqual(0, exitCode, "test runner failed with exit code %i" % exitCode)
+  def _runCommand(self, command, logPath = None, expectedExitCode = 0):
+    out = None if logPath is None else file(logPath, 'w+')
+
+    exitCode = subprocess.call(command, shell=True, stdout=out, stderr=subprocess.STDOUT)
+    self.assertEqual(expectedExitCode, exitCode, "test runner failed with exit code %i, expected %i" % (exitCode, expectedExitCode))
+
+    if logPath is not None:
+        out.close()
```
rossberg-chromium: One further style nit I just noted: inconsistent indentation, rest of the file uses 2

ml-proto/runtests.py
```diff
@@ -33,28 +38,34 @@ def _compareLog(self, fileName, logPath):
...
+    expectedExitCode = 1 if ".fail." in fileName else 0
+
     # Run original file
     logPath = auxFile(fileName.replace("test/", "test/output/").replace(".wast", ".wast.log"))
-    self._runCommand(("%s %s > %s") % (interpreterPath, fileName, logPath))
+    self._runCommand(("%s %s") % (interpreterPath, fileName), logPath, expectedExitCode)
     self._compareLog(fileName, logPath)
 
+    if expectedExitCode != 0:
+        return
```
rossberg-chromium: same here

bnjbvr:
Right, the test was failing because we are now very tied to line numbers, which is unfortunate. Addressed the two styles nits and updated the expected output to match line numbers.

rossberg-chromium:
Cool, thanks! LGTM and landing.

PR #287, opened 2016-05-13
Batch of additional tests

rossberg-chromium:
For starters, more tests for:
* block
* unreachable
* nop
* comments

Also introduced a bit of sugar for exports. You can now write an export name as part of the function def itself:
```
(func "func-name" $func-name ...)
```
or, if the symbolic name isn't needed, just
```
(func "func-name" ...)
```
avoiding the tedious triple repetition of the name like before.


ml-proto/test/block.wast
```diff
@@ -1,35 +1,165 @@
...
+    (block (br 0 (i32.const 8)) (br 0) (br 0 (f64.const 8)))
+    (block (br 0 (i32.const 8)) (br 0 (call $nop)) (br 0 (f64.const 8)))
+    (block (br 0 (i32.const 8)) (br 0) (br 0 (f32.const 8)) (i64.const 3))
+    (block (br 0) (br 0 (i32.const 8)) (br 0 (f64.const 8)) (br 0 (nop)))
+    (block (br 0) (br 0 (i32.const 8)) (br 0 (f32.const 8)) (i64.const 3))
+    (block (block (br 0) (br 1 (i32.const 8))) (br 0 (f32.const 8)) (i64.const 3))
   )
 
-  (func $effects (result i32)
+  (func "effects" $fx (result i32)
```
kripken: why should this be valid? i don't understand what two names for a function means

binji: This is syntactic sugar for:
```
(export "effects" $fx)
(func $fx ...)
```

kripken: Oh, thanks! I missed this part of the issue text...

ml-proto/test/comments.wast
```diff
@@ -0,0 +1,69 @@
...
+
+(;)))))))))));)
+
+(;comment";)
+
+(;comment"";)
+
+(;comment""";)
+
+(;äh,scheiße;)
```
sunfishcode: Would you mind rewording this to avoid the possibility of appearing unprofessional?

pjuftring: How about "Heiße Würstchen"? This has umlaut and sharp s as well :)

rossberg-chromium: Done.

sunfishcode:
lgtm

PR #288, opened 2016-05-13
Misc tests

sunfishcode:
A few more floating point tests, but also includes a test for memory size overflow.

### DIFF
--- a/ml-proto/test/memory.wast
+++ b/ml-proto/test/memory.wast
@@ -2,6 +2,7 @@
 (module (memory 0 0))
 (module (memory 0 1))
 (module (memory 1 256))
+(module (memory 0 65535))
 (module (memory 0 0 (segment 0 "")))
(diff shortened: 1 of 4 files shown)


ml-proto/test/memory.wast
```diff
@@ -31,6 +32,10 @@
   (module (memory 1 2 (segment 0 "a") (segment 2 "b") (segment 1 "c")))
   "data segment not disjoint and ordered"
 )
+(assert_invalid
+  (module (memory 0 65536))
```
rossberg-chromium: How about also adding a test with a value that overflows the 32-bit (resp 64-bit range), to check that there is no accidental wrap-around?

rossberg-chromium:
lgtm

ml-proto/test/memory.wast
sunfishcode: Ok; tests added.

PR #289, opened 2016-05-13
Add some typechecking tests

binji:


binji:
I don't understand this error:
```
.......................................string:1.1-1.43: no module defined
F.
```

The tests pass locally for me; am I holding it wrong? :)

rossberg-chromium:
That's because the test runner tries to convert every test to binary, and that fails if there is no module defined. Adding a dummy `(module)` should fix it.

rossberg-chromium:
Other than that, lgtm.

PR #292, opened 2016-05-20
Next batch of tests

rossberg-chromium:
This time:
- loop
- br
- return
- bunch of additions to previous tests

Plus a bug fix in the pretty printer.

### DIFF
--- a/ml-proto/test/block.wast
+++ b/ml-proto/test/block.wast
@@ -46,43 +46,62 @@
     (f32.gt (block (call $nop) (f32.const 3)) (block (call $nop) (f32.const 3)))
   )
 
-  (func "br-bare" (result i32)
+  (func "break-bare" (result i32)
     (block (br 0) (unreachable))
+    (block (br_if 0 (i32.const 1)) (unreachable))
+    (block (br_table 0 (i32.const 0)) (unreachable))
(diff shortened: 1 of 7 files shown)


ml-proto/test/block.wast
```diff
@@ -46,43 +46,62 @@
     (f32.gt (block (call $nop) (f32.const 3)) (block (call $nop) (f32.const 3)))
   )
 
-  (func "br-bare" (result i32)
+  (func "break-bare" (result i32)
```
sunfishcode: Please do not rename "br" to "break" in the tests; the instruction is named "br".

rossberg-chromium: This is to match the loop test cases, where the names distinguish between breaking and continuing (see below). So the naming does not refer to the operator, but to the control flow transfer happening.

sunfishcode: Ok. I guess I'm already in the habit of thinking in terms of https://github.com/WebAssembly/design/pull/652. If we do that, we can simplify these tests.

rossberg-chromium:
Ping :)

rossberg-chromium:
Any opposition to landing these tests?

lukewagner:
lgtm

PR #295, opened 2016-06-03
Yet another batch of tests

rossberg-chromium:
This time:
- functions (func, param, result, local)
- call
- call_indirect
- get_local
- set_local

Also, syntax now disallows signs for any numbers that are uints in the binary encoding.


ml-proto/host/lexer.mll
```diff
@@ -88,18 +88,19 @@ let escape = ['n''t''\\''\'''\"']
...
 
-let sign = ('+' | '-')?
-let num = sign digit+
-let hexnum = sign "0x" hexdigit+
-let int = num | hexnum
+let sign = ('+' | '-')
+let num = digit+
+let hexnum = "0x" hexdigit+
+let nat = num | hexnum
+let int = sign nat
```
sunfishcode: Would it make sense to make `int` be `sign? nat`, so that productions like `literal` don't have to match both `int` and `nat`?

Also, what is `nat` short for?

kripken: Maybe "natural number"?

ml-proto/test/func.wast
```diff
@@ -0,0 +1,442 @@
...
+  (func "return-block-void" (return (block (i32.const 1) (call $dummy))))
+  (func "return-block-drop" (return (block (call $dummy) (i32.const 1))))
+  (func "return-block-i32" (result i32)
+    (return (block (call $dummy) (i32.const 77)))
+  )
+
+  (func "break-nullary" (br 0))
+  (func "break-void" (br 0 (call $dummy)))
+  (func "break-drop" (br 0 (i32.const 1)))
+  (func "break-i32" (result i32) (br 0 (i32.const 79)))
```
kripken: where is this `br` going to? do functions automatically generate a label scope?

ml-proto/host/lexer.mll
rossberg-chromium: @sunfishcode, tokens necessarily define disjoint lexical classes, so to carve out a subclass you need to union in the parser.

Yes, `nat` is "natural".

ml-proto/test/func.wast
rossberg-chromium: Yes. In particular, the spec defines `return` as a shorthand for the corresponding break.

kripken: Ok, thanks.

binji: Interesting, I thought we backtracked on that (guess I better fix sexpr-wasm :))

rossberg-chromium:
Any opposition to landing these tests as well?

lukewagner:
Nope

rossberg-chromium:
Merged into master manually, because Github.

PR #296, opened 2016-06-15
md-proto: a specification written in prose

sunfishcode:
Pursuant to the desire expressed in [README.md](https://github.com/WebAssembly/spec/blob/master/README.md) for a human-readable specification written in prose, this PR introduces _md-proto_, a specification document written in Github-flavored Markdown.

Markdown is very simple, and doesn't have support for some fancy formatting options like auto-generated tables of contents, footnotes, and so on, however so far in this document I've found it to be just sufficient enough to produce a relatively readable and concise spec.

The document here is not yet complete; most notably it currently focuses on function bodies and doesn't contain all the surrounding contents of modules, and some of the validation logic needs further work, however I hope there's enough content to prove the overall concept here.

### DIFF
--- a/md-proto/WebAssembly.md
+++ b/md-proto/WebAssembly.md
@@ -25,7 +25,7 @@ These types describe various data structures present in WebAssembly modules:
 0. [Array](#array)
 0. [String](#string)
 
-> These types are not used to describe values at runtime.
+> These types aren't used to describe values at runtime.
 
 #### Index
 
@@ -38,13 +38,13 @@ An *index* is a 32-bit unsigned integer value.
 An *array* is an [index] indicating a number of elements, plus a sequence of
 that many elements.
 
-> Array elements need not all be the same size in some representations.
+> Array elements needn't all be the same size in some representations.
 
 #### String
 
 A *string* is an [array] of bytes.
 
-> Strings in this context may contain arbitrary bytes and are not required to be
+> Strings in this context may contain arbitrary bytes and aren't required to be
 valid UTF-8 or any other format.
 
 ### Module Contents
@@ -61,7 +61,7 @@ and associated data.
  - Present known sections are checked to be ordered in the section sequence as
    they are ordered in the [enumeration of the Known Sections](#known-sections).
 
-> Some representations do not represent some of the known sections literally;
+> Some representations don't represent some of the known sections literally;
 they may be combined with other sections or implied by specialized syntax.
 
 > The initial release of WebAssembly will use an expected version of `0x0`.
@@ -141,6 +141,8 @@ The Memory Section contains:
 
 **Validation:**
  - The maximum size is checked to be at least the initial size.
+ - If the maximum size multiplied by the [page size](#pages) would be
+   unrepresentable in an unsigned `iPTR`, validation fails.
 
 #### Export Section
 
@@ -156,6 +158,9 @@ An export contains:
 **Validation:**
  - The function index of each array element is checked to be within the bounds
    of the [Code Section](#code-section) array.
+ - If any two exports have the same function string, validation fails.
+
+TODO: Memory exports (which require the presence of a Memory Section).
 
 #### Start Section
 
@@ -167,6 +172,8 @@ The Start Section contains a function [index]. See
 **Validation:**
  - The index is checked to be within the bounds of the
    [Code Section](#code-section) array.
+ - The function signature indexed in the [Type Section](#type-section) is
+   checked to have an empty parameter list and an empty return list.
 
 #### Code Section
 
@@ -190,22 +197,26 @@ format, positions are represented with a special syntax.
 
 **Name:** `data`
 
-The Data Section contains an [array] of [string] and offset pairs describing
-data to be loaded into linear memory as part of
+The Data Section contains an [array] of data initializers.
+
+A *data initializer* consists of a [string] and an offset. It describes data to
+be loaded into linear memory as part of
 [linear memory instantiation](#linear-memory-instantiation).
 
 **Validation:**
  - For each element of the array, the sum of the offset and the length of the
    string is checked to be less than the initial size declared in the
    [Memory Section](#memory-section).
+ - If any byte of linear memory would be initialized by more than one data
+   initializer, validation fails.
 
 #### Name Section
 
 **Name:** `name`
 
-The Names Section does not change execution semantics and malformed constructs,
+The Names Section doesn't change execution semantics and malformed constructs,
 such as out-of-bounds indices, in this section cause the section to be ignored,
-and do not trigger validation failures.
+and don't trigger validation failures.
 
 The Names Section contains an [array] of function name descriptors, which each
 describe names for the function with the corresponding index in the module, and
@@ -276,7 +287,7 @@ To merge two validation types:
 #### Function Validation Initialization
 
 If the instruction sequence is empty, or if the last instruction in the sequence
-is not an `end`, validation fails.
+isn't an `end`, validation fails.
 
 The current position starts at the first position. The type stack begins empty.
 The control-flow stack starts with one entry. This entry's [validation type] is
@@ -313,7 +324,10 @@ If the instruction's signature has no return types:
    limit value, validation fails.
  - `void` is pushed onto the type stack.
 
-TODO: Handle unused `void` values left behind on the stack.
+TODO: Handle unused `void` values left behind on the stack. Or, eliminate `void`
+values from the type stack altogether.
+
+> There are no implicit type conversions in WebAssembly.
 
 #### Function Return Validation
 
@@ -332,7 +346,7 @@ WebAssembly code execution requires an *instance* of a module, which contains a
 reference to the module plus additional information added during instantiation,
 which consists of the following steps:
  - The entire module is first [validated](#validation). If there are any
-   failures, instantiation aborts and does not produce an instance.
+   failures, instantiation aborts and doesn't produce an instance.
  - If a [Memory Section](#memory-section) is present,
    [Linear memory is instantiated](#linear-memory-instantiation).
  - [A recursion limit is selected](#recursion-limit).
@@ -380,8 +394,8 @@ are created:
    function.
  - A *current position*.
 
-> Implementations need not create a literal array to store the locals, or
-literal stacks to manage values at runtime.
+> Implementations needn't create a literal array to store the locals, or literal
+stacks to manage values at runtime.
 
 #### Function Execution Initialization
 
@@ -454,12 +468,12 @@ Types
 | `i32` | 32
 | `i64` | 64
 
-Integer types are not inherently signed or unsigned. They may be interpreted as
+Integer types aren't inherently signed or unsigned. They may be interpreted as
 signed or unsigned by individual instructions. When interpreted as signed, a
 [two's complement] interpretation is used.
 
 > The [minimum signed integer value] is supported; consequently, two's
-complement signed integers are not symmetric around zero.
+complement signed integers aren't symmetric around zero.
 
 #### Booleans
 
@@ -551,11 +565,12 @@ are the addressing unit of linear memory spaces.
 #### Effective Address
 
 The *effective address* of a linear memory access is computed by adding `$base`
-and `$offset` at infinite precision, so that there is no overflow.
+and `$offset`, both interpreted as unsigned, at infinite precision, so that
+there is no overflow.
 
 #### Alignment
 
-**Slow:** If the effective address is not a multiple of `$align`, the access is
+**Slow:** If the effective address isn't a multiple of `$align`, the access is
 *misaligned*, and the instruction may execute very slowly.
 
 > When `$align` is greater than or equal to the size of the access, the access
@@ -641,10 +656,10 @@ The called function &mdash; the *callee* &mdash; is
 recursion depth value plus one as its recursion depth value. The return value of
 the call is defined by the execution.
 
-**Trap:** Call stack overflow, if the recursion depth value is greater than the
+**Trap:** Call Stack Overflow, if the recursion depth value is greater than the
 [recursion limit](#recursion-limit).
 
-> This means that implementations are not permitted to perform implicit
+> This means that implementations aren't permitted to perform implicit
 opportunistic tail-call elimination.
 
 > The execution state of the function currently being executed remains live
@@ -652,7 +667,7 @@ during the call, and the execution of the called function is performed
 independently. In this way, calls form a stack-like data structure called the
 *call stack*.
 
-> Implementations need not pass a literal argument to represent the recursion
+> Implementations needn't pass a literal argument to represent the recursion
 depth.
 
 #### Call Validation
@@ -672,8 +687,8 @@ WebAssembly comparison instructions compare two values and return a [boolean]
 result value.
 
 > In accordance with IEEE 754-2008, for the comparison instructions, negative
-zero is considered equal to zero, and NaN values are not less than, greater
-than, or equal to any other values, including themselves.
+zero is considered equal to zero, and NaN values aren't less than, greater than,
+or equal to any other values, including themselves.
 
 ### T: Shift Instruction Family
 
@@ -695,37 +710,37 @@ instructions such as [`shr_s`](#integer-shift-right-signed).
 
 ### G: Generic Integer Instruction Family
 
-Except where otherwise specified, these instructions do not specifically
-interpret their operands as explicitly signed or unsigned, and therefore do not
+Except where otherwise specified, these instructions don't specifically
+interpret their operands as explicitly signed or unsigned, and therefore don't
 have an inherent concept of overflow.
 
 ### S: Signed Integer Instruction Family
 
 Except where otherwise specified, these instructions interpret their operand
 values as signed, return result values interpreted as signed, and [trap] when
-the result value can not be represented as such.
+the result value can't be represented as such.
 
 ### U: Unsigned Integer Instruction Family
 
 Except where otherwise specified, these instructions interpret their operand
 values as unsigned, return result values interpreted as unsigned, and [trap]
-when the result value can not be represented as such.
+when the result value can't be represented as such.
 
 ### F: Floating-Point Instruction Family
 
 Instructions in this family follow the [IEEE 754-2008] standard, except that:
 
- - They support only "non-stop" mode, and floating-point exceptions are not
+ - They support only "non-stop" mode, and floating-point exceptions aren't
    otherwise observable. In particular, neither alternate floating-point
    exception handling attributes nor the non-computational operations on status
    flags are supported.
 
  - They use the IEEE 754-2008 `roundTiesToEven` rounding attribute, except where
-   otherwise specified. Non-default directed rounding attributes are not
+   otherwise specified. Non-default directed rounding attributes aren't
    supported.
 
 When the result of any instruction in this family (which excludes `neg`, `abs`,
-and `copysign`) is a NaN, the sign bit and the significand field (which does not
+and `copysign`) is a NaN, the sign bit and the significand field (which doesn't
 include the implicit leading digit of the significand) of the NaN are computed
 by one of the following rules, selected [nondeterministically]:
 
@@ -733,7 +748,7 @@ by one of the following rules, selected [nondeterministically]:
    may [nondeterministically] select any of them to be the result value, but
    with the most significant bit of the significand field overwritten to be `1`.
 
- - If the implementation does not choose to use an input NaN as a result value,
+ - If the implementation doesn't choose to use an input NaN as a result value,
    or if there are no input NaNs, the result value has a [nondeterministic] sign
    bit, a significand field with `1` in the most significant bit and `0` in the
    remaining bits.
@@ -742,7 +757,7 @@ TODO: Monitor https://github.com/WebAssembly/design/pull/713.
 
 Implementations are permitted to further implement the IEEE 754-2008 section
 "Operations with NaNs" recommendation that operations propagate NaN bits from
-their operands, however it is not required.
+their operands, however it isn't required.
 
 > All computations are correctly rounded, subnormal values are fully supported,
 and negative zero, NaNs, and infinities are all produced as result values to
@@ -906,7 +921,7 @@ stack bottom. Its return value is the value of its operand, if it has one.
    control-flow stack bottom.
  - `any` is pushed onto the type stack.
 
-> Implementations need not literally perform a branch before performing the
+> Implementations needn't literally perform a branch before performing the
 actual function return.
 
 #### Unreachable
@@ -920,7 +935,7 @@ actual function return.
 **Validation:**
  - `any` is pushed onto the type stack.
 
-> The `unreachable` instruction is meant to represent code that is not meant to
+> The `unreachable` instruction is meant to represent code that isn't meant to
 be executed except in the case of a bug in the application.
 
 #### End
@@ -1012,7 +1027,7 @@ the value given in the operand.
 | `tee_local` | `<$id: i32> (T) : (T)`      |          | 0x19
 
 The `tee_local` instruction sets the value in the locals array at index `$id` to
-to the value given in the operand. Its return value is the value of its operand.
+the value given in the operand. Its return value is the value of its operand.
 
 #### Select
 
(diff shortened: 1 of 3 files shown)


sunfishcode:
[Here is a rendered view](https://github.com/WebAssembly/spec/blob/md-proto/md-proto/WebAssembly.md) of the proposed document.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1693 @@
...
+
+> `any` is effectively serving as a kind of bottom type, and `void` is
+effectively serving as a kind of top type. Note that in some type theory
+contexts, it's common to use these names for exactly opposite purposes.
+
+#### Type Stack Pop
+
+To perform a pop from the type stack:
+ - If the type stack's size is less than the control flow stack's top's depth
+   value, return `void`.
```
JSStats: Might it be clearer to just describe this as failing validation if attempting to pop past the top depth?

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1693 @@
...
+ - Otherwise pop a valid from the type stack and return that.
```
JSStats: 'a valid' ?

sunfishcode: Fixed.

md-proto/WebAssembly.md
sunfishcode: It doesn't necessarily fail validation though. Also, in general, the validation code does need further work.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1693 @@
...
+are no instructions following it, to the return position. Then the remembered
+instruction is validated as follows.
+
+If the instruction has a **Validation** clause, that clause describes its
+validation. Otherwise, the following generic validation is performed:
+
+For each operand type in the instruction's signature, a type is popped from the
+type stack and checked to be the same as the corresponding operand type. Each
+return type of the instruction's signature is then pushed onto the type stack.
+
```
JSStats: It might only be the `call` operator that returns multiple values, and prior discussions pointed to storing the results in local variables, with perhaps the first value as a single expression value. So here it is proposed to have one stack element for each of these values, which might be fine, but then we need the support to work with them efficiently, and if they are dropped or popped in order then this fits the pattern, but if needed in some other order then it seems that the only option is to store them in a local until needed?

sunfishcode: I am hoping to focus on the overall approach of this document first, rather than dive into details of validation which are  being discussed elsewhere.

JSStats: Ok, the approach looks great to me, good to have a written specification. Might it need the W3C template eventually?

sunfishcode: If we pursue W3C standardization, we can adjust the formatting as needed. The document is written in Markdown which is pretty easy to convert to other formats as needed.

sunfishcode:
I've now added the initial text describing modules, including all the *known sections*. It's a little tricky separating what belongs in the description of modules from what belongs in the binary format description, and there's room for further iteration here, but hopefully it's a good start for discussion.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1884 @@
...
+instruction is validated as follows.
+
+If the instruction has a **Validation** clause, that clause describes its
+validation. Otherwise, the following generic validation is performed:
+
+For each operand [type] in the instruction's signature, a type is popped from
+the type stack and checked to be the same as the corresponding operand type.
+Each return type of the instruction's signature is then pushed onto the type
+stack.
+
```
JSStats: This might be ok, but this is proposing to push *no* values on the type stack if the operator returns no values, and currently a `void` type element is pushed for an operator returning no values (even for `drop` which does no change the size of the type stack).

`drop` did not change the size of the type stack for good reason, so that `i32_const 1, i32_const 2, drop, i32.eqz` is not valid - it's not really popping rather writing over the top of the stack with the `void` type. If we change the semantics to push each value to separate stack elements then I am not sure if the code would have such a simple text format with expressions. (Assuming `void` is the type of an operator returning no values)

For example, how would the second value in `call $fn_returning_i32_i64` even be invalidated (or `drop`ped), as `drop` just writes over the top-of-stack type.

Another other path to take would be to embrace this proposal and not push a type on the stack for an operator returning no values, and this would eliminate the `void` type. The `drop` operator would then pop the stack.

There are different ways this could go but this description does not appear to present a clear option, not a clear stack machine if that was the intention.

If we are going to keep the `void` type, then `drop` could be extended to invalidate any entry up the stack, not just the top.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1884 @@
...
+If the instruction's signature has no return types, and the size of the value
+stack is greater than the control-flow stack's top's depth, validation fails
+with a type error.
+
```
JSStats: This uses the term 'values stack' but before 'type stack'. I don't think this would work at present, as `i32.cont 1, drop, return 0` does not leave an empty type stack at the return point, rather a stack with one `void` type element. CIL seems to have this empty-stack requirement at return points, but that does not make it a good thing to follow, and CIL is a stack machine with other burdens and failed to look ahead to multiple-value support.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1884 @@
...
+Then, [validation](#function-body-validation) is resumed.
+
+#### Function Return Validation
+
+When a *return* is prompted, one [type] for each type in the function
+signature's return type is popped from the type stack and type-checked against
+its corresponding return type. The type stack is then checked to be empty. If
+there are sufficient stack elements and all checks have passed, function
+validation is successful.
+
```
JSStats: The description does not make it clear if 'type stack is then checked to be empty' is not true then is it a validation failure, and if not then why bother checking?

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1884 @@
...
+### Function Execution
+
+Function execution can be prompted by a call instruction, by module execution,
+or by the embedding environment.
+
+The input to execution of a function consists of:
+ - The function to be executed.
+ - The incoming argument values, one for each argument [type] of the function.
+ - A *depth* value.
```
JSStats: What is the '*depth* value' passed to a function?

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,1884 @@
...
+ - Otherwise, if the control-flow types are the same, the result is that control
+   flow type.
+ - Otherwise, validation fails with a type error.
+
+#### Function Validation Initialization
+
+The current position starts at the first position. The type stack begins empty.
+The control-flow stack starts with one entry. This entry's control-flow type is
+the first return type of the function, if it has any, or `void` otherwise. Its
+block depth is zero.
```
JSStats: What about the other return types of the function? Would this mean that a `br` could only return a single value, and that a `return` would be needed to return multiple-values?

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2007 @@
...
+### Function Execution
+
+Function execution can be prompted by a [call-family instruction][L], by
+[module execution](#module-execution), or by the embedding environment.
+
+The input to execution of a function consists of:
+ - The function to be executed.
+ - The incoming argument values, one for each argument [type] of the function.
+ - A *recursion depth* value.
+
```
JSStats: Again, what is this 'recursion depth' value??

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2007 @@
...
+For the duration of the execution of a function body, several data structures
+are created:
+ - A *control-flow stack*, which holds [labels] for reference from branch
+   instructions.
+ - A *value stack*, which carries values between instructions.
+ - A *locals* array, containing an element for each argument of the function in
+   the argument's [type], as well as an element for each local declaration in
+   the function.
+ - A *current position*.
```
JSStats: You could consider having the arguments pushed onto the type stack, rather than stored in locals, which would parallel the results of a operation being pushed.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2007 @@
...
+For each operand [type] in the instruction's signature in reverse order, a value
+is popped from the value stack and provided as the corresponding operand value.
+The instruction is then executed as described in the
+[Instructions Section](#instructions) entry describing it. Each of the
+instruction's return values are then pushed onto the value stack.
+
+If the current position is now past the end of the sequence,
+[function return execution](#function-return-execution) is initiated and
+execution of the function is thereafter complete.
+
```
JSStats: Why not check that values are not popped past the control-flow blocks top-of-stack? It seems really odd that this would allow crossing a control-flow stack-top boundary when popping arguments, and then allow the results to be pushed back, and only then check if the final stack position against the control-flow stack-top boundary?

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2007 @@
...
+immediately terminated and abnormal termination is reported to the embedding
+environment.
+
+#### Function Return Execution
+
+One value for each return [type] in the function signature in reverse order is
+popped from the value stack. If the function execution was prompted by a call
+instruction, these values are provided as the call's return value. Otherwise,
+they are provided to the embedding environment.
+
```
JSStats: Right, so the tailing result values could be dropped. `call $fn_returning_two_values, drop, return` would return one value, the first.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2007 @@
...
+Otherwise, [validation](#function-body-validation) is resumed.
+
+TODO: `unreachable`, `br`, `return`, and `br_table` return type validation.
+
+#### Generic Instruction Validation
+
+For each operand [type] in the instruction's signature in reverse order, a
+type is popped from the type stack and checked to match it. Each return type of
+the instruction's signature is then pushed onto the type stack.
+
```
JSStats: So, can this be read as `nop` having no 'return type's, and thus pushing no values? This is perfectly workable, but what worries me is that now the following is valid `i32.const 1, i32.const 2, nop, i32.add` and what would this look like in the wip text format?

It might be presentable as the following:
```
{
   const i32 c1 = 1;
   const i32 c2 = 2;
   nop;
   c1 + c2;
}
```

But this would be a very brittle language, values would need to be written and consumed in the stack ordering. While presentable, it would not be something anyone would want to write.

md-proto/WebAssembly.md
sunfishcode: The other mentions of the phrase "recursion depth" are meant to explain what it's used for.

md-proto/WebAssembly.md
sunfishcode: This would be a significant language change which ought to be proposed in its own PR.

md-proto/WebAssembly.md
sunfishcode: This is the execution section; rejecting programs that pop past the control-flow stack top's limit is handled in the validation section.

md-proto/WebAssembly.md
sunfishcode: The current spec doesn't support multiple return values in general.

md-proto/WebAssembly.md
sunfishcode: The example you give here is caught by the validation rule in the very next paragraph, since `nop` has no return values.

JSStats: @sunfishcode Hmm, sorry I'm not following this. The validation rule below does not seem to affect this, or that is my reading of it. The `i32.const` operators will push two values on the stack and if the `nop` pushes no new values then then `i321.add` has sufficient arguments and would seem to pass this validation rule. Perhaps I am looking at this from the wrong point of view, and perhaps it's obvious to you?

My concern with the move to a stack machine is that the stack machine code might not have an good parallel in an expression based text format.

Also how do all the mandatory `drop` operators look in the text format now? Currently each needs an AST node as an argument, but if operators with no results do not push to the stack then the following would be valid too and how would it be presented? `i32.const 1, i32.const 2, drop, drop` versus `i32.const 1, drop, i32.const 2, drop`

md-proto/WebAssembly.md
JSStats: @sunfishcode Ok, thanks, so this is defining the recursion depth limit. Perhaps this is an implementation detail and need not be defined so explicitly, rather a general note that the stack is limited etc.

md-proto/WebAssembly.md
JSStats: @sunfishcode You do appear to be proposing some significant language changes here without them being raised in design PRs? If you do not want to discuss them here then perhaps this PR should stick to the status quo.

md-proto/WebAssembly.md
JSStats: @sunfishcode Ok, thanks, I misread this, but perhaps other people will. Adding your response 'rejecting programs that pop past the control-flow stack top's limit is handled in the validation section' to this section might help people.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2007 @@
...
+Then, if the instruction's signature has no return types, and the length of the
+type stack is greater than the control-flow stack top's limit value, validation
+fails.
+
```
JSStats: Re-asking this in the *validation* section. Why not check that values are not popped past the control-flow blocks top-of-stack? It seems really odd that this would allow crossing a control-flow stack-top boundary when popping arguments, and only then (and only for operators returning no values) check the final stack position against the control-flow stack-top boundary?

md-proto/WebAssembly.md
sunfishcode: I'm not aware of any significant changes, other than things that are actively being discussed elsewhere, and things which are marked with TODO comments.

md-proto/WebAssembly.md
sunfishcode: It's checked for in the Function Validation section, because popping past the control-flow stack top's limit is something that applies in all situations, not just when validating instructions that use generic validation.

md-proto/WebAssembly.md
sunfishcode: The `nop`s signature has no return types, and after the `i32.const` instructions the length of the type stack will be greater than the control-flow stack top's limit value, so validation fails.

JSStats: @sunfishcode Sorry, I am still lost. Let me try to describe the stack.

```
Initial stack: ..., block start
i32.const 1: ..., block start, i32 const 1,
i32.const 2: ..., block start, i32 const 1, i32.const 2
nop: ..., block start, i32 const 1, i32.const 2
```
Now why would validation fail here, it has not popped past the block top?

md-proto/WebAssembly.md
JSStats: @sunfishcode Yes, thank you, I see the `pop` validate check now, but am still lost on what the purpose of the above clause is? If the pop's are already validated then how could the above then fail validation, and what does it have to do with a lack of results for an instruction?

md-proto/WebAssembly.md
sunfishcode: The paragraph I indicated says:

> Then, if the instruction's signature has no return types, and the length of the
type stack is greater than the control-flow stack top's limit value, validation
fails.

Nop's signature has no return types. And in your example, the type stack is greater than the control-flow stack top's limit value after the `nop`. Consequently, validation fails.

md-proto/WebAssembly.md
sunfishcode: The purpose of the clause is to reject the kind of code in the example [above](https://github.com/WebAssembly/spec/pull/296#discussion_r68485930).

sunfishcode:
Another nice thing about Markdown is that it automatically creates HTML anchors for all the headings, so one can refer to individual spec paragraphs using links. For example, say I want to refer to [xor](https://github.com/WebAssembly/spec/blob/md-proto/md-proto/WebAssembly.md#integer-bitwise-exclusive-or).

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2037 @@
...
+    - A *limit* value.
+ - A *[type] stack*.
+ - A *current position*.
+
+During function validation, if either stack is empty when an element is to be
+popped from it, or an element is accessed by index less than zero or not less
+than the stack's length, validation fails. If a [validation type] is popped from
+the type stack such that the length of the type stack becomes less than the
+control-flow stack top's limit value, validation fails.
+
```
JSStats: Just a suggestion but 'greater than or equal to the stacks length' reads better to me than 'not less
+than'. It might be obvious to you, but perhaps defining a direction for the stack would help to. Then the spec could consistently use 'less' for a check at one end and 'greater' for a check at the other end.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2037 @@
...
+#### Generic Instruction Validation
+
+For each operand [type] in the instruction's signature in reverse order, a
+[validation type] is popped from the type stack and checked to match it. Each
+return type of the instruction's signature is then pushed onto the type stack.
+
+If the instruction's signture has no return types:
+ - If the length of the type stack is greater than the control-flow stack top's
+   limit value, validation fails.
```
JSStats: Ok, I think I understand the proposal now, thank you. But this seems a significant change, and would block multiple use of values on the stack, block calls pushing multiple values on the stack, a very limiting constraint for wasm. It's fine for an expressionless encoding as then there are no values on a values stack, and this seems to be where this is heading, to a heavy dependence on local variables which might be fine.

Adding back the pushing of the `void` type will also break the potential multiple value call support. It will require multiple value results be written directly to local variables rather than pushed on the stack, for example:

```
call $fn_returning_two_values : pushing two values on the stack.
set_local $l1 : popping one value, and pushing void.
set_local $l2 : fails because either popping the void or the stack is not empty.
```
This would fail to validate under either constraint. Either the stack is not empty or a `void` type is push and the first call result push could no be consumed.

I don't think these are even compatible constraints: as then `(block (nop) (nop))` would be invalid as the first pushes `void`.

sunfishcode: I've now removed all language which started to anticipate multiple return values.

The `(nop) (nop)` problem is interesting. This is a consequence of the new explicit drop rule which I hadn't yet discovered. Real implementations will presumably need to solve this too. I added a TODO about it. It may be that a no-push-void approach is a good way to handle this, but no-push-void is still experimental in other ways, and I don't want it to be a distraction here.

md-proto/WebAssembly.md
sunfishcode: I've now reworded this to just say "out of bounds" instead of the awkward "less than 0 or not less than the size" phrasing.

sunfishcode:
Several notable updates:
 - A new Introduction section has been added which introduces WebAssembly and the major concepts, with links into the main spec text.
 - Function validation has been significantly revised, with an emphasis on the high-level properties that are being checked and on the validation invariants, which helps put the validation rules in context.
 -The "recursion limit" concept that several people found confusing has now been replaced by a concept of abstract "call-stack resources".
 - More TODO comments have been added to point out areas that are known to be particularly "imcomplet" and "inkorrect".

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2326 @@
...
+
+The unit of WebAssembly code is the *module*. Modules primarily consist of a
+collection of sections. There are sections describing a WebAssembly's
+interactions with other modules ([imports](#import-section) and
+[exports](#export-section)), sections declaring [data](#data-section) and other
+implements used by the module, and sections defining [*functions*]. For more
+information, see the [Module section](#module).
+
+WebAssembly code must be validated before it can be instantiated and executed.
+WebAssembly is designed to allow validation to be performed in in a single
```
JSStats: nit 'in in'

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2326 @@
...
+
+### Function Validation
+
+The major requirements for function validation are:
+ - Control-flow constructs must form properly nested *regions*. `loop`, `block`,
+   and the function entry pair with `end`, and `if` can pair with `end` or
+   `else`, which is then paired with `end`.
+ - The number and types of all elements that can be on the value stack at
+   execution of each instruction is required be the same for all control-flow
+   paths to that instruction.
```
JSStats: Was pondering if the order needs to be the same too? Does this clause make a `swap` operator problematic?

Also it seems that the one exception would be setting a stack element to a value that could not be consumed because if it were then it would be statically detected and the code would not validate? This seems to be permitting the `drop` operator.

md-proto/WebAssembly.md
sunfishcode: Fixed.

md-proto/WebAssembly.md
sunfishcode: The order needs to be the same too; I've now reworded the text to say "the sequence of types" has to be the same, which should imply number, types, and order. A `swap` operator would be theoretically ok, as long as it still satisfies the property that at each instruction, every path gives it a stack with the same sequence of types.

The problem of conflicting types in a control-flow merge where the result is ultimately unused is indeed something that will need further attention. I've added a TODO.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2370 @@
...
+#### Generic Instruction Validation
+
+For each operand [type] in the instruction's signature in reverse order, a
+[type] is popped from the type stack and required to match it. Each return type
+of the instruction's signature is then pushed onto the type stack.
+
+If the instruction's signature has no return types:
+ - If the length of the type stack is greater than the control-flow stack top's
+   limit value, validation fails.
+
```
JSStats: This constraint does not seem practical as the following would not validate: `call $fn_returning_two_values, drop`

Also if this design is going to practically require modifying the stack, via `swap` operators etc, then it seems necessary to compare all the stack elements at each control flow path merge point and this would lose the advantage that the constant stack elements had which is that they are out of the set of definitions needing to be compared at these points. The expressionless encoding has this same property that all the definitions need to be compared a each control flow merge point, and it is one of it's weaknesses, although it might have been mitigated if values could be invalidated or zeroed.

I see paths forward, but the direction this is going is not one of them. Could concede to the expressionless encoding which has many advantages, but code size might suffer and it will need meta information to be readable. Could concede to the `get_value` proposal, have constant definitions build up on the stack within blocks and optimize for this by discarding them all at the end of the block etc, this path seems to lead to more readable code, but a disadvantage is that the stack of definitions may be large.

sunfishcode: Functions with multiple return values are not permitted in this version of the spec. If they are added, I agree that they would want to be accompanied by a mechanism to access their return values in a flexible way.

md-proto/WebAssembly.md
```diff
@@ -0,0 +1,2357 @@
...
+    - If the element is a [type], a type is pushed onto the type stack.
+    - If the element is a list of type parameters, they are required to be
+      bound, and the type bound to each type parameter is pushed onto the type
+      stack.
+ - If the instruction has a **Validation** clause, its requirements are
+   required.
+ - If the instruction's signature has no return types, or it is a
+   [control-flow barrier][Q], the new length of the type stack is required to be
+   at most the control-flow stack top's limit value.
+
```
JSStats: The problem is not just one of accessing these values, rather this constraint does not even permit values to be stored in locals for further use, even if they could be accessed. Even with a `get_value` operator, e.g. `call $fn_returning_two_values, get_value 1, store_local $l1` would be invalid.

A `drop_all` operator seems possible for the case in which none of the results were required. A `multiple_set_local` operator seems possible that would transfer all the values into locals but then the call operator might as well just write all the results directly into locals and we end up at the expressionless encoding.

It is possible to convert an AST with multiple value support into a stack machine code but perhaps what is missing are practical constraints that would limit this stack machine code to being presentable as an AST. That is, perhaps wasm can be an AST or it can be a stack machine but not both, and this has been heading down the path of the stack machine. Perhaps there are some constrains, but if they are not known to anyone then that seem a blocker, unless people want to concede the AST text format and have a stack machine text format.

sunfishcode:
This draft now has validation and execution semantics for all instructions. The validation logic has been significantly simplified and refactored to highlight its invariants.

Several TODOs remain, though the majority of them are now in areas where there are open spec change discussions.

JSStats:
Github is reporting 'not shown because the diff is too large' for this now making it impossible to comment in the changes. Might a rebase help.

sunfishcode:
The file is over github's limit to display as a single diff; it doesn't look like rebasing will change that. A full rendered view of the document can be viewed [here](https://github.com/WebAssembly/spec/blob/md-proto/md-proto/WebAssembly.md).

sunfishcode:
Closing this PR, which was never granted the dignity of a response.

shurcooL:
> The file is over github's limit to display as a single diff; it doesn't look like rebasing will change that.

I think there have been changes on GitHub, but it loads fine now. It requires pressing a "Load diff" button over the large file.

PR #297, opened 2016-06-17
Fix tests broken @ 0507cf1f (func_fields order)

binji:


rossberg-chromium:
LGTM, thanks! Not sure why I didn't run into these when I rebased the branch...

PR #298, opened 2016-06-20
Make dropping of values explicit

rossberg-chromium:
As of WebAssembly/design#694. This is a big change: it modifies the static & dynamic semantics, introduces `drop` and `tee_local` operators (with tests), and adapts tons of existing tests.

While there I also updated some opcode uses to match the current Binary.md


kripken:
What does "Require arity 1 for drops and breaks" mean?

rossberg-chromium:
@kripken, that uses like `(drop (nop))` are disallowed, where the the operand has type void (a.k.a., is a tuple with arity 0). That was a further simplification requested by Luke.

kripken:
And the same for breaks, so `(br (nop))` is disallowed too?

rossberg-chromium:
Yes, or `(return (nop))`. This is consistent with calls, as `(call $f (nop))` was never allowed either.

kripken:
Sounds fine, then I was just confused by the title "Require arity 1 for drops and breaks" which seemed to suggest that a value is needed, i.e., break without a value isn't possible? Is there some other meaning to arity here that is confusing me?

rossberg-chromium:
Sorry, I agree the patch name may be misleading. Arity is a bit overloaded: there is syntactic arity (number of arguments) and there is semantic arity (number of values). If we e.g. allowed multiple values, then you would envision allowing things like `(return (call $func-returning-multiple-values))` where the syntactic arity is 1 but the semantic arity is larger.

kripken:
I see, thanks.

JSStats:
Decisions need to be made on the handling of multiple values - the semantics need to be decided. If this matter is not given some attention then wasm will end up in the same place that CIL did where multiple values need to be handled by passing around a structure value which would be very poor outcome given that we already have two available options to consider and explore:

1. Use local variables and have calls write results to local variables. This is the expressionless proposal.

2. Write multiple values to separate values stack elements and include good support for accessing them, such as the `get_value` proposal. Subjectively this gives nicer code to read as each value stack element can be presented as a block local constant variable.

In both cases `return` needs to have the semantics that it accepts a list of single values, and there are no multiple value expressions. In neither case do we have a stack machine.

rossberg-chromium:
This has been rebased and merged to binary-0xc branch.

PR #300, opened 2016-06-24
Updates for binary format

### LINKED ISSUE #299: (module "...") form is parsed incorrectly
This looks like a merge error: some lines are transposed in the section parser.  The test cases do not catch this because all the tests are trivial (no sections) or should fail anyway.  Better test case:

(module "\00asm\0b\00\00\00\04type\04\01\40\00\00\08function\02\01\00\04code\03\01\00\00")

Also, the expected function type is wrong (0x05 rather than 0x40 as documented).
```
diff --git a/ml-proto/spec/decode.ml b/ml-proto/spec/decode.ml
index b3b4790..681e383 100644
--- a/ml-proto/spec/decode.ml
+++ b/ml-proto/spec/decode.ml
@@ -124,7 +124,7 @@ let value_type s =
 let expr_type s = vec1 value_type s
 
 let func_type s =
-  expect 0x05 s "invalid function type";
+  expect 0x40 s "invalid function type";
   let ins = vec value_type s in
   let out = expr_type s in
   {ins; out}
@@ -447,9 +447,9 @@ let id s =
 let section tag f default s =
   if eos s then default else
   let start_pos = pos s in
+  if id s <> tag then (rewind start_pos s; default) else
   let size = vu s in
   let id_pos = pos s in
-  if id s <> tag then (rewind start_pos s; default) else
   let s' = substream s (id_pos + size) in
   let x = f s' in
   require (eos s') s' (pos s') "junk at end of section";
```

rossberg-chromium:
To sync with more recent changes in BinaryFormat.md. Also bumps version to 0xc.

Fixes #299.

### DIFF
--- a/ml-proto/host/encode.ml
+++ b/ml-proto/host/encode.ml
@@ -1,6 +1,6 @@
 (* Version *)
 
-let version = 0x0b
+let version = 0x0c
 
 
 (* Encoding stream *)
@@ -85,7 +85,7 @@ let encode m =
     let expr_type t = vec1 value_type t
 
     let func_type = function
-      | {ins; out} -> u8 0x05; vec value_type ins; expr_type out
+      | {ins; out} -> u8 0x40; vec value_type ins; expr_type out
 
     (* Expressions *)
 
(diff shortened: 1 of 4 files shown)


kripken:
Are we not doing a `0xc` branch in this repo as we are in design?

rossberg-chromium:
@kripken, good point, taking that out. All other changes should match what's on trunk in /design.

rossberg-chromium:
Any concerns with landing this?

lukewagner:
lgtm

PR #301, opened 2016-06-27
Refine and extend tests for function signatures

rossberg-chromium:
- Extend call_indirect test to check that signature is nominal.
- Avoid use of implicit signatures in call_indirect.wast tests.
- Add tests to func.wast to verify correct implicit signature desugaring.

### DIFF
--- a/ml-proto/test/call_indirect.wast
+++ b/ml-proto/test/call_indirect.wast
@@ -15,6 +15,7 @@
   (type $i32-i64 (func (param i32 i64) (result i64)))
   (type $f64-f32 (func (param f64 f32) (result f32)))
   (type $i64-f64 (func (param i64 f64) (result f64)))
+  (type $over-i64-duplicate (func (param i64) (result i64)))
 
   (func $const-i32 (type $out-i32) (i32.const 0x132))
   (func $const-i64 (type $out-i64) (i64.const 0x164))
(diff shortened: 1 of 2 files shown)


kripken:
It seems rather odd and surprising to "reuse" function signatures in this way. Why not just create a new signature when one is not specified? In practice this will just matter for small tests anyhow.

rossberg-chromium:
I'd be fine with changing it, although that would create rather pathological type sections with lots of redundancy. Deferring to @lukewagner, since he implemented this.

lukewagner:
Actually, design/#682 just changed AstSemantics.md#calls to have structural equality of function signatures (motivated by dynamic linking which would otherwise need some first-class notion of type that had to be imported/exported so it could be nominally equal).  There is not special case for "nominal when intra-module, structural when inter-module", so I think we should spec and test the opposite of what's in this PR.

rossberg-chromium:
PTAL. Removed the checks for nominal type tests, now superseded by #314.

kripken:
Looks like this PR still adds tests that "reuse" function signatures implicitly, which as discussed above seems rather confusing. In the discussion above I don't see any objections to not doing that?

kripken:
To be more clear, what I'm proposing is to not merge the changes left in this PR.

And #314 changes checks to be structural which is great.

rossberg-chromium:
Well, we should have tests for the desugaring, and we need to agree what the desugaring is.

With `call_indirect` checks becoming structural I think the current desugaring actually has no longer any draw-backs, but some pros: it avoids artificially blowing up the type section and produces exactly the code that you'd otherwise produce manually.

kripken:
Ok, fair enough.

kripken:
(lgtm)

PR #305, opened 2016-07-05
Add tests for conflicting-but-ultimately-dropped values.

sunfishcode:
This is a PR against the void branch, which adds a few more tests.

Specifically, this tests constructs which merge values from two control paths, to ensure that the types match even when the resulting value is dropped.

### DIFF
--- a/ml-proto/test/br.wast
+++ b/ml-proto/test/br.wast
@@ -418,7 +418,7 @@
   "unknown label"
 )
 
-;; Test that the value operand of br_if matches the result type of the block
+;; Test that the value operand of br matches the result type of the block
 ;; it branches to, even if the result of the block is dropped.
 (assert_invalid
   (module
--- a/ml-proto/test/br.wast
+++ b/ml-proto/test/br.wast
@@ -417,3 +417,26 @@
   (module (func $large-label (br 0x100000001)))
   "unknown label"
 )
+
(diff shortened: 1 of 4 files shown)


ml-proto/test/br.wast
```diff
@@ -417,3 +417,26 @@
   (module (func $large-label (br 0x100000001)))
   "unknown label"
 )
+
+;; Test that the value operand of br_if matches the result type of the block
```
kripken: comment says br_if, but no br_if in test?

ml-proto/test/br_table.wast
```diff
@@ -1385,3 +1385,25 @@
   "unknown label"
 )
 
+;; Test that the value operand of br_if matches the result type of the block
```
kripken: also a copy-pasted br_if

ml-proto/test/br.wast
sunfishcode: fixed, thanks!

ml-proto/test/br_table.wast
sunfishcode: fixed, thanks!

rossberg-chromium:
LGTM

sunfishcode:
Merging with LGTM, and note that this is just on the void branch, rather than master at this time.

PR #311, opened 2016-07-15
Add various verification tests for 'memory'

icefox:


rossberg-chromium:
You only need to write .fail. tests for cases that produce syntax errors. All other cases are better served by more concrete `assert_invalid` or assert_trap` checks. Also, a number of tests in this PR already exist, check out `memory.wast` in particular.

icefox:
Ah sorry I completely missed the existing checks in memory.wast, i'll close this and review it to see if I added any new checks

PR #315, opened 2016-08-02
Use assert_return_nan for test of fxx.min and fxx.max with nan inputs.

gahaas:
The spec tests for f32.min, f32.max, f64.min, and f64.max with NaN inputs were more strict than required by the design doc. The changes in this PR adjust these spec tests to the requirements in the design doc.


sunfishcode:
LGTM.

The NaN propagation semantics have changed, and the tests have not been updated for the changes in general. This PR for min/max is a good start.

PR #316, opened 2016-08-04
Implement tables & multiple memories

rossberg-chromium:
As of WebAssembly/design#682. Still misses import/export abilities.

This changes & extends the S-expression syntax for tables and memories in the following way:
```
<elem_type>: anyfunc
<table>:     (table <nat> <nat>? <elem_type>)
<memory>:    (memory <nat> <nat>?)
<elem>:      (elem <expr> <var>*)
<data>:      (data <expr> <string>*)
```
In particular, memory segments are no longer part of the `memory` definition, in anticipation of the ability to import memory. Same for tables. This also mirrors the Wasm section structure more closely.

However, it is pretty tedious to count table elements in the common case, so the following shorthand is available:
```
(table <elem_type> (elem <var>*))    ;; = (table <size> <size> <elem_type>) (elem (i32.const 0) <var>*)
```
which pretty much behaves like the previous table syntax.
For symmetry, I introduced the analogous shorthand for memories, which turns out to be useful in tests as well:
```
(memory (data <string>*))   ;; = (memory <size> <size>) 
(data (i32.const 0) <string>*)
```
where `<size>` is the strings' total length in (rounded up) page units.

In the future, we can extend this syntax to name tables and memories, e.g.:
```
<table>:   (table <name>? <nat> <nat>? <elem_type>)
<memory>:  (memory <name>? <nat> <nat>?)
<elem>:    (elem <var>? <expr> <var>*)
<data>:    (data <var>? <expr> <string>*)
```


kripken:
What is the plan with the `stack` branch - will this be merged there as well as to `binary-0xc`? Or will `stack` eventually be merged to `binary-0xc` without doing that first?

rossberg-chromium:
Excellent question. I meant to merge everything into binary-0xc eventually. But I'm still cleaning up the stack branch somewhat, and I suppose reviewing it may also take some time. So I didn't want everything else to block on that.

Guess that means that I gonna have a merge-conflict fest at some point...

kripken:
In the tests it looks like `data` can either be nested in `memory` or not. What is the difference?

rossberg-chromium:
See the first comment above. ;)

kripken:
Ok, sorry, I thought it was just about tables.

lukewagner:
The s-expr changes lgtm

lukewagner:
I think we also need new syntax for importing a tables/memories; we currently use:
  `( import <name>? <string> <string> ( table <nat> <nat>? <elem_type> ) )`
  `( import <name>? <string> <string> ( memory <nat> <nat>? ) )`

rossberg-chromium:
Any concerns about landing this?

lukewagner:
righto, generally lgtm

Issue #286, opened 2016-05-12
Implement the new NaN semantics


sunfishcode:
The reference interpreter and testsuite should be updated to reflect https://github.com/WebAssembly/design/pull/660. I expect to work on this at some point, but I don't know when I'll find sufficient time. I'd be happy to help anyone else interested in working on this in the meantime.

One possible implementation would be to extend `assert_return_nan`. Currently, `assert_return_nan` tests for either negative or positive default NaN. If we made it also accept any nan that appears as an operand, but with the fraction field MSB set to 1, then we could use it both for cases like this where there are no nan operands:

f32.wast:
`(assert_return_nan (invoke "add" (f32.const -infinity) (f32.const infinity))`

_and_ cases like this where there's a `nan` operand and we want to validate that we either get that operand with the fraction field MSB set to 1, or one of the default NaN values:

float_misc.wast:
`(assert_return (invoke "f32.add" (f32.const nan:0x200000) (f32.const 1.0)) (f32.const nan:0x600000))`

rossberg-chromium:
In general, there is no relation between arguments and result, so looking at the arguments would perhaps be a bit weird. Instead, what if `assert_result_nan` simply allowed specifying a list of acceptable NaN values?

sunfishcode:
The list of acceptable NaN values is a function of the argument list, and it's the same function for all arithmetic operators (`neg`, `abs`, `copysign` notwithstanding), so it'd be very redundant to specify them in a separate list on each test. However it's just convenience for testcases, so I'd be ok either way.

rossberg-chromium:
That's the case for _operators_, but not for _functions_, which are what gets invoked in an assert. I can easily write one where the result has nothing whatsoever to do with the arguments. ;)

sunfishcode:
What I've proposed here is indeed a specialized mechanism. While specialized mechanisms typically have fewer uses than general-purpose mechanisms, they can typically offer those uses they do have greater advantages.

Specifically, my proposal would keep a large number of existing tests readable and maintainable, and makes it easier to add tests for new future operators, and makes it easier to tweak the NaN semantics (in case that becomes necessary, which is a very real possibility). The general-purpose mechanism proposed here would not have these advantages, even if it would have other advantages.

sunfishcode:
https://github.com/WebAssembly/design/pull/716 is now merged in the design repo and has updated NaN propagation semantics.

sunfishcode:
[...]

PR #321, opened 2016-08-23
Remove obsolete nan-propagation.wast.

gahaas:
NaN propagation has been removed from the design repo, which made the nan-propagation tests obsolete. this PR removes the nan-propagation tests.


rossberg-chromium:
Does this require changes to the implementation as well? If so, we should perhaps do them together.

In any case, can you please retarget the PR on the binary-0xc branch?

titzer:
The change landed on master, IIUC. If so, then it's probably best to go
ahead and remove these tests on master, too.

rossberg-chromium:
I see. In that case sgtm, but we should merge to binary-0xc after.

@sunfishcode, you probably know best: IIUC the change is purely a relaxation, but is there anything that can/should be simplified in the impl of arithmetics now to reflect it?

sunfishcode:
This change lgtm.

`determine_binary_nan` and `determine_unary_nan` could indeed be simplified, though as you observe, it's not necessary.

I've updated https://github.com/WebAssembly/spec/issues/286 with notes about the remaining missing NaN test coverage.

rossberg-chromium:
@gahaas, feel free to land this. Bonus points if you also make the simplifications that @sunfishcode mentioned (see `float.ml`). :)

PR #322, opened 2016-08-23
Merge natural alignment change to binary-0xc

AndrewScheidecker:
Use the stored size as the natural alignment for memory operations, rather than the size of the value (4 or 8 bytes).


rossberg-chromium:
Awesome, thanks!

PR #324, opened 2016-08-31
Implement import/export of tables, memories, globals

rossberg-chromium:
This patch implements imports/exports as of of WebAssembly/design#682 and removes `call_import`.

This change required a lot of refactoring in the interpreter:
- Imports and exports (obviously) have been completely restructured.
- Index spaces for imports and functions got merged.
- There is a separate module `Instance` representing module instances.
- Exported functions are represented as original AstFunc (paired with an instance), or as untyped HostFunc.
- Module initialisation became quite a bit more complex.

[Update: made export syntax nicer.]
The patch also changes and extends the S-expr syntax of modules significantly, as follows:
```
func_sig:   ( type <var> ) | <param>* <result>?
global_sig: <type>
table_sig:  <nat> <nat>? <elem_type>
memory_sig: <nat> <nat>?

global:  ( global <name>? <global_sig> )
table:   ( table <name>? <table_sig> )
         ( table <name>? <elem_type> ( elem <var>* ) )  ;; = (table <name>? <size> <size> <elem_type>) (elem (i32.const 0) <var>*)
elem:    ( elem <expr> <var>* )
memory:  ( memory <name>? <memory_sig> )
         ( memory <name>? ( data <string>* ) )          ;; = (memory <name>? <size> <size>) (data (i32.const 0) <string>*)
data:    ( data <expr> <string>* )

import:  ( import <string> <string> <imkind> )
imkind:  ( func <name>? ( <func_sig> ) )
         ( global <name>? ( <global_sig> ) )
         ( table <name>? ( <table_sig> ) )
         ( memory <name>? ( <memory_sig> ) )
export:  ( export <string> <exkind> <var> )
exkind:  func | global | table | memory
```
To avoid repetition, there are also convenient shorthands to specify exports inline (and write imports symmetrically):
```
func:    ( func <name>? (export <string>) <func_sig> <local>* <expr>* )
         ( func <name>? (import <string> <string>) <func_sig>)
global:  ( global <name>? ( export <string> ) <global_sig> )
         ( global <name>? ( import <string> <string> ) <global_sig> )
table:   ( table <name>? ( export <string> ) <table_sig> )
         ( table <name>? ( export <string> ) <elem_type> ( elem <var>* ) )
         ( table <name>? ( import <string> <string> ) <table_sig> )
memory:  ( memory <name>? ( export <string> ) <memory_sig> )
         ( memory <name>? ( export <string> ) ( data <string>* ) )
         ( memory <name>? ( import <string> <string> ) <memory_sig> )
```
These  forms generalise the previous inline string name syntax for functions, which I removed -- the new forms is somewhat more verbose but also more searchable and symmetric.

Furthermore, it is now forbidden to put an import _after_ a real definition, in order to avoid confusion about the merged index space.

The "spectest" pseudo module now defines a couple of table/memory/global objects for testing imports. Furthermore, there is a new `assert_unlinkable` command that verifies that linking fails (at module instantiation time).

Unfortunately, that still doesn't give a good way of testing the import/export semantics. I'd like to extend script syntax in a follow-up to allow more meaningful tests with multiple modules, and ways to observe non-function exports.


rossberg-chromium:
I rethought the export syntax and settled for a nicer shorthand. You can see the difference in the test diffs of commit 930b78d (e.g. https://github.com/WebAssembly/spec/pull/324/commits/930b78d7f937fac96fc19a70fa0359da4efe8bc5#diff-a9bee80497844ab6f0450df04af9f8f3).

Original comment updated with new description. Please let me know if you do not like this better. :)

kripken:
New export syntax is a nice improvement I think.

lukewagner:
Yes, the syntax looks nice and symmetric.  I'm afraid I don't have time to read through all the impl changes but feel free to merge when you're ready.  When we update and execute the updated test suite, that will help to identify discrepancies.

AndrewScheidecker:
Could you make `global_sig` distinguish between mutable and immutable globals as part of this change?

rossberg-chromium:
@AndrewScheidecker, that's in PR #331. This one is more than large enough already. But I gonna land them together anyway.

PR #325, opened 2016-09-01
Script support for testing dynamic linking

rossberg-chromium:
This adds script facilities for defining multiple modules, naming them, and _registering_ them. It also refactors the command grammar slightly and adds an "action" to `get` the value of an exported global (analogous to `invoke` for functions).
```
cmd:
  <module>                                  ;; define, validate, and initialize module
  <action>                                  ;; perform action and print results
  ( register <string> <name>? )             ;; register module for imports
  ( assert_return <action> <expr>? )        ;; assert action has expected results
  ( assert_return_nan <action> )            ;; assert action results in NaN
  ( assert_trap <action> <failure> )        ;; assert action traps with given failure string
  ( assert_invalid <module> <failure> )     ;; assert module is invalid with given failure string
  ( assert_unlinkable <module> <failure> )  ;; assert module fails to link module with given failure string
  ( input <string> )                        ;; read script or module from file
  ( output <name>? <string>? )              ;; output module to stout or file

action:
  ( invoke <name>? <string> <expr>* )       ;; invoke function export
  ( get <name>? <string> )                  ;; get global export
```
Module names remain optional. If omitted at use sites, the last defined module is implied as before.

A _registered_ module is available for importing. See `test/linking.wast` for a number of examples.

### DIFF
--- a/ml-proto/README.md
+++ b/ml-proto/README.md
@@ -226,7 +226,8 @@ cmd:
   ( output <string>? )                      ;; output module to stout or file
 
 action:
-  ( invoke <name> <expr>* )                 ;; invoke export
+  ( invoke <name> <expr>* )                 ;; invoke function export
+  ( get <name> )                            ;; get global export
 ```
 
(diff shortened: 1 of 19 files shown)


lukewagner:
Cool, looks useful, lgtm

rossberg-chromium:
Merged into binary-0xc.

PR #329, opened 2016-09-02
A couple of simple 0xc binary changes

rossberg-chromium:
- Numeric section names
- Element section before code section
- END opcode for functions (plus enabled simplification of decoding stream)
- Limit length of LEBs


rossberg-chromium:
Merged into stack manually.

PR #333, opened 2016-09-05
Fixed a spec error in the code section

wibblymat:
The binary encoding of a Function Body should start with the size of the AST, followed by declaring the locals. Instead we were declaring the locals first.

Also made a minor change to the README because `ocamlbuild` is no longer shipped with `ocaml` in Homebrew.

Finally, changed the `runtests.py` script to put quotes around all command arguments. This lets you run the tests from a path that has a space in it.


rossberg-chromium:
LGTM, thanks!

PR #337, opened 2016-09-10
Implement conversion of test scripts to JavaScript

rossberg-chromium:
Now that the interpreter is almost 0xc-complete, this should come in handy.

This patch extends the script language with a couple of meta facilities. The first is quoting of scripts (by wrapping them into `(script <name>? ...)`); this binds a script in the same way a module def binds a module. The second is the extension of the `output` command to `.js` target files, which converts the referenced script to an equivalent JS runner.

More importantly, you can access these facilities with simple command line switches. For example,
```
wasm -d script.wast -o script.js
```
will convert a test script to self-contained JavaScript, with all Wasm modules embedded as binary strings.

Caveat: All `invoke`'s currently become calls directly from JavaScript, no module wrappers. Thus tests involving i64 params/results do not work yet. Will be addressed in a follow-up.

[Edit]

Extended to also support converting whole scripts to modified `.wast` files:
```
wasm -d script.wast -o script2.wast
wasm -d script.wast -o script2.bin.wast
```
When the output file extension is `.bin.wast`, then all embedded modules are represented as binary strings, otherwise textual. The only exception are modules in a newly introduced `assert_malformed` command, which retain their original format, since cross-coding is meaningless in that case.

The test runner is updated to use this functionality.


rossberg-chromium:
Extended to also support cross-converting `.wast` scripts, such that all embedded modules become either textual or binary:
```
wasm -s script.wast -o script2.wast
wasm -s script.wast -o script2.bin.wast
```
Updated original comment.

rossberg-chromium:
Any opposition to landing this and #341 ? Together these patches should enable using ml-proto directly for the same functionality as sexpr2wasm, especially for 0xc tests.

titzer:
lgtm

rossberg-chromium:
Conflicts resolved and merged.

PR #341, opened 2016-09-12
Create wrapper modules for tests that JS can't handle

rossberg-chromium:
Extends the JS conversion to create Wasm wrapper modules for invocations that involve i64.

Hopefully we can get rid of this one day... :)


rossberg-chromium:
Merged into script-convert.

PR #349, opened 2016-09-21
Fix wrapper i64 modules

titzer:
Not sure if the name $$ is always used for imports to the wrapper; I assume so.

### DIFF
--- a/ml-proto/host/js.ml
+++ b/ml-proto/host/js.ml
@@ -197,7 +197,7 @@ let of_string = of_string_with add_char
 let of_wrapper x_opt name wrap_action wrap_assertion at =
   let x = of_var_opt x_opt in
 	let bs = wrap x name wrap_action wrap_assertion at in
-  "instance(" ^ of_bytes bs ^ ", " ^ "{" ^ x ^ "}).exports.run()"
+  "instance(" ^ of_bytes bs ^ ", " ^ "{$$: " ^ x ^ "[\"exports\"]}).exports.run()"
 
 let of_float z =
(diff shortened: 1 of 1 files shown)


ml-proto/host/js.ml
```diff
@@ -197,7 +197,7 @@ let of_string = of_string_with add_char
 let of_wrapper x_opt name wrap_action wrap_assertion at =
   let x = of_var_opt x_opt in
 	let bs = wrap x name wrap_action wrap_assertion at in
-  "instance(" ^ of_bytes bs ^ ", " ^ "{" ^ x ^ "}).exports.run()"
+  "instance(" ^ of_bytes bs ^ ", " ^ "{$$: " ^ x ^ "[\"exports\"]}).exports.run()"
```
rossberg-chromium: The import module's name is x, so this should be "{" ^ x ^ ": " ^ x ^ ".exports} ...."  (no need to quote 'exports').

rossberg-chromium:
Oops, that explains a few things. :) Thanks.

rossberg-chromium:
LGTM modulo suggestion.

ml-proto/host/js.ml
titzer: Ah ha, that's the way I had it originally but I wasn't sure if the import was always called $$ or "x" was the name of the module in the generated imports, too.

rossberg-chromium: $$ is used as a default for unnamed modules (which most in the test suite are, but it would fail for other cases).

PR #350, opened 2016-09-21
Test that a function with >4KB of locals cannot avoid a stack overflow trap

AndrewScheidecker:
I don't know if this is the right place for this, but I made this test as a proof of concept of a vulnerability in WAVM. I doubt the browser implementations are vulnerable to this, but any implementation that doesn't emit stack probes or otherwise check for stack overflow in the function prologue will have the same problem.

On Windows, LLVM will correctly emit stack probes for functions that use more than 4KB of stack space, and WAVM passes this test. On other platforms, LLVM doesn't emit stack probes, and WAVM does not pass this test.


AndrewScheidecker:
Update: made it less likely that the test gets lucky and hits a guard page at the end of the stack.

rossberg-chromium:
LGTM with one nit: can you break those long lines (ideally at 80 cols)?

AndrewScheidecker:
> LGTM with one nit: can you break those long lines (ideally at 80 cols)?

Sure, updated.

Issue #343, opened 2016-09-17
Undropped br_ifs in block.wast test?


kripken:
In `binary_0xc` the block test has

````
  (func (export "break-repeated") (result i32)
    (block i32
      (br 0 (i32.const 18))
      (br 0 (i32.const 19))
      (br_if 0 (i32.const 20) (i32.const 0))
      (br_if 0 (i32.const 20) (i32.const 1))
      (br 0 (i32.const 21))
      (br_table 0 (i32.const 22) (i32.const 4))
      (br_table 0 0 0 (i32.const 23) (i32.const 1))
      (i32.const 21)
    )
  )
````

Since `br_if` now returns its value, those `br_ifs` should normally be dropped, I believe. I guess the reason they aren't dropped here is `br` clearing the stack?

If that's the case, then I'd like to propose moving such tests to `stack.wast`. The motivation is that Binaryen doesn't intend to support such "stacky" code ([as it doesn't fit well with its internal optimization IR](https://github.com/WebAssembly/binaryen/issues/663)). We would like to pass the test suite sans `stack.wast`, basically. (So far I've maintained a list of such code in other tests, to ignore them, but I can only ignore a whole module, and in the case here, that means ignoring a lot of non-stacky parts.) If that sounds ok I can make a PR.

rossberg-chromium:
Yes, that's an oversight. Because they are dead code, and the interpreter does not currently validate dead code (as we allowed), they went unnoticed. Needs fixing.

There are probably a few other cases of invalid dead code lingering in the test suite that were overlooked during some semantics changes (@sunfishcode also found a couple related to blocks). I'm working on reintroducing full type-checking for dead code, which should help find all these.

JSStats:
Is it really invalid? The `br` after the `br_if` does unwind.

rossberg-chromium:
@JSStats, oops, yes, you are right. It's valid.

@kripken, I would prefer just adding the drops to this test for now. We probably need a more thorough discussion about what to do post MVP. I'm fine with keeping the "stacky" tests separate for now, but to be honest, I'm not convinced it's a sustainable strategy long term, once we generalise the stack semantics even further (like e.g. with multiple values or block inputs). Especially since neither the syntax nor the type system makes any such distinction, so there's gonna be zero guidance.

kripken:
@rossberg-chromium: Ok, I'll look into adding drops for this. I see it happen in a bunch more places meanwhile, so I'm not sure how practical it is yet.

[...]

PR #351, opened 2016-09-21
Stack test cleanup

kripken:
As discussed in #343, this adds `drop` to `br_if`s where necessary, and moves some stacky tests into `stack.wast`. With these changes, the "non-stacky" test files don't have stacky stuff, and binaryen can pass them (with just these disabled: stack, nop, typecheck).

### DIFF
--- a/ml-proto/test/block.wast
+++ b/ml-proto/test/block.wast
@@ -198,11 +198,6 @@
 )
 ;)
 
-;; TODO(stack): move this elsewhere
-(module (func $type-break-num-vs-void
-  (block (drop (i32.const 66)) (br 0))
-))
-
 (assert_invalid
   (module (func $type-break-last-void-vs-num (result i32)
     (block i32 (br 0))
--- a/ml-proto/test/block.wast
+++ b/ml-proto/test/block.wast
@@ -75,8 +75,8 @@
     (block i32
       (br 0 (i32.const 18))
       (br 0 (i32.const 19))
-      (br_if 0 (i32.const 20) (i32.const 0))
-      (br_if 0 (i32.const 20) (i32.const 1))
+      (drop (br_if 0 (i32.const 20) (i32.const 0)))
+      (drop (br_if 0 (i32.const 20) (i32.const 1)))
       (br 0 (i32.const 21))
       (br_table 0 (i32.const 22) (i32.const 4))
       (br_table 0 0 0 (i32.const 23) (i32.const 1))
@@ -200,7 +200,7 @@
 
 ;; TODO(stack): move this elsewhere
 (module (func $type-break-num-vs-void
-  (block (i32.const 66) (br 0))
+  (block (drop (i32.const 66)) (br 0))
 ))
 
 (assert_invalid
--- a/ml-proto/test/br.wast
+++ b/ml-proto/test/br.wast
@@ -56,7 +56,7 @@
     (block i32 (br_if 0 (br 0 (i32.const 8)) (i32.const 1)) (i32.const 7))
   )
   (func (export "as-br_if-value-cond") (result i32)
-    (block i32 (br_if 0 (i32.const 6) (br 0 (i32.const 9))) (i32.const 7))
+    (block i32 (drop (br_if 0 (i32.const 6) (br 0 (i32.const 9)))) (i32.const 7))
   )
 
   (func (export "as-br_table-index")
@@ -247,7 +247,7 @@
       (i32.const 1)
       (block i32
         (drop (i32.const 2))
-        (br_if 0 (i32.const 4) (br 0 (i32.const 8)))
+        (drop (br_if 0 (i32.const 4) (br 0 (i32.const 8))))
         (i32.const 16)
       )
     )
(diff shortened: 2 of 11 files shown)


ml-proto/test/block.wast
```diff
@@ -200,7 +200,7 @@
 
 ;; TODO(stack): move this elsewhere
 (module (func $type-break-num-vs-void
-  (block (i32.const 66) (br 0))
+  (block (drop (i32.const 66)) (br 0))
```
JSStats: It's not clear why this code needs changing, it would have worked with the AST without change. Could I ask why this is problematic for binaryen at present?

kripken: In the AST model, a value must be used or dropped. The 66 constant is not used, therefore it must be dropped.

JSStats: At one point at least the AST did support code such as `(i32.add (i32.const 1) (br ...))` and `(block (i32.const 1) (i32.const 2))` and 'statements' are consistent with an AST.

kripken: As I understand the AST, in your first example the 1 is used by the add. The second example was supported before explicit drop was decided, but after explicit drop the first of the two constants must be dropped.

JSStats: It seems ok for the AST apart from some constraint being assumed of 'explicit drop', but this constraint is overly restrictive as this code is valid with 'explicit drop'. I don't know your code, but might it just be a matter of relaxing a constraint, a simply matter to resolve in binaryen? This might not address all these cases, but perhaps a good number of them. If you could point me to the failing check then that might help me understand the challenge?

kripken: > It seems ok for the AST apart from some constraint being assumed of 'explicit drop'

Exactly. But explicit drop is what was decided, for better or for worse.

JSStats: As far as I am aware the 'explicit drop' constraints do not make these examples invalid, it's just binaryen, and we seem to agree that some of these are not incompatible with an AST either. I don't think these spec tests should be changed just to follow a more restrictive interpretation being made in binaryen.

The tests that would be incompatible with the binaryen AST for now are those that use an ordering that would demand `first` or lexical scoped constants. I think binaryen should work with the rest, relax the constraints, and just separate the remainder.

rossberg-chromium: I think you can just remove this test here, it is superseded by the tests in unwind.wast (#344).

ml-proto/test/br.wast
```diff
@@ -56,7 +56,7 @@
     (block i32 (br_if 0 (br 0 (i32.const 8)) (i32.const 1)) (i32.const 7))
   )
   (func (export "as-br_if-value-cond") (result i32)
-    (block i32 (br_if 0 (i32.const 6) (br 0 (i32.const 9))) (i32.const 7))
+    (block i32 (drop (br_if 0 (i32.const 6) (br 0 (i32.const 9)))) (i32.const 7))
```
rossberg-chromium: He he, most of these are probably also fixed in #345.

ml-proto/test/call.wast
```diff
@@ -191,14 +191,6 @@
   "type mismatch"
 )
 
-;; TODO(stack): move these elsewhere
```
rossberg-chromium: Equivalent tests exist in nop.wast, so you can just delete this.

ml-proto/test/call_indirect.wast
```diff
@@ -285,21 +285,6 @@
   "type mismatch"
 )
 
-;; TODO(stack): move these elsewhere
```
rossberg-chromium: Same here, this can just go away.

ml-proto/test/func.wast
```diff
@@ -475,11 +475,6 @@
 )
 ;)
 
-;; TODO(stack): move this elsewhere
```
rossberg-chromium: This one is subsumed by unwind.wast (#344) as well.

ml-proto/test/loop.wast
```diff
@@ -280,10 +280,3 @@
   "type mismatch"
 )
 
-;; TODO(stack): move these elsewhere
```
rossberg-chromium: Same here.

ml-proto/test/br_if.wast
```diff
@@ -14,10 +14,10 @@
     (block (call $dummy) (call $dummy) (br_if 0 (get_local 0)))
   )
   (func (export "as-block-first-value") (param i32) (result i32)
-    (block i32 (br_if 0 (i32.const 10) (get_local 0)) (return (i32.const 11)))
+    (block i32 (drop (br_if 0 (i32.const 10) (get_local 0))) (return (i32.const 11)))
```
titzer: I don't understand why a drop is required here, since br_if leaves its value on the stack in the false case, and then falls off the end of the block, and the block is annotated as i32.

kripken: The `return` is also in that block, i.e. with indentation it looks like this:

````
(block $block0 i32
  (br_if $block0
    (i32.const 10)
    (get_local $0)
  )
  (return
    (i32.const 11)
  )
)
````

kripken:
@rossberg-chromium: ok, I pushed an update addressing your comments. also some fixes i noticed when running the tests.

rossberg-chromium:
LGTM

PR #358, opened 2016-10-11
Remove validation of segment bounds

rossberg-chromium:
Eliminate the validation constraint that segments have to be ordered and within bounds. With the new import/export semantics, these checks are no longer possible in general, because segment offsets can depend on imports (in which case the interpreter previously produced a "crash" exception).

Bounds checks now happen at link time. Tests adapted and extended accordingly.


rossberg-chromium:
@lukewagner, are you okay with this change?

titzer:
lgtm

lukewagner:
lgtm, thanks!

PR #360, opened 2016-10-13
Test traps from start function; check JS Error classes

rossberg-chromium:
Noticed that there was no test for traps from module instantiation. Added a new form of `assert_trap` taking a module definition.

Also refined JS translation to check that the correct exceptions are thrown.


gahaas:
lgtm

PR #361, opened 2016-10-14
Add required files for OPAM + ocamlfind distribution

msprotz:
@rossberg-chromium 

This puts packages under the "Wasmlib" module... so, there would be "Wasmlib.Encode", etc. 

This is mostly to discuss the style, conventions, etc. -- I don't except this to be accepted immediately.

Here's what happens after you run "make install":

```
jonathan@chartreuse:~/Code/kremlin (wasm) $ ocaml
        OCaml version 4.03.0

# #require "topfind";;
Unknown directive `require'.
# #use "topfind";;
- : unit = ()
Findlib has been successfully loaded. Additional directives:
  #require "package";;      to load a package
  #list;;                   to list the available packages
  #camlp4o;;                to load camlp4 (standard syntax)
  #camlp4r;;                to load camlp4 (revised syntax)
  #predicates "p,q,...";;   to set these predicates
  Topfind.reset();;         to force that packages will be reloaded
  #thread;;                 to enable threads

- : unit = ()
# #require "wasmlib";;
/Users/jonathan/.opam/4.03.0/lib/ocaml/unix.cma: loaded
/Users/jonathan/.opam/4.03.0/lib/ocaml/bigarray.cma: loaded
/Users/jonathan/.opam/4.03.0/lib/ocaml/str.cma: loaded
/Users/jonathan/.opam/4.03.0/lib/wasmlib: added to search path
/Users/jonathan/.opam/4.03.0/lib/wasmlib/wasmlib.cmo: loaded
# #show Wasmlib.Encode;;
module Encode :
  sig
    exception Code of Wasmlib.Source.region * string
    val version : int32
    val encode : Wasmlib.Ast.module_ -> string
  end
```

After this is accepted, I plan on submitted a corresponding OPAM package request. This should be trivial, since I already wrote the `opam` file at the root of the repository, meaning you can already do `opam pin https://github.com/msprotz/spec && opam install wasmlib`.


rossberg-chromium:
Thanks for taking this over -- I was looking at OPAM the other day, but didn't get very far...

The only concern I have is that this adds more files to top directories. In particular, the `opam` file in the root directory is a bit out of place, since only `ml-proto` is an OCaml project in this repo (and we will soon add non-OCaml stuff).

Is it possible to somehow move most of the extra files to (ideally, a subdirectory of) `ml-proto`? Bonus points if we can also hide away `_tags`. :)

msprotz:
No worries. I've removed `opam` -- this file is meant to go into the OPAM repository anyhow, I just put it there so that I would have an easy way to install this into a student's OPAM.

I've moved META to `findlib/` and I also install `mli`'s now in the `ocamlfind` package (it's good practice).

I don't know of any way off the top of my head to store `_tags` elsewhere...

msprotz:
Also, when I add this into the `opam-repository`, I will most certainly be asked to point to a "release" .tar.gz instead of a git repository. Do you have a snapshot somewhere?

rossberg-chromium:
Thanks! Zipfile & Tarfile download is actually built into github (but like everything else with github's interface, hard to find and badly documented). Just use this link:

https://github.com/WebAssembly/spec/archive/master.tar.gz

One more question: Do you think we could auto-generate the mlpack file? E.g., we could tweak the directory structure such that it would simply include all modules from certain directories. That would avoid redundant manual bookkeeping.

Also, it'd be great if you could add some brief section about the package to the README.

rossberg-chromium:
Oh, and is there a specific convention suggesting to call the packaged module Wasmlib instead of just Wasm?

msprotz:
Yeah but the OPAM folks object to having me point to the master release, because your master may break at a later point in the future, and this will make their CI red for no reason. So I guess any relatively stable tag would be good, just to have a "fixed" release; OPAM has a "dev-repo" feature that allows users to switch to the master version if they wish to, but that doesn't break OPAM's CI.

I'll try write some unix magic in the Makefile to auto-generate the pack structure... something along the lines of `find | basename | sort | uniq`, but then this makes the Makefile less portable (OSX variants, i.e. gfind vs find).

If you prefer Wasm instead of Wasmlib, no objection on my side.

msprotz:
For the record it seems hard to auto-generate the pack, because some files in the repository do not seem to compile (unused?).

```
jonathan@chartreuse:~/Code/webasm-spec/ml-proto (master) $ find given host spec -iname '*.ml*' | xargs -I % basename % .ml | xargs -I % basename % .mli | sort | uniq | grep -v .mly > wasmlib.mlpack 
jonathan@chartreuse:~/Code/webasm-spec/ml-proto (master) $ make opt unopt
ocamlbuild    -I given -I host -I host/import -I spec -libs str -libs bigarray -quiet main.native
mv main.native wasm.opt
ocamlbuild    -I given -I host -I host/import -I spec -libs str -libs bigarray -quiet wasmlib.cmx
+ /Users/jonathan/.opam/4.03.0/bin/ocamlc.opt -c -bin-annot -I host -I spec -I given -I host/import -o host/prettyprint.cmo host/prettyprint.ml
File "host/prettyprint.ml", line 22, characters 4-14:
Error: Unbound constructor MemorySize
Command exited with code 2.
make: *** [_build/wasmlib.cmx] Error 10
```

rossberg-chromium:
No problem, we can create some kind of release branch (although we actually try to treat master as our release branch, and do forward dev in branches like binary-0xc).

Would you actually need `find` for the make target? I'm perfectly happy to tweak directory structure such that `ls` on a fixed set of dirs is enough.

Oops, `prettyprint` was a temporary debug thing, I must have added it accidentally at some point. Removed.

msprotz:
The mlpack file is now auto-generated... and the Makefile seems right.

Yes, a release branch, or even a tag, would be enough.

Concerning the use of find, well... it's just convenient and it's in the intersection of BSD find and GNU find, so it doesn't seem like too much of a burden. If the directory structure were flat, then I could just use ls indeed, e.g. `ls {spec,host,given}/*.ml{,i}`.

Or I could just include module that have an *.ml... not sure what is the best way to get everything into the pack, but the `find` I have right now seems ok.

msprotz:
Ok, this is ready to merge -- I've done a big cleanup and various tests, as always, there were some unexpected difficulties, but packaging should be ok now.

rossberg-chromium:
Cool, thanks! One question: regarding the name I actually wondered more about the module name than the package name. Would plain `Wasm` be inappropriate?

msprotz:
The ocamlfind package and the opam package are now called wasm.

For the name of the pack itself, there's already the executable named wasm, so I'm not sure how to build a pack of the same name... :-/. Do you have any ideas?

The command that I wrote to auto-generate the wasmlib.mlpack file doesn't work on Linux because of a difference between GNU basename and BSD basename... do you have a Linux machine at hand to try it out? Otherwise, we can put the mlpack file under version control...

msprotz:
Auto-generating the wasmlib.mlpack file doesn't seem like a great idea, because now in addition to writing a cross-platform shell line, I need to remove the module that starts the interpreter... i.e. the main module.

I've managed to rename wasmlib into wasm

msprotz:
Everything now seems good except that my shell pattern in the Makefile is not expanded on Linux (the `_build/wasm.{o,cmo...}`)

msprotz:
Confirmed that everything now install fine on a linux box.

rossberg-chromium:
Great, thanks! Still seems to fail on travis, though -- the environment doesn't seem to recognise `basename -a`.

Do you also want to add something about the package to the README?

msprotz:
I don't know... the man page for GNU basename on my cygwin setup claims that -a is supported...

msprotz:
and my colleague's linux machine also has basename -a that works...

rossberg-chromium:
Hm, bummer. But perhaps something more braindead works? E.g., the winmake target just pipes through a couple of greps and seds.

rossberg-chromium:
Cool! Note that I landed some renamings yesterday that you will need to adapt to to resolve the conflicts. In particular, the whole `ml-proto/` dir is now called `interpreter/`. And you probably want to use the Makefile's `$DIRS` variable in the mlpack target to make it agnostic to other directory changes.

rossberg-chromium:
Ah, I was too slow. :)

msprotz:
Hehe yes. I can't believe it took me so long to get it right, but there we are!

rossberg-chromium:
LGTM. Thanks for the work and patience.

msprotz:
No worries. Let me know once you have a stable branch / tag I can point my OPAM package to, and I'll finish the work on the OPAM side.

rossberg-chromium:
Sure, timing is perfect. We have just finalised binary version 0xD in the design repo, which primarily defrags the opcode space, and is intended to be feature-complete wrt Wasm 1.0. Once I have adapted the interpreter (next week), I'll create some kind of release branch.

rossberg-chromium:
Okay, I created branch `pre-release-0xd`, which reflects the latest version 0xD.

PR #369, opened 2016-10-27
Soft fail loop tests

bnjbvr:
Two commits:
- first one to update `.gitignore`, following a name change somewhere (?).
- second one: I've checked with @sunfishcode, and it seems those two tests from `loop.wast` are part of the unreachable test cases, which should only softly fail. The rationale, in both cases, is that a `br` to a loop top label is always void, and the function's end is unreachable, since we have infinite loops; so per spec, the validation error is a soft fail. Does it make sense? (SM doesn't issue validation errors for these two test cases, according to this reasoning)

### DIFF
--- a/interpreter/.gitignore
+++ b/interpreter/.gitignore
@@ -2,7 +2,7 @@
 *.byte
 *.opt
 *.zip
-wasmlib.mlpack
+wasm.mlpack
 _build
 wasm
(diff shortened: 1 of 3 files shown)


rossberg-chromium:
Yes, makes sense. LGTM.

PR #371, opened 2016-10-28
Reduce the requested amount of pages in the biggest grow test case

bnjbvr:
Really a dumb request: we have these tests checked in in automation in SM, and among all the test machines they run on VMs which have less than 512 MB of memory. One of the resizing tests requests 10K pages of memory which is 625 MB of memory, so this consistently triggers out-of-memory (grow returning -1) in our infra. While we could tweak the test infra to handle this as an expected failures, it sounds like it defeats the purpose of the test itself, and we could just lower the amount of pages requested maybe? 800 pages is 50 MB, which is still a high number and works in our infra. I think the test was requesting a lot of memory in general; not sure it was testing for values above an certain implicit threshold.


rossberg-chromium:
Sounds good to me.

Issue #370, opened 2016-10-27
Interpreter/aux is an invalid name on Windows


Cellule:
Currently it is impossible to clone the repository on Windows because it is not allowed to name a folder aux.

rossberg-chromium:
Ouch, sorry about that! Renamed it to util/.

PR #375, opened 2016-11-02
Rename aux/ to util/

rossberg-chromium:
Addresses #370.

Also some clean-ups to runtests.py and fixes to README.


interpreter/_tags
```diff
@@ -1,2 +1,2 @@
 true:       bin_annot
-<{aux,host,spec,text,host/import}/*.cmx>:    for-pack(Wasm)
+<{util,host,spec,text,host/import}/*.cmx>:    for-pack(Wasm)
```
gahaas: util should come after text I think, to preserve the alphabetic ordering.

gahaas:
lgtm

interpreter/_tags
rossberg-chromium: Fixed to mirror import dep order.

PR #377, opened 2016-11-03
Parameterise runtests script

rossberg-chromium:
Usage is now
```
runtests.py [--wasm <wasm-command>] [--js <js-command>] [file ...]
```
which allows selecting tests or setting the JS command from the command line, e.g.
```
runtests.py --js 'd8 --expose-wasm' test/address.wast
```

### DIFF
--- a/interpreter/runtests.py
+++ b/interpreter/runtests.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
 import os
 import os.path
 import unittest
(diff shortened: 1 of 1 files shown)


gahaas:
lgtm

interpreter/runtests.py
```diff
@@ -1,109 +1,105 @@
...
-    try:
-      os.remove(path)
-    except OSError:
-      pass
-    return path
 
-class RunTests(unittest.TestCase):
-  def _runCommand(self, command, logPath = None, expectedExitCode = 0):
-    out = None if logPath is None else file(logPath, 'w+')
+def usage():
```
binji: why not use argparse?

rossberg-chromium: Because I am no Python pro. :)

PR #381, opened 2016-11-10
Remove exact nan-pattern tests

gahaas:
This PR removes exact some remaining exact nan-pattern tests because they became obsolete in  WebAssembly/design#716. Similar tests without bit pattern matching exist in f32.wast and f64.wast.


rossberg-chromium:
LGTM, but Dan should sign off as well.

gahaas:
@sunfishcode Could you please take a look?

sunfishcode:
lgtm. As I commented in https://github.com/WebAssembly/spec/issues/286, these are important cases to cover, but for now we lack a way to effectively test them.

lukewagner:
@bnjbvr Did we test these using the magic test mode so that the engine spits out the literal bit contents?

bnjbvr:
@lukewagner Yes indeed; any spec test that uses equality with a precise NaN pattern checks against that same pattern. All the spec tests are run under the magic test mode, in Spidermonkey CI.

PR #382, opened 2016-11-10
Represent alignment in log2 in AST

rossberg-chromium:
To match the binary format. Ill-formed alignment is now a syntax error instead of a validation error, because it cannot be represented to binary.


gahaas:
lgtm

PR #390, opened 2016-11-18
wasm.cmx is in _build not in current directory

kyoDralliam:
A slight change seems to be needed in the makefile.


rossberg-chromium:
Ah, thanks! LGTM

PR #392, opened 2016-11-24
Intro with design goals

rossberg-chromium:
Adds intro that mainly summarises the design goals. Leaves out meta topics like design process and future features which don't belong into the spec itself.

### DIFF
--- a/document/introduction/index.rst
+++ b/document/introduction/index.rst
@@ -48,8 +48,6 @@ The design goals of WebAssembly are the following:
 
   * **Portable**: makes no architectural assumptions that are not broadly supported across modern hardware.
 
-  * **Versatile**: can be targetted equally by both state-of-the-art optimizing compilers and single-pass baseline compilers.
-
 WebAssembly code is also intended to be easy to display and debug, especially in environments like web browsers, but such features are beyond the scope of this specification.
 
 
--- a/document/introduction/index.rst
+++ b/document/introduction/index.rst
@@ -34,7 +34,7 @@ The design goals of WebAssembly are the following:
 
   * **Open**: programs can interoperate with their environment in a simple and universal manner.
 
-* Efficient, checkable, and portable *representation*:
+* Efficient and portable *representation*:
 
   * **Compact**: a binary format that is fast to transmit by being smaller than typical text or native code formats.
 
--- a/document/introduction/index.rst
+++ b/document/introduction/index.rst
@@ -3,7 +3,7 @@ Introduction
 
 WebAssembly (abbreviated Wasm [#wasm]_) is a *safe, portable, low-level code format*
 designed for efficient execution and compact representation.
-Its main goal is to enable high performance applications on the Web, but it does not bake in any Web-specific assumptions or features, so can be employed in other environments as well.
+Its main goal is to enable high performance applications on the Web, but it does not make any Web-specific assumptions or provide Web-specific features, so can be employed in other environments as well.
 
 WebAssembly is an open standard developed by a `W3C Community Group <https://www.w3.org/community/webassembly/>`_ that includes representatives of all major browser vendors.
 
--- a/document/introduction/index.rst
+++ b/document/introduction/index.rst
@@ -22,7 +22,7 @@ The design goals of WebAssembly are the following:
 
   * **Fast**: executes with near native code performance, taking advantage of capabilities common to all contemporary hardware.
 
-  * **Safe**: code is validated and executes in a memory-safe, sandboxed environment preventing data corruption or security breaches.
+  * **Safe**: code is validated and executes in a memory-safe [#memorysafe]_, sandboxed environment preventing data corruption or security breaches.
 
   * **Well-defined**: fully and precisely defines valid programs and their behavior in a way that is easy to reason about informally and formally.
 
@@ -52,3 +52,5 @@ WebAssembly code is also intended to be easy to inspect and debug, especially in
 
 
 .. [#wasm] A contraction of "WebAssembly", not an acronym, hence not using all-caps.
+
+.. [#memorysafe] No program can break WebAssembly's memory model. Of course, it cannot guarantee that an unsafe language compiling to WebAssembly does not corrupt its own memory layout, e.g. inside WebAssembly's linear memory.
--- a/document/introduction/index.rst
+++ b/document/introduction/index.rst
@@ -24,23 +24,23 @@ The design goals of WebAssembly are the following:
 
   * **Safe**: code is validated and executes in a memory-safe, sandboxed environment preventing data corruption or security breaches.
 
-  * **Deterministic**: fully and precisely defines valid programs and their behavior in a way that is easy to reason about informally and formally.
+  * **Well-defined**: fully and precisely defines valid programs and their behavior in a way that is easy to reason about informally and formally.
 
-  * **Hardware-independent**: can run on all modern hardware and CPUs, desktop or mobile devices and embedded systems alike.
+  * **Hardware-independent**: can be compiled on all modern architectures, desktop or mobile devices and embedded systems alike.
 
-  * **Language-independent**: does not privilege nor penalize any particular programming or object model.
+  * **Language-independent**: does not privilege any particular language, programming model, or object model.
 
   * **Platform-independent**: can be embedded in browsers, run as a stand-alone VM, or integrated in other environments.
 
   * **Open**: programs can interoperate with their environment in a simple and universal manner.
 
-* Efficient, safe, and portable *representation*:
+* Efficient, checkable, and portable *representation*:
 
   * **Compact**: a binary format that is fast to transmit by being smaller than typical text or native code formats.
 
   * **Modular**: programs can be split up in smaller parts that can be transmitted, cached, and consumed separately.
 
-  * **Efficient**: can be decoded, validated, and compiled to native code in a fast single pass, equally with either just-in-time (JIT) or ahead-of-time (AOT) compilation.
+  * **Efficient**: can be decoded, validated, and compiled in a fast single pass, equally with either just-in-time (JIT) or ahead-of-time (AOT) compilation.
 
   * **Streamable**: allows decoding, validation, and compilation to begin as soon as possible, before all data has been seen.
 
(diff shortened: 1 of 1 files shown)


document/introduction/index.rst
```diff
@@ -1,6 +1,56 @@
...
+
+  * **Efficient**: can be decoded, validated, and compiled to native code in a fast single pass, equally with either just-in-time (JIT) or ahead-of-time (AOT) compilation.
+
+  * **Streamable**: allows decoding, validation, and compilation to begin as soon as possible, before all data has been seen.
+
+  * **Parallelizable**: allows decoding, validation, and compilation to be split into many independent parallel tasks.
+
+  * **Portable**: makes no architectural assumptions that are not broadly supported across modern hardware.
+
+  * **Versatile**: can be targetted equally by both state-of-the-art optimizing compilers and single-pass baseline compilers.
```
lukewagner: "Targetted" sounds like you're talking about single-pass baseline compilers *producing* wasm.  But looking back at the short history of wasm and the many large and small changes we've made in part motivated by single-pass baseline compiler *consumers*, I would think you mean to refer to that instead.  If so, perhaps you could clarify?  (Otherwise, this sounds like a new high-level goal that has not been heretofore officially recognized (e.g., in HighLevelGoals.md) and so deserves a proper discussion in separate issue.)

rossberg-chromium: Fair enough. It has come up as an implicit assumption repeatedly in discussions, but I agree that we haven't explicitly listed it. Removed.

document/introduction/index.rst
```diff
@@ -1,6 +1,54 @@
 Introduction
 ============
 
-.. todo::
+WebAssembly (abbreviated Wasm [#wasm]_) is a *safe, portable, low-level code format*
+designed for efficient execution and compact representation.
+Its main goal is to enable high performance applications on the Web, but it does not bake in any Web-specific assumptions or features, so can be employed in other environments as well.
```
titzer: bake in is a bit informal, so maybe just "make Web-specific assumptions or require Web-specific features".

I won't bikeshed (teehee) the actual text here, but the points lgtm

titzer:
This LGTM modulo a little wordsmithing which we can do incrementally once we get rolling.

document/introduction/index.rst
rossberg-chromium: Done.

document/introduction/index.rst
```diff
@@ -1,6 +1,54 @@
...
+
+  * **Efficient**: can be decoded, validated, and compiled to native code in a fast single pass, equally with either just-in-time (JIT) or ahead-of-time (AOT) compilation.
+
+  * **Streamable**: allows decoding, validation, and compilation to begin as soon as possible, before all data has been seen.
+
+  * **Parallelizable**: allows decoding, validation, and compilation to be split into many independent parallel tasks.
+
+  * **Portable**: makes no architectural assumptions that are not broadly supported across modern hardware.
+
+WebAssembly code is also intended to be easy to display and debug, especially in environments like web browsers, but such features are beyond the scope of this specification.
```
lukewagner: I was implicitly assuming this spec would end up defining the "official" (linear) text format; are you thinking that too would be in a separate document?

document/introduction/index.rst
```diff
@@ -1,6 +1,54 @@
...
+
+  * **Fast**: executes with near native code performance, taking advantage of capabilities common to all contemporary hardware.
+
+  * **Safe**: code is validated and executes in a memory-safe, sandboxed environment preventing data corruption or security breaches.
+
+  * **Deterministic**: fully and precisely defines valid programs and their behavior in a way that is easy to reason about informally and formally.
+
+  * **Hardware-independent**: can run on all modern hardware and CPUs, desktop or mobile devices and embedded systems alike.
+
+  * **Language-independent**: does not privilege nor penalize any particular programming or object model.
```
sunfishcode: To be clear, I intend this comment as feedback on how this bullet item reads to someone with my perspective. If I'm mistaken about what this wording is intending to say, I'd appreciate being corrected and, if possible, updated wording.

document/introduction/index.rst
```diff
@@ -1,6 +1,54 @@
...
+
+  * **Platform-independent**: can be embedded in browsers, run as a stand-alone VM, or integrated in other environments.
+
+  * **Open**: programs can interoperate with their environment in a simple and universal manner.
```
rossberg-chromium: I agree that "open" is a bit overloaded, but it's commonly used it in the sense of an "open architecture". Not sure "connectable" has the same connotation. Maybe we can find something better? (I absolutely want to avoid saying "interoperable" here.)

document/introduction/index.rst
```diff
@@ -1,6 +1,54 @@
...
+
+* Efficient, safe, and portable *representation*:
```
rossberg-chromium: Removed. (I intended a brief term summarising that it's well-defined and cannot be used as a Trojan horse like some other file formats, but then there is no bullet for it, so better lose it).

document/introduction/index.rst
```diff
@@ -1,6 +1,54 @@
...
+
+  * **Compact**: a binary format that is fast to transmit by being smaller than typical text or native code formats.
+
+  * **Modular**: programs can be split up in smaller parts that can be transmitted, cached, and consumed separately.
+
+  * **Efficient**: can be decoded, validated, and compiled to native code in a fast single pass, equally with either just-in-time (JIT) or ahead-of-time (AOT) compilation.
```
rossberg-chromium: Deleted the "to native code". The "compile" needs to be in there, because the latter half of the sentence refers to it.

document/introduction/index.rst
lukewagner: Makes sense, lgtm

document/introduction/index.rst
```diff
@@ -1,6 +1,54 @@
...
+
+.. index:: design goals, portability
+
+The design goals of WebAssembly are the following:
+
+* Fast, safe, and portable *semantics*:
+
+  * **Fast**: executes with near native code performance, taking advantage of capabilities common to all contemporary hardware.
+
+  * **Safe**: code is validated and executes in a memory-safe, sandboxed environment preventing data corruption or security breaches.
```
sunfishcode: The idiom of using linear memory as an address space is fundamentally "unsafe" because there's no bounds checking for arrays allocated within the address space, or protections for "deallocated" memory (in a malloc+free sense). At one level, this isn't WebAssembly's problem, however this use case is the main reason why WebAssembly has linear memory, so it's a little unfriendly to ignore it. Corrupted linear memory can participate in some forms of security breaches, even though it's sandboxed from the underlying browser/OS.

It's obviously a matter of perspective. From my perspective, just saying "sandboxed" is sufficient and avoids confusion.

rossberg-chromium: Wasm programs cannot access memory in unspecified ways, and even the attempt of doing so (for linear memories) does not lead to undefined behaviour or crashes. Memory safety is the established term for that, so I'm not sure what the confusion would be. It is what necessitates validation, bounds checks, checked calls, etc. Not mentioning it here would completely miss the motivation for all of these. It's different from sandboxing, which is about controlling access to host resources and is much more coarse-grained.

jfbastien: I believe @sunfishcode is making the distinction between:

1. wasm protecting the user from malicious executables
2. wasm protecting developer code running within wasm from bugs in that code

When Rust says it's "memory safe" it means: write code in Rust, and bugs won't bit you too bad. That's 2. above. Run Rust inside wasm and you have these properties. Run C++ inside wasm and you don't.

Either way the user is still safe from the code affecting them. Imagine the code running inside wasm is some banking thing, and it has a bug which gets exploited. Is the user still "safe"? The code can't affect the rest of the machine, but their bank account may be in trouble.

Saying "memory safe" is misleading in that sense, because wasm is meant to host programs which may themselves *not* be safe. It's misleading because someone reading this document could be approaching it from perspective 1. or 2. The current writeup caters to 1., and people coming from 2. are misled.

This is a rehash of https://github.com/WebAssembly/design/issues/205

rossberg-chromium:
> Saying "memory safe" is misleading in that sense, because wasm is meant
to host programs which may themselves *not* be safe. It's misleading

So would it be misleading to call JavaScript memory-safe, because it can
host C programs through asm.js? Then no language could be called
memory-safe.

Let's not mix up levels. An unsafe language can always be compiled into a
safe language. Still, no program compiled into Wasm can break _Wasm's_
memory model. That's an absolutely crucial property. Of course, a program
in some source language can break _that language's_ memory model. But that
is neither a property nor a concern of WebAssembly, and completely outside
this spec. We're talking about Wasm here, not some other language.

Note also that Wasm's memory safety is exactly the reasons why a compiler
from an unsafe language like C has to do extra work such as maintaining a
shadow stack or building a function table, in order to map its unsafe
constructs into the safe environment of Wasm. Without stating this design
goal people would have to wonder why.

wllang:
@rossberg-chromium They seem to just be saying that "memory safe" is a loaded term, not claiming the matter in misunderstood, so it just needs re-wording. Perhaps you could replace it with a note that linear memory accesses are sandboxed to the extent of the linear memory supplied to the wasm code.

titzer:
I think the confusion arises because "memory safe" in this context does not refer to the linear memory, but to the "meta" memory in which the WASM engine is running. In that sense, yes WASM programs are memory safe in that the cannot access or alter the WASM memory; they can only access their own data structures, one of which is confusingly called the linear memory.

@rossberg-chromium's point is actually that WASM is memory safe in the same way that Java and C# are; programs cannot access the memory in which the virtual machine is implemented; they can only access their own data structures, which in their case are objects and arrays.

sunfishcode:
Linear memory in WebAssembly is specifically intended to be used as an address space, and specifically for C/C++. This is quite different from the relationship between JS and asm.js. Calling it "memory-safe"  and that "data corruption or security breaches" are "prevented" risks confusing people.

jfbastien:
> Let's not mix up levels. An unsafe language can always be compiled into a safe language. Still, no program compiled into Wasm can break _Wasm's_ memory model. That's an absolutely crucial property. Of course, a program in some source language can break _that language's_ memory model. But that is neither a property nor a concern of WebAssembly, and completely outside this spec. We're talking about Wasm here, not some other language.

We agree 100%.

I (and IIUC @sunfishcode) are asking you to state this explicitly. wasm is meant to run other programs, people come to it to run other programs, of course some of these people will be confused and mix levels up. Let's not be obtuse and mislead them because they should have known better.

This is the intro, where scope is defined. Let's.

rossberg-chromium:
@jfbastien, I added a footnote explaining what it means.

PR #394, opened 2016-12-07
winmake.bat now produces wasm.exe to be used on Windows

Cellule:
I was renaming the produced binary to `wasm.exe` manually for a while and I figured I should open a PR to change it directly in the source


rossberg-chromium:
Ah, thanks, this is a good fix, but note the comment at the beginning: this file is auto-generated. To make the change permanent you'll need to adjust the Makefile, specifically, change line 79 to
```
		| sed s:host/main.d.byte:%NAME%.exe: \
```

Cellule:
Oh right, I didn't realize the Makefile generated that file

rossberg-chromium:
Thanks, LGTM.

Issue #393, opened 2016-12-07
Binary Tests


wanderer:
It might be nice to have some binary tests. Binaries have a few things that can't be expressed in s-expressions such as a custom section.

rossberg-chromium:
Yes. We had deferred them so far because the binary format was still undergoing many changes. But now might be a good time.

Note that the S-expression format _can_ express binaries, see binary.wast, which might be a good starting place for adding at least basic stuff. Contributions welcome. :)

rossberg-chromium:
Seems like this has been addressed, even though we should have more such tests.

PR #397, opened 2016-12-10
added custom section tests

wanderer:
related #393


rossberg-chromium:
Thanks, LGTM.

While at it, how about a few additional tests, especially that the custom section can appear between arbitrary other sections? :)

wanderer:
@rossberg-chromium ok i will add a few more

PR #398, opened 2016-12-12
Add test for start function being an import

rossberg-chromium:
Cf. WebAssembly/design#896.

### DIFF
--- a/interpreter/test/start.wast
+++ b/interpreter/test/start.wast
@@ -2,6 +2,7 @@
   (module (func) (start 1))
   "unknown function"
 )
+
 (assert_invalid
   (module
     (func $main (result i32) (return (i32.const 0)))
@@ -16,6 +17,7 @@
   )
   "start function"
 )
+
 (module
   (memory (data "A"))
   (func $inc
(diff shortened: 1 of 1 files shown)


titzer:
lgtm

interpreter/test/start.wast
```diff
@@ -87,6 +89,11 @@
   (start $main)
 )
 
+(module
+  (func $print (import "spectest" "print"))
+  (start $print)
+)
+
```
rossberg-chromium: No, at least on the Web, importing a host function never fails, because they have no types and all JS functions are variadic. The "print" function in "spectest" simulates that behaviour. There already are other tests exercising it.

jfbastien: OK gotcha.

My point still stands though: we need to have a test where an import's signature is declared as something else than `void ()`. That test needs to fail.

rossberg-chromium: We have a whole range of such tests, see imports.wast.

rossberg-chromium: Ah, or did you mean a test where a non-void function is used for start? That exists as well, somewhere in this file.

There is no test that uses an imported function for that, mind you, but the type-checking of the start function should be orthogonal to where the function is coming from. But I'm happy to LGTM it if you want to add another test for that. :)

PR #399, opened 2016-12-12
Check all segment bounds beforehand

rossberg-chromium:
...and add tests that memories/tables remain unchanged in case of error. Addresses WebAssembly/design#897

### DIFF
--- a/interpreter/spec/eval.ml
+++ b/interpreter/spec/eval.ml
@@ -359,19 +359,23 @@ let init_closure (inst : instance) (clos : closure) =
   | _ -> assert false
 
 let init_table (inst : instance) (seg : table_segment) =
-  let {index; offset; init} = seg.it in
+  let {index; offset = const; init} = seg.it in
   let tab = table inst index in
-  let offset = i32 (eval_const inst offset) offset.at in
-  try Table.blit tab offset (List.map (fun x -> Func (func inst x)) init)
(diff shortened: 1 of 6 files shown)


jfbastien:
I would rather check bounds as we apply segments, instead of up front. IMO it's not a useful property to purposefully apply segments and *rely* on halfway failure, but it's faster to check each bounds while applying them versus doing two passes. I'd therefore do a single pass, and fail halfway. This optimizes for the useful case (everything succeeds) and leaves the useless case in a well-specified state.

rossberg-chromium:
@jfbastien, I agree that's simpler, but if you prefer that semantics then I suggest you create a PR to change JS.md.

jfbastien:
Done: https://github.com/WebAssembly/design/pull/902

gahaas:
lgtm

rossberg-chromium:
@jfbastien, are you fine with landing this for the time being? Until we have consensus for changing the design the spec should better match the current design.

jfbastien:
@rossberg-chromium with that rationale we could as well merge https://github.com/WebAssembly/design/pull/902, no? I'm OK revisiting later, so feel free to merge as long as we discuss this.

rossberg-chromium:
@jfbastien, thanks. There doesn't seem to be consensus on the broader change from #902, but the overriding goal is to get design and spec into sync. We can always adjust the spec when #902 evolves.

PR #402, opened 2016-12-16
Malformed tests for mutable encoding

Cellule:
Add tests for invalid global mutable encoding value

### DIFF
--- a/interpreter/test/globals.wast
+++ b/interpreter/test/globals.wast
@@ -95,3 +95,15 @@
   (module (global i32 (get_global 1)) (global i32 (i32.const 0)))
   "unknown global"
 )
+
+(module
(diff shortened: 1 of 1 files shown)


interpreter/test/globals.wast
```diff
@@ -95,3 +95,15 @@
   (module (global i32 (get_global 1)) (global i32 (i32.const 0)))
   "unknown global"
 )
+
+;;(module
```
rossberg-chromium: No, I think it's good to have. But in this case, why not uncomment it? Can't harm.

Cellule:
I don't have merge rights on this repo, can someone merge this for me please?

Cellule:
Thanks

PR #403, opened 2016-12-16
Allows to run test on Windows

Cellule:
- Use `wasm.exe` instead of `./wasm` to run tests on Windows
- wasm.exe doesn't recognize paths with `\\`, which is the default separator on Windows
- Windows doesn't recognize arguments wrapped using `'argument'`, it needs to use `"argument"`

### DIFF
--- a/interpreter/runtests.py
+++ b/interpreter/runtests.py
@@ -15,7 +15,7 @@ expectDir = "expected-output"
 outputDir = os.path.join(inputDir, "output")
 
 parser = argparse.ArgumentParser()
-parser.add_argument('--wasm', metavar='<wasm-command>', default='./wasm')
+parser.add_argument('--wasm', metavar='<wasm-command>', default=os.path.join(os.getcwd(), 'wasm'))
 parser.add_argument('--js', metavar='<js-command>')
 parser.add_argument('--out', metavar='<out-dir>', default=outputDir)
 parser.add_argument('file', nargs='*')
@@ -55,7 +55,7 @@ class RunTests(unittest.TestCase):
     # Run original file
     expectedExitCode = 1 if ".fail." in inputFile else 0
     logPath = self._auxFile(outputPath + ".log")
-    self._runCommand(("%s '%s'") % (wasmCommand, inputPath), logPath, expectedExitCode)
+    self._runCommand(('%s "%s"') % (wasmCommand, inputPath), logPath, expectedExitCode)
     self._compareFile(expectPath + ".log", logPath)
 
     if expectedExitCode != 0:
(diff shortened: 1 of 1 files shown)


interpreter/runtests.py
```diff
@@ -13,9 +13,10 @@
 inputDir = "test"
 expectDir = "expected-output"
 outputDir = os.path.join(inputDir, "output")
+isWindows = os.name == 'nt'
 
 parser = argparse.ArgumentParser()
-parser.add_argument('--wasm', metavar='<wasm-command>', default='./wasm')
+parser.add_argument('--wasm', metavar='<wasm-command>', default=('./wasm' if not isWindows else 'wasm.exe'))
```
rossberg-chromium: Would `os.path.join(".", "wasm")` work?

interpreter/runtests.py
```diff
@@ -30,6 +31,8 @@
 
 class RunTests(unittest.TestCase):
   def _runCommand(self, command, logPath, expectedExitCode = 0):
+    if isWindows:
+      command = command.replace('\'', '"').replace('\\', '/')
```
rossberg-chromium: Ah, sorry, now I get it.

rossberg-chromium:
lgtm

interpreter/runtests.py
shurcooL: One has .exe extension, the other doesn't.

rossberg-chromium: Sure, but that's not needed when used as a command, is it?

Cellule: `os.path.join('.', 'wasm')` results in `./wasm` even on Windows which is not valid.
However, I found that `os.path.join(os.getcwd(), 'wasm')` is a valid command

interpreter/runtests.py
Cellule: Yeah the underlying problem is that the interpreter doesn't recognize `\` in paths, but I wanted to make a quick fix for the time being, I might look into this later on.

rossberg-chromium: I think the problem was merely insufficient quotation in main.ml. I just pushed a quick fix:

https://github.com/WebAssembly/spec/commit/6aa5c271fd2f5ebbdb9c0a3a63fced0e43ccdcd0

Can you rebase and see whether that makes this line unnecessary? Then the isWindows flag might be unnecessary as well.

Cellule:
I was able to remove the `.replace("\\", "/")` with the latest fix. 
However, the single quote was a problem in that script only. On Windows the command `wasm 'my path'` passes the argument `'my path'` to the program instead of `my path` (no quotes)

Cellule:
I don't have merge rights on this repo, can someone merge this for me please?

Cellule:
Thanks

Issue #404, opened 2016-12-19
Do elements need to have valid function indices?


wanderer:
for example, this seems to be valid currently but would not be valid if function indices had to exist.
```
(module
  (type $T (func (result i32)))
  (table anyfunc (elem 0 $t2 $some_non_existance_index 9999999) )

  (func $t1 (type $T) (i32.const 1))
  (func $t2 (type $T) (i32.const 2))

  (func (export "callt") (param $i i32) (result i32)
    (call_indirect $T (get_local $i))
  )
)
```

binji:
This is already invalid in the spec interpreter:
```
test.wast:3.30-3.55: syntax error: unknown function $some_non_existance_index
```
and if you comment that out:
```
test.wast:4.30-4.37: invalid module: unknown function 9999999
```

wanderer:
ah thanks @binji, it seems to pass through wahbt without error.

binji:
Interesting, I guess there isn't a spec test for this yet. I've got a fix for wabt here: https://github.com/WebAssembly/wabt/pull/250

wanderer:
ill add a test

PR #405, opened 2016-12-19
added test with invalid table indices

wanderer:
related https://github.com/WebAssembly/spec/issues/404

### DIFF
--- a/interpreter/test/call_indirect.wast
+++ b/interpreter/test/call_indirect.wast
@@ -363,9 +363,6 @@
 
 ;; invalid table
 (assert_invalid 
-  (module
-    (table anyfunc (elem 0 1))
-    (func (result i32) (i32.const 1))
-  )
-  "unknown function 1"
+  (module (table anyfunc (elem 0 0)))
+  "unknown function 0"
 )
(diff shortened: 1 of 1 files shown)


interpreter/test/call_indirect.wast
```diff
@@ -360,3 +360,12 @@
   )
   "unknown type"
 )
+
+;; invalid table
+(assert_invalid 
+  (module
+    (table anyfunc (elem 0 $t1 44))
+    (func $t1 (result i32) (i32.const 1))
```
rossberg-chromium: What is the purpose of including $t1 in this test?

wanderer: I did have mutlpe functions there, I have updated to the minimal test that wabt was failing at.

rossberg-chromium: Ah, I actually meant the entire function, not just the name. AFAICS, the minimal test would be just
```
(module (table anyfunc (elem 0 0)))
```

wanderer: ah good point! Updated

PR #408, opened 2017-01-23
Tests for checking segments against actual table/memory sizes

rossberg-chromium:
For imports, segment bounds checks ought to be performed wrt to the actual object sizes, not the declared ones.

Also, add convenience target to Makefile to enable running individual tests like
```
make test/memory
```


gahaas:
lgtm

PR #414, opened 2017-01-28
Fix tests to reflect new NaN semantics

rfk:
Here's an initial attempt at https://github.com/WebAssembly/spec/issues/286, let me know if it looks like it's on the right track.  AFAICT it works as-is, but I still plan to:

* Think more about the naming here.  Since asserting a canonical nan is the more common case, I wonder if `assert_return_nan()` should retain its existing semantics, and e.g. `assert_return_nondet_nan` could be added that allows any quiet NaN.

* Add some additional tests such as those suggested in https://github.com/WebAssembly/spec/issues/286#issuecomment-261384346


interpreter/host/js.ml
```diff
@@ -129,6 +129,14 @@ let harness =
   "    throw new Error(\"Wasm return value NaN expected, got \" + actual);\n" ^
   "  };\n" ^
   "}\n" ^
+  "function assert_return_canonical_nan(action) {\n" ^
```
rossberg-chromium: Nit: Insert one empty line before this

interpreter/host/run.ml
```diff
@@ -365,6 +365,18 @@ let run_assertion ass =
     trace ("Asserting return...");
     let got_vs = run_action act in
     let is_nan =
+      (* Accept any NaN that's one everywhere pos_nan is one *)
+      match got_vs with
+      | [Values.F32 got_f32] -> (Int32.logand (F32.to_bits got_f32) (F32.to_bits F32.pos_nan)) = (F32.to_bits F32.pos_nan)
```
rossberg-chromium: Nit: 80 characters

interpreter/host/run.ml
```diff
@@ -365,6 +365,18 @@ let run_assertion ass =
...
+      | [Values.F64 got_f64] -> (Int64.logand (F64.to_bits got_f64) (F64.to_bits F64.pos_nan)) = (F64.to_bits F64.pos_nan)
+      | _ -> false
+    in assert_result ass.at is_nan got_vs print_endline "nan"
+
+  | AssertReturnCanonicalNaN act ->
+    trace ("Asserting return...");
+    let got_vs = run_action act in
+    let is_nan =
+      (* Accept only the two canonical NaNs *)
```
rossberg-chromium: Nit: instead of this comment, rename the variable to `is_canoncial_nan`

interpreter/host/run.ml
```diff
@@ -365,6 +365,18 @@ let run_assertion ass =
     trace ("Asserting return...");
     let got_vs = run_action act in
     let is_nan =
+      (* Accept any NaN that's one everywhere pos_nan is one *)
+      match got_vs with
+      | [Values.F32 got_f32] -> (Int32.logand (F32.to_bits got_f32) (F32.to_bits F32.pos_nan)) = (F32.to_bits F32.pos_nan)
+      | [Values.F64 got_f64] -> (Int64.logand (F64.to_bits got_f64) (F64.to_bits F64.pos_nan)) = (F64.to_bits F64.pos_nan)
```
rossberg-chromium: Same here

interpreter/host/js.ml
```diff
@@ -198,14 +218,30 @@ let assert_return lits ts at =
   in [], List.flatten (List.rev_map test lits)
 
 let assert_return_nan ts at =
-  let var i = Int32.of_int i @@ at in
-  let init i t = [SetLocal (var i) @@ at] in
-  let test i t =
-    [ GetLocal (var i) @@ at;
-      GetLocal (var i) @@ at;
-      Compare (eq_of t) @@ at;
+  let test t =
```
rossberg-chromium: The previous version was intended to work correctly for multiple return values. I believe that dropping the auxiliary locals will consume them in the wrong order, inconsistent with `ts`. (Though instead of using locals you can probably reverse `ts`.)

interpreter/host/js.ml
```diff
@@ -198,14 +218,30 @@ let assert_return lits ts at =
...
+      (* it can be any NaN that's one everywhere the canonical NaN is one *)
+      Const (canonical_nan_of t' @@ at) @@ at;
+      Binary (and_of t') @@ at;
+      Const (canonical_nan_of t' @@ at) @@ at;
+      Compare (eq_of t') @@ at;
+      Test (Values.I32 I32Op.Eqz) @@ at;
+      BrIf (0l @@ at) @@ at ]
+  in [], List.flatten (List.map test ts)
+
+let assert_return_canonical_nan ts at =
```
rossberg-chromium: The only difference between the two functions is the initial mask AFAICS. How about making that a parameter?

gahaas:
I like it.

rossberg-chromium:
Regarding naming, I tend to agree that `assert_return_nan` should retain its original meaning (provided we have any use for it), and the variant from this change should be named `assert_return_quiet_nan` for clarity.

sunfishcode:
The current WebAssembly documentation currently avoids using the terms "quiet" or "signaling", because WebAssembly's "signaling" NaNs don't signal anything, and are only distinguishable from "quiet" NaNs in the semantics by their bit pattern. How about `assert_return_arithmetic_nan`? 

We indeed don't have a use for an assert that accepts any NaN under the current semantics.

rfk:
Thanks for the feedback all, I hope to find some more time to iterate this in the next couple of days.

rossberg-chromium:
@sunfishcode, but 'quiet NaN' is the IEEE terminology for the values it is checking for. Does it get clearer if we invent our own name to mean the same thing?

sunfishcode:
The current NaN rules are designed to accomodate signaling NaNs that actually signal, but I can't predict when this will be exercised. Until then, "WebAssembly doesn't care about signaling NaNs" is tempting as a sufficient and unambiguous mental model for most people, since there are some widespread misconceptions about what signaling NaN actually means.

Also, the use of "arithmetic" in this very specific sense is not my invention.

However, I don't feel strongly about it. My main interest here is in being consistent. If we're going to talk about "quiet" NaNs in the spec/design repos, we should probably define what it means somewhere and use it consistently.

rossberg-chromium:
We might want to wait for the resolution of WebAssembly/design#976. Even the naming question may become moot then.

binji:
Sounds like https://github.com/WebAssembly/design/issues/976 was resolved, can this move forward?

rossberg-chromium:
@binji, sgtm. Still needs comments to be addressed.

rfk:
I'll try to revisit the PR sometime this week.

rfk:
It looks like #417 has introduced the "arithmetic_nan" terminology, so I'm going to go ahead with:

* `assert_return_nan` -> returns positive or negative canonical nan
* `assert_return_arithmetic_nan` -> returns any acceptable nan bitpattern

rfk:
OK, I've updated this and added some more tests for the behaviour of min/max.

rossberg-chromium:
>
> It looks like #417 <https://github.com/WebAssembly/spec/pull/417> has
> introduced the "arithmetic_nan" terminology, so I'm going to go ahead with:
>
>    - assert_return_nan -> returns positive or negative canonical nan
>    - assert_return_arithmetic_nan -> returns any acceptable nan bitpattern
>
>
Shouldn't it be the other way round?

rfk:
> Shouldn't it be the other way round?

It's entirely possible that I mis-understood the terminology here - @sunfishcode can you clarify?

sunfishcode:
For `assert_return_arithmetic_nan`, saying it accepts "any acceptable nan bitpattern" is a little unclear, but the code in the PR looks right.

For `assert_return_nan`, this was named `assert_return_canonical_nan` in the earlier patch, and I think we wanted to keep that name. That would leave us with nothing named `assert_return_nan`, but I think that's what we want, given how the language now works.

rfk:
> this was named assert_return_canonical_nan in the earlier patch, and I think we wanted to keep that name

OK, sounds good, I'll update to match.

rfk:
Updated based on naming feedback above.

test/core/float_exprs.wast
```diff
@@ -861,35 +861,35 @@
...
 (assert_return (invoke "f64.no_fold_le_select" (f64.const nan) (f64.const 0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_le_select" (f64.const 0.0) (f64.const -0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_le_select" (f64.const -0.0) (f64.const 0.0)) (f64.const -0.0))
-(assert_return (invoke "f64.no_fold_gt_select" (f64.const 0.0) (f64.const nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_gt_select" (f64.const 0.0) (f64.const nan)))
 (assert_return (invoke "f64.no_fold_gt_select" (f64.const nan) (f64.const 0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_gt_select" (f64.const 0.0) (f64.const -0.0)) (f64.const -0.0))
 (assert_return (invoke "f64.no_fold_gt_select" (f64.const -0.0) (f64.const 0.0)) (f64.const 0.0))
-(assert_return (invoke "f64.no_fold_ge_select" (f64.const 0.0) (f64.const nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_ge_select" (f64.const 0.0) (f64.const nan)))
```
sunfishcode: Sorry for not noticing this earlier, but all these tests involving `select` can actually use plain `assert_return` to test for a specific bit-pattern, because `select` is defined to return one of its operands verbatim.

test/core/float_exprs.wast
```diff
@@ -908,35 +908,35 @@
...
 (assert_return (invoke "f64.no_fold_le_if" (f64.const nan) (f64.const 0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_le_if" (f64.const 0.0) (f64.const -0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_le_if" (f64.const -0.0) (f64.const 0.0)) (f64.const -0.0))
-(assert_return (invoke "f64.no_fold_gt_if" (f64.const 0.0) (f64.const nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_gt_if" (f64.const 0.0) (f64.const nan)))
 (assert_return (invoke "f64.no_fold_gt_if" (f64.const nan) (f64.const 0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_gt_if" (f64.const 0.0) (f64.const -0.0)) (f64.const -0.0))
 (assert_return (invoke "f64.no_fold_gt_if" (f64.const -0.0) (f64.const 0.0)) (f64.const 0.0))
-(assert_return (invoke "f64.no_fold_ge_if" (f64.const 0.0) (f64.const nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_ge_if" (f64.const 0.0) (f64.const nan)))
```
sunfishcode: Similarly, these tests involving `if` can use plain `assert_return` because `if` returns its value verbatim.

test/core/float_exprs.wast
```diff
@@ -956,35 +956,35 @@
...
 (assert_return (invoke "f64.no_fold_le_select_to_abs" (f64.const 0.0)) (f64.const -0.0))
 (assert_return (invoke "f64.no_fold_le_select_to_abs" (f64.const -0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_gt_select_to_abs" (f64.const nan:0x4000000000000)) (f64.const -nan:0x4000000000000))
-(assert_return (invoke "f64.no_fold_gt_select_to_abs" (f64.const -nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_gt_select_to_abs" (f64.const -nan)))
 (assert_return (invoke "f64.no_fold_gt_select_to_abs" (f64.const 0.0)) (f64.const -0.0))
 (assert_return (invoke "f64.no_fold_gt_select_to_abs" (f64.const -0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_ge_select_to_abs" (f64.const nan:0x4000000000000)) (f64.const -nan:0x4000000000000))
-(assert_return (invoke "f64.no_fold_ge_select_to_abs" (f64.const -nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_ge_select_to_abs" (f64.const -nan)))
```
sunfishcode: Same as above for these `select` tests.

test/core/float_exprs.wast
```diff
@@ -1003,35 +1003,35 @@
...
 (assert_return (invoke "f64.no_fold_le_if_to_abs" (f64.const 0.0)) (f64.const -0.0))
 (assert_return (invoke "f64.no_fold_le_if_to_abs" (f64.const -0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_gt_if_to_abs" (f64.const nan:0x4000000000000)) (f64.const -nan:0x4000000000000))
-(assert_return (invoke "f64.no_fold_gt_if_to_abs" (f64.const -nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_gt_if_to_abs" (f64.const -nan)))
 (assert_return (invoke "f64.no_fold_gt_if_to_abs" (f64.const 0.0)) (f64.const -0.0))
 (assert_return (invoke "f64.no_fold_gt_if_to_abs" (f64.const -0.0)) (f64.const 0.0))
 (assert_return (invoke "f64.no_fold_ge_if_to_abs" (f64.const nan:0x4000000000000)) (f64.const -nan:0x4000000000000))
-(assert_return (invoke "f64.no_fold_ge_if_to_abs" (f64.const -nan)) (f64.const nan))
+(assert_return_canonical_nan (invoke "f64.no_fold_ge_if_to_abs" (f64.const -nan)))
```
sunfishcode: Same as above for these `if` tests.

rfk:
Thanks @sunfishcode, I've reverted the testcases you identified.

sunfishcode:
Looks good to me, and the main changes have general approval above. Thanks for working on this @rfk!

sunfishcode:
@rfk Have you joined the WebAssembly W3C Community Group? https://www.w3.org/community/webassembly/

rfk:
> Have you joined the WebAssembly W3C Community Group?

Hmm, I thought I had, but I don't see myself in the list of participants there.  I'll try again now...

rfk:
OK, I'm now successfully showing up in https://www.w3.org/community/webassembly/participants

sunfishcode:
Thanks!

PR #416, opened 2017-02-01
Script to build wast.js library from interpreter

rossberg-chromium:
This adds a Makefile target `wast.js` cross-compiling the interpreter into a stand-alone JS library, e.g. making the S-expr format available from within JS. Currently, it defines a global object `WebAssemblyText` with a static method `encode`. Example use:
```
let source = '(module (func (export "f") (param i32 i32) (result i32) (i32.add (get_local 0) (get_local 1))))'
let binary = WebAssemblyText.encode(source)
(new WebAssembly.Instance(new WebAssembly.Module(binary))).exports.f(3, 4)  // => 7
```
The inverse is also possible:
```
WebAssemblyText.decode(binary)  // => "<pretty-printed version of source>"
```
Building the library requires node.js and BuckleScript, but using it has no dependencies. The generated file is checked into test/js-api for easy use in tests.


eholk:
Awesome! I've thought something like this would be handy for awhile. I've written a number of .wast tests and it'd be nice to avoid the step of converting to .wasm.

binji:
@eholk I made an emscripten build of wabt that does this too, but the API is a little lower-level: https://github.com/WebAssembly/wabt/blob/master/demo/demo.js#L100

rossberg-chromium:
I also added the inverse decode function to turn a binary back into a string. Might be handy in some debugging cases.

Checked the generated file into test/js-api, so that it can be used for tests easily.

Ms2ger:
It doesn't look like this code is used in this repository. Does anyone know if it should be? If not, perhaps we can remove it again.

PR #417, opened 2017-02-01
Test NaN bitpattern rules in a harness-independent way.

sunfishcode:
Add some tests to check that the underlying platform follows the
basic NaN bitpattern rules without depending on the handling of NaNs
in the test harness or embedding environment.


rossberg-chromium:
LGTM, but probably also has a dependency on WebAssembly/design#976.

sunfishcode:
This change just makes the spec repo test what's already specified in the design repo more thoroughly, so I'm taking the liberty of merging, as it's desirable to keep the two in sync. If #976 leads to changes in the design, these tests can be updated accordingly.

PR #418, opened 2017-02-06
Remove soft-fail mode

rossberg-chromium:
As discussed off-line, we converged on making full checking a requirement. So this PR removes the option for weak checking and turns the soft-fail test into using hard assert_invalid.

This goes along with changes to WebAssembly/design#894.


lukewagner:
lgtm

PR #419, opened 2017-02-07
Test builder: introduce CLI options for specific sub-tasks

bnjbvr:
This should allow us building only interesting subsets of tests for different purposes:

- a browser vendor can build the JS and HTML tests only.
- for the website, we'll be able to build the front page (in a single HTML page) only.

I need to make sure this will interact well at least with Spidermonkey and the webassembly website gh-pages before asking for review. If other browser vendors could confirm this fits their workflow too, this would be great.

### DIFF
--- a/test/build.py
+++ b/test/build.py
@@ -6,6 +6,7 @@ import os
 import glob
 import subprocess
 import shutil
+import multiprocessing as mp
 
 CWD = os.getcwd()
 WASM_EXEC = os.path.join(CWD, '..', 'interpreter', 'wasm')
@@ -42,17 +43,27 @@ def ensure_wasm_executable(path_to_wasm):
 
 
 # JS harness.
+def convert_one_wast_file(inputs):
+    wast_file, js_file = inputs
+    print('Compiling {} to JS...'.format(wast_file))
+    return run(WASM_EXEC, wast_file, '-h', '-o', js_file)
+
 def convert_wast_to_js(out_js_dir):
     """Compile all the wast files to JS and store the results in the JS dir."""
+
+    inputs = []
+
     for wast_file in glob.glob(os.path.join(WAST_TESTS_DIR, '*.wast')):
         # Don't try to compile tests that are supposed to fail.
         if 'fail.wast' in wast_file:
             continue
 
-        print('Compiling {} to JS...'.format(wast_file))
         js_filename = os.path.basename(wast_file) + '.js'
         js_file = os.path.join(out_js_dir, js_filename)
-        result = run(WASM_EXEC, wast_file, '-h', '-o', js_file)
+        inputs.append((wast_file, js_file))
+
+    pool = mp.Pool(processes=8)
+    for result in pool.imap_unordered(convert_one_wast_file, inputs):
         if result.returncode != 0:
             print('Error when compiling {} to JS: {}', wast_file, result.stdout)
 
--- a/test/build.py
+++ b/test/build.py
@@ -16,6 +16,9 @@ JS_TESTS_DIR = os.path.join(SCRIPT_DIR, 'js-api')
 HTML_TESTS_DIR = os.path.join(SCRIPT_DIR, 'html')
 HARNESS_DIR = os.path.join(SCRIPT_DIR, 'harness')
 
+HARNESS_FILES = ['testharness.js', 'testharnessreport.js', 'testharness.css']
+WPT_URL_PREFIX = '/resources'
+
 # Helpers.
 def run(*cmd):
     return subprocess.run(cmd,
@@ -67,7 +70,7 @@ def convert_wast_to_js(out_js_dir):
         if result.returncode != 0:
             print('Error when compiling {} to JS: {}', wast_file, result.stdout)
 
-def build_js(out_js_dir):
+def build_js(out_js_dir, include_harness=False):
     print('Building JS...')
     convert_wast_to_js(out_js_dir)
 
@@ -80,6 +83,8 @@ def build_js(out_js_dir):
 
     print('Copying JS test harness to the JS out dir...')
     for js_file in glob.glob(os.path.join(HARNESS_DIR, '*')):
+        if os.path.basename(js_file) in HARNESS_FILES and not include_harness:
+            continue
         shutil.copy(js_file, harness_dir)
 
     print('Done building JS.')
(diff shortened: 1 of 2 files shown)


bnjbvr:
(not a work-in-progress anymore)

So let me recap: the script has been specialized for different separate objectives, instead of generating everything at once:

- `build.py --js /tmp/out-dir` will convert all the wast tests to JS into the `out-dir`, copy out the pure JS tests there. The testharness is not included, since shells which want to run this need an adapter for polyfilling the `testharness.js` APIs (`assert_equals`, etc.).

- `build.py --html /mozilla/testing/web-platform/mozilla/tests/wasm` will do the same as above, as well as generating one Web Platform test case per JS test case (so one for each WAST test -- those in `core/`, and one for each pure JS test -- those in `js-api`, and one for each HTML test -- those in `html`). These tests are ready to be used in a WPT environment; somebody who'd want to run them would just need to update the WPT `Manifest.json` before running. The test harness is not included, since the WPT server will just... serve it. This has been tested within Gecko, and it works like a charm.

- `build.py --front /website/docs/test` allow us to generate only the *one-click test page* for the website; see for instance [this page](https://bnjbvr.github.io/wasm-website/docs/tests/) for the current output of this command. The PR adding support for doing this flawlessly in the website repo is https://github.com/WebAssembly/website/pull/73 and depends on this PR.

@mtrofin and/or @rossberg-chromium and/or anybody who'd want to, can you please have a look? Happy to answer any questions.

test/build.py
```diff
@@ -39,66 +45,64 @@ def ensure_wasm_executable(path_to_wasm):
...
-    ensure_dir(OUT_JS_DIR)
-    ensure_dir(OUT_HTML_DIR)
-
 # JS harness.
-def replace_js_harness(js_file):
-    test_func_name = os.path.basename(js_file).replace('.', '_').replace('-', '_')
+def convert_one_wast_file(inputs):
+    wast_file, js_file = inputs
+    print('Compiling {} to JS...'.format(wast_file))
+    return run(WASM_EXEC, wast_file, '-h', '-o', js_file)
```
rossberg-chromium: Why `-h`?

test/build.py
```diff
@@ -39,66 +45,64 @@ def ensure_wasm_executable(path_to_wasm):
...
-    with open(js_file, 'w') as f:
-        f.write('\n'.join(content))
+    inputs = []
 
-def convert_wast_to_js():
-    """Compile all the wast files to JS and store the results in the JS dir."""
-    for wast_file in glob.glob(os.path.join(WAST_DIR, '*.wast')):
+    for wast_file in glob.glob(os.path.join(WAST_TESTS_DIR, '*.wast')):
         # Don't try to compile tests that are supposed to fail.
         if 'fail.wast' in wast_file:
```
rossberg-chromium: The runner in test/core checks for `.fail.`, probably should be in sync.

rossberg-chromium: The runner in test/core checks for .fail., probably should be in sync.

test/harness/index.js
```diff
@@ -134,7 +141,7 @@ function assert_invalid(bytes) {
         } catch(e) {
             assert_true(e instanceof WebAssembly.CompileError, "expected invalid failure:");
         }
-    }, "A wast module that should be invalid or malformed.");
+    }, testNum() + "A wast module that should be invalid or malformed.");
```
rossberg-chromium: Instead of doing this at every call site to `test`, why not call a wrapper that does it?

test/build.py
bnjbvr: The output JS file is intended to be used in both a pure JS environment (shell) and the WPT environment; the WPT environment redefines all the testing primitives, as you know.

test/harness/index.js
bnjbvr: Good idea, thanks.

test/build.py
bnjbvr: I see that, but this script doesn't call the `run.py` script which is in test/core. Are you suggesting to use it, or to implement specific logic that tests that the result is incorrect indeed? (the latter doesn't seem too valuable, because we can't specifically check that the failure happens at compile/instanciation/runtime)
Or something else?

test/build.py
rossberg-chromium: Er, sorry, never mind, I was confusing `-h` with `--help`. :)

test/build.py
rossberg-chromium: Much simpler: only suggesting to use the same condition to identify failure tests, i.e., test for the same substring.

bnjbvr:
Thanks for the review, I've updated the PR.

rossberg-chromium:
Thanks. I'll give @mtrofin a chance to comment too before merging.

mtrofin:
Apologies for only getting on this just now. 

I only have a nit comment (a typo).

If you didn't mind, I'd like our colleague Eric Holk to take a peek, too - he has integrated this in our tree & CQ, and is, by now, way more familiar with it all than I am. I pointed him this way.

Thanks!

bnjbvr:
@mtrofin I haven't found your comment about the typo; could it be that it were not submitted? Thanks for the review!

test/harness/index.js
```diff
@@ -140,7 +151,7 @@ function assert_invalid(bytes) {
 const assert_malformed = assert_invalid;
 
 function assert_soft_invalid(bytes) {
-    test(() => {
+    uniqueTest(() => {
```
eholk: What about switching the order of the arguments to `uniqueTest`? I'd find it easier to read if the test name came before the code for the test.

test/harness/index.js
```diff
@@ -16,7 +16,14 @@
 
 'use strict';
 
-// WPT's assert_throw uses a list of predefined, hardcoded know errors. Since
+let testNum = (function() {
```
bnjbvr: It doesn't really provide a stable test identifier for **all** the test cases, but it does for all the ones that are automatically generated from the wast core tests. js-api and HTML tests still manage their id by themselves.

This is needed by the WPT harness. The [wptrunner](https://github.com/w3c/wptrunner) has the ability to automatically generate expectation files, which encode whether a single test is known to be passing or failing when run in a specific browser. For this to work, the (filename, test_id) couple needs to be unique, which makes sense. This is the reason why I've added the unique test id.

test/harness/index.js
bnjbvr: I liked the consistency with the WPT `test` function, which takes arguments in the same order. I don't have a strong opinion though, so I am alright with changing it, if you strongly prefer it.

test/harness/index.js
eholk: Sounds good to me.

test/harness/index.js
eholk: Consistency with WPT `test` seems like a good enough reason. I prefer having the name first, but I think consistency is a stronger argument than my personal preferences, so I'd leave this as is.

bnjbvr:
Thanks for the reviews @eholk @mtrofin and @rossberg-chromium! I've addressed all your comments, so this might be ready for merge.

PR #420, opened 2017-02-08
Fix rounding in f32.convert_u/i32 and f64.convert_u/i64.

sunfishcode:
Fix interpreter's implementations of f32.convert_u/i32 and f64.convert_u/i64
to round large odd numbers correctly.


rossberg-chromium:
LGTM, deferring to @gahaas for float expertise

interpreter/spec/f32_convert.ml
```diff
@@ -15,13 +15,13 @@ let convert_s_i32 x =
 
 (*
  * Similar to convert_u_i64 below, the high half of the i32 range are beyond
- * the range where f32 can represent odd numbers.
+ * the range where f32 can represent odd numbers, though we do need to adjust
+ * the least significant bit to round correctly.
  *)
 let convert_u_i32 x =
```
gahaas: I think f32.convert_u/i64 also needs to be changed.

gahaas: lgtm otherwise

sunfishcode:
The fix used here isn't sufficient for f32.convert_u/i64; I've now filed https://github.com/WebAssembly/spec/issues/421 for that.

PR #422, opened 2017-02-13
Few more checks for unreachable br_table

rossberg-chromium:
Check that inconsistent label types are detected in unreachable code.


lukewagner:
Thanks, lgtm

PR #425, opened 2017-02-17
Add test for interesting unreachable + select + if combination

lukewagner:
This PR adds an interesting case that combines unreachable, select, and if.

### DIFF
--- a/test/core/unreached-invalid.wast
+++ b/test/core/unreached-invalid.wast
@@ -67,7 +67,8 @@
   "type mismatch"
 )
 (assert_invalid
-  (module (func $type-select-any (result i32) (i32.const 1) (if i64 (then (i64.const 0)) (else (unreachable) (select)))))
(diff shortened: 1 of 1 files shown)


test/core/unreached-invalid.wast
```diff
@@ -66,6 +66,10 @@
   ))
   "type mismatch"
 )
+(assert_invalid
+  (module (func $type-select-any (result i32) (i32.const 1) (if i64 (then (i64.const 0)) (else (unreachable) (select)))))
```
rossberg-chromium: Nit: line break. Maybe a more precise name, too?

lukewagner: Renamed to "$unreached-select-in-reachable-if-with-value"

lukewagner:
Yeah, happy to move it somewhere that's obvious, but I don't have a good feel for what's where.  What the test is really testing, though, is that the unbound-type-variable pushed by `select` gets unified with the `if`'s value type to be `i64`.

PR #426, opened 2017-02-17
Fix test of empty module name in instanciation

bnjbvr:
According to https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-constructor, if `Type(Get(importObject, i.module_name))` is not object, then we should throw a `TypeError`, not a `LinkError`. This should be the case here, since `importingModule` tries to import `("", "f")`, the passed imports object's key `""` is undefined.

@rossberg-chromium / @mtrofin, PTAL.

### DIFF
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -334,8 +334,9 @@ test(() => {
     assertThrows(() => new Instance(emptyModule, null), TypeError);
     assertThrows(() => new Instance(importingModule, null), TypeError);
     assertThrows(() => new Instance(importingModule, undefined), TypeError);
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
LGTM, but is there any reason why the import object in this test even has a property 't' instead of just being an empty object?

bnjbvr:
No reason, really. However I think what you're saying could be another test, since it might catch different issues. This one catches cases where the module name would be incorrectly interpreted. I'll add your test though.

bnjbvr:
@rossberg-chromium, PTAL.

rossberg-chromium:
LGTM, merging

PR #428, opened 2017-02-21
Add tests for `if`

rossberg-chromium:
Also, fix a number of assert_invalid tests that weren't properly adapted when we introduced block signatures.


test/core/if.wast
```diff
@@ -0,0 +1,389 @@
+;; Test `block` operator
```
titzer: s/block/if/

PR #433, opened 2017-02-23
Update version to 1

binji:


binji:
Any reason to hold off on this, or should we land when travis is green?

bnjbvr:
Can you also update `kWasmV0` in test/harness/wasm-constants.js?

eholk:
@bnjbvr I opened #434 to update wasm-constants.js, although it'd probably make more sense to include that change here instead.

binji:
Just updated here, sorry for the delay. :)

PR #436, opened 2017-02-27
More LinkError that should be TypeError in async instanciate

bnjbvr:
Sorry, I didn't catch them because we had issues with our `promise_test` JS polyfill, and `jsapi.js` doesn't succeed as a WPT (for some unrelated reason).

@rossberg-chromium, can you please take a look? Thanks!


lukewagner:
This is just the async version of #426, which already merged, so assuming no objections here and merging.

Issue #438, opened 2017-03-02
partially out of bounds store writes to memory


sunfishcode:
This testcase:
```
(module
    (memory 1 1)

    (func (export "store16") (param $v i32)
      (i32.store16 (i32.const 0xffff) (get_local $v))
    )

    (func (export "load8_u") (result i32)
      (i32.load8_u (i32.const 0xffff))
    )
)

(assert_trap (invoke "store16" (i32.const 0x01234567)) "out of bounds memory access")
(assert_return (invoke "load8_u") (i32.const 0))
```

fails in the spec interpreter. The spec interpreter appears to be storing the first byte of the 2-byte store, before detecting that the second byte is out of bounds and trapping.

Semantics.md in the design repo just says "out of bounds accesses trap", which leaves some room for ambiguity, but my understanding is that stores should trap before storing anything.

jfbastien:
Could this also be clarified in the design repo?

sunfishcode:
This is now fixed in the spec interpreter and the design repo.

PR #439, opened 2017-03-02
Don't affect memory on partial out-of-bounds stores

rossberg-chromium:
Addresses #438.

### DIFF
--- a/test/core/memory_trap.wast
+++ b/test/core/memory_trap.wast
@@ -51,6 +51,12 @@
   (func (export "i64.store") (param $a i32) (param $v i64)
     (i64.store (get_local $a) (get_local $v))
   )
+  (func (export "f32.store") (param $a i32) (param $v i32)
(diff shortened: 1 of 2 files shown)


test/core/memory_trap.wast
```diff
@@ -31,3 +31,93 @@
...
+  )
+  (func (export "i64.store16") (param $a i32) (param $v i64)
+    (i64.store16 (get_local $a) (get_local $v))
+  )
+  (func (export "i64.store32") (param $a i32) (param $v i64)
+    (i64.store32 (get_local $a) (get_local $v))
+  )
+  (func (export "i64.store") (param $a i32) (param $v i64)
+    (i64.store (get_local $a) (get_local $v))
+  )
```
sunfishcode: I suggest testing f32 and f64 stores too, since it's not uncommon for JIT implementations to handle those separately from integer stores.

rossberg-chromium: Right, done.

gahaas:
lgtm

Issue #415, opened 2017-01-31
[test] JS API tests should have a stable test ID


eholk:
The in-progress tests at https://github.com/bnjbvr/wasm-spec/blob/gh-pages/test/js-api/jsapi.js should include a unique identifier that won't change as we add more tests. This will make it easier to track and talk about correctness across different engines by being able to precisely say something like "test 114 is failing."

sunfishcode:
Are the description strings passed to test calls, such as "'WebAssembly' object", sufficient?

eholk:
As long as they are unique, I suppose these would work. A more concise identifier might be nice though. At the moment, we have a bunch of tests whose description is "unexpected success in assertInstantiateError".

@mtrofin - Do you have any thoughts on this?

mtrofin:
(If I understand @sunfishcode's proposal correctly) the "WebAssembly object" description string covers a bunch of asserts that may fail independently. 

Suppose there is a point in time when an engine is non-compliant. They want to whitelist failing tests (presumably having a tracking number for the respective issue) until the fix the non-compliance. While that happens, someone unintentionally introduces a new regression which trips a different assert in the same group. That would go undetected until the first regression is addressed.

I feel that, ideally, we would track this external to the tests themselves, but I can't figure out how. 

A proposal: 
- change assert to accept a mandatory parameter which is a description (e.g. TestWasmInstanceExportsIsRO)
- assert is implemented on the official side and dynamically checks test IDs are unique
- assert requires an assert_impl to be provided by host.

bnjbvr:
Another idea: tests could be split up in "groups" or "sections" (I don't know if `testharness.js` has this notion of grouping, but if it has not, we could add it), and we don't bulk tests under a single `test()` or `promise_test()` call; instead, each test has its own `test()` call and a description.

That is, instead of:
```
test(() => {
  assert_something();
  assert_something_else();
  f();
}, "A summary of the tests group");
```

we'd have

```
section(() => {
  test(() => assert_something(...), "something");
  test(() => assert_something_else(...), "something else");
  test(() => f(), "random function call");
}, "A summary of the tests group");
```

The nice side effect would be that `section()` could be rendered differently in the front page, nicely grouping tests (and hiding all those which pass).

I agree with @mtrofin that grouping tests can be undesirable, for the exact reason he mentioned. My initial grouping in `jsapi.js` was more to illustrate the fact that the same test harness can be used across runtimes (browser, shells, front page) and implementations (v8, sm).

[...]

PR #440, opened 2017-03-07
Add more descriptive messages for assertInstantiateError and assertInstantiateSuccess tests

eholk:
This PR gives each of these tests a unique message (although they are a bit verbose, so I'm open to changing them), to make it easier to see exactly what is failing.

@bnjbvr, @rossberg-chromium - could you take a look?

### DIFF
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -714,7 +714,22 @@ test(() => {
     assert_equals(instantiate, instantiateDesc.value);
     assert_equals(instantiate.length, 1);
     assert_equals(instantiate.name, "instantiate");
+
+    // Keep track of all the messages for assertInstantiateError and assertCompileSuccess
+    // to make sure they're all unique.
+    const testNames = new Set();
+    function ensureUniqueTest(message) {
+        if (testNames.has(message)) {
+            if (quit instanceof Function) {
+                print("Test names must all be unique, but '" + message + "' already exists.");
+                quit(1);
+            }
+        }
+        testNames.add(message);
+    }
+
     function assertInstantiateError(args, err, message) {
+        ensureUniqueTest(message);
         promise_test(() => {
             return instantiate(...args)
                 .then(m => {
@@ -751,6 +766,7 @@ test(() => {
     assertInstantiateError([complexImportingModuleBinary, {"c": {"d": scratch_memory}}], TypeError, "instantiate on complexImportingModuleBinary, {'c': {'d': scratch_memory}} throws TypeError");
 
     function assertInstantiateSuccess(module, imports, message) {
+        ensureUniqueTest(message);
         promise_test(()=> {
             return instantiate(module, imports)
                 .then(result => {
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -750,7 +750,7 @@ test(() => {
     assertInstantiateError([complexImportingModuleBinary, {}], TypeError, "assertInstantiateError([complexImportingModuleBinary, {}] => TypeError)");
     assertInstantiateError([complexImportingModuleBinary, {"c": {"d": scratch_memory}}], TypeError, "assertInstantiateError([complexImportingModuleBinary, {'c': {'d': scratch_memory}}] => TypeError)");
 
-    function assertInstantiateSuccess(module, imports) {
+    function assertInstantiateSuccess(module, imports, message) {
         promise_test(()=> {
             return instantiate(module, imports)
                 .then(result => {
@@ -768,19 +768,19 @@ test(() => {
                         assert_equals(desc.enumerable, true);
                         assert_equals(desc.configurable, true);
                     }
-                })}, 'unexpected failure in assertInstantiateSuccess');
+                })}, message);
     }
-    assertInstantiateSuccess(emptyModule);
-    assertInstantiateSuccess(emptyModuleBinary);
-    assertInstantiateSuccess(emptyModuleBinary.buffer);
-    assertInstantiateSuccess(importingModule, {"":{f:()=>{}}});
-    assertInstantiateSuccess(importingModuleBinary, {"":{f:()=>{}}});
-    assertInstantiateSuccess(importingModuleBinary.buffer, {"":{f:()=>{}}});
+    assertInstantiateSuccess(emptyModule, undefined, "assertInstantiateSuccess(emptyModule, undefined)");
+    assertInstantiateSuccess(emptyModuleBinary, undefined, "assertInstantiateSuccess(emptyModuleBinary, undefined)");
+    assertInstantiateSuccess(emptyModuleBinary.buffer, undefined, "assertInstantiateSuccess(emptyModuleBinary.buffer, undefined)");
+    assertInstantiateSuccess(importingModule, {"":{f:()=>{}}}, 'assertInstantiateSuccess(importingModule, {"":{f:()=>{}}})');
+    assertInstantiateSuccess(importingModuleBinary, {"":{f:()=>{}}}, 'assertInstantiateSuccess(importingModuleBinary, {"":{f:()=>{}}})');
+    assertInstantiateSuccess(importingModuleBinary.buffer, {"":{f:()=>{}}}, 'assertInstantiateSuccess(importingModuleBinary.buffer, {"":{f:()=>{}}})');
     assertInstantiateSuccess(complexImportingModuleBinary, {
         a:{b:()=>{}},
         c:{d:scratch_memory},
         e:{f:scratch_table},
-        g:{'⚡':1}});
+        g:{'⚡':1}}, "assertInstantiateSuccess(complexImportingModuleBinary, {...})");
 }, "'WebAssembly.instantiate' function");
 
 })();
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -728,27 +728,27 @@ test(() => {
     }
     var scratch_memory = new WebAssembly.Memory({initial:1});
     var scratch_table = new WebAssembly.Table({element:"anyfunc", initial:1, maximum:1});
-    assertInstantiateError([], TypeError, "assertInstantiateError([] => TypeError)");
-    assertInstantiateError([undefined], TypeError, "assertInstantiateError([undefined] => TypeError)");
-    assertInstantiateError([1], TypeError, "assertInstantiateError([1] => TypeError)");
-    assertInstantiateError([{}], TypeError, "assertInstantiateError([{}] => TypeError)");
-    assertInstantiateError([new Uint8Array()], CompileError, "assertInstantiateError([Uint8Array()] => CompileError)");
-    assertInstantiateError([new ArrayBuffer()], CompileError, "assertInstantiateError([ArrayBuffer()] => CompileError)");
-    assertInstantiateError([new Uint8Array("hi!")], CompileError, "assertInstantiateError([Uint8Array('hi!')] => CompileError)");
-    assertInstantiateError([new ArrayBuffer("hi!")], CompileError, "assertInstantiateError([ArrayBuffer('hi!')] => CompileError)");
-    assertInstantiateError([importingModule], TypeError, "assertInstantiateError([importingModule] => TypeError)");
-    assertInstantiateError([importingModule, null], TypeError, "assertInstantiateError([importingModule, null] => TypeError)");
-    assertInstantiateError([importingModuleBinary, null], TypeError, "assertInstantiateError([importingModuleBinary, null] => TypeError)");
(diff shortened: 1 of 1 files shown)


bnjbvr:
I appreciate that these messages are just plain different than what we had before, but given that they might appear on the web site at some point, I'd go even further and have them document for humans what they test. For instance:

```
assertInstantiateError([], TypeError, "instantiate with no args throws a TypeError"); 
```

What do you think?

eholk:
@bnjbvr Thanks, I think your suggestion makes for much more readable names. I pushed some new updates to the name.

test/js-api/jsapi.js
```diff
@@ -728,27 +728,27 @@ test(() => {
...
-    assertInstantiateError([complexImportingModuleBinary, {"c": {"d": scratch_memory}}], TypeError, "assertInstantiateError([complexImportingModuleBinary, {'c': {'d': scratch_memory}}] => TypeError)");
+    assertInstantiateError([], TypeError, "instantiate with no args throws TypeError)");
+    assertInstantiateError([undefined], TypeError, "instantiate on undefined throws TypeError)");
+    assertInstantiateError([1], TypeError, "instantiate on a number throws TypeError)");
+    assertInstantiateError([{}], TypeError, "instantiate on empty object throws TypeError)");
+    assertInstantiateError([new Uint8Array()], CompileError, "instantiate on empty Uint8Array throws CompileError)");
+    assertInstantiateError([new ArrayBuffer()], CompileError, "instantiate on empty ArrayBuffer throws CompileError)");
+    assertInstantiateError([new Uint8Array("hi!")], CompileError, "instantiate on Uint8Array('hi!') throws CompileError)");
+    assertInstantiateError([new ArrayBuffer("hi!")], CompileError, "instantiate on ArrayBuffer('hi!') throws CompileError)");
```
shurcooL: There's an extra closing parenthesis inside the strings above (but not the ones below). Is that unintentional?

eholk: Nope, just sloppy search and replace on my part. It should be fixed now. Thanks for catching this!

test/js-api/jsapi.js
```diff
@@ -724,33 +724,33 @@ test(() => {
...
+    assertInstantiateError([{}], TypeError, "instantiate on empty object throws TypeError");
+    assertInstantiateError([new Uint8Array()], CompileError, "instantiate on empty Uint8Array throws CompileError");
+    assertInstantiateError([new ArrayBuffer()], CompileError, "instantiate on empty ArrayBuffer throws CompileError");
+    assertInstantiateError([new Uint8Array("hi!")], CompileError, "instantiate on Uint8Array('hi!') throws CompileError");
+    assertInstantiateError([new ArrayBuffer("hi!")], CompileError, "instantiate on ArrayBuffer('hi!') throws CompileError");
+    assertInstantiateError([importingModule], TypeError, "instantiate on importingModule throws TypeError");
+    assertInstantiateError([importingModule, null], TypeError, "instantiate on importingModule, null throws TypeError");
+    assertInstantiateError([importingModuleBinary, null], TypeError, "instantiate on importingModuleBinary, null throws TypeError");
+    assertInstantiateError([emptyModule, null], TypeError, "instantiate on emptyModule throws TypeError");
+    assertInstantiateError([importingModuleBinary, null], TypeError, "instantiate on importingModuleBinary, null throws TypeError");
```
bnjbvr: pre-existing, but this test is actually duplicated above and could be removed.

test/js-api/jsapi.js
```diff
@@ -724,33 +724,33 @@ test(() => {
...
-    function assertInstantiateSuccess(module, imports) {
+    assertInstantiateError([], TypeError, "instantiate with no args throws TypeError");
+    assertInstantiateError([undefined], TypeError, "instantiate on undefined throws TypeError");
+    assertInstantiateError([1], TypeError, "instantiate on a number throws TypeError");
+    assertInstantiateError([{}], TypeError, "instantiate on empty object throws TypeError");
+    assertInstantiateError([new Uint8Array()], CompileError, "instantiate on empty Uint8Array throws CompileError");
+    assertInstantiateError([new ArrayBuffer()], CompileError, "instantiate on empty ArrayBuffer throws CompileError");
+    assertInstantiateError([new Uint8Array("hi!")], CompileError, "instantiate on Uint8Array('hi!') throws CompileError");
+    assertInstantiateError([new ArrayBuffer("hi!")], CompileError, "instantiate on ArrayBuffer('hi!') throws CompileError");
+    assertInstantiateError([importingModule], TypeError, "instantiate on importingModule throws TypeError");
```
bnjbvr: How about "instantiate a module importing functions without passing an imports object throws TypeError"?

test/js-api/jsapi.js
```diff
@@ -724,33 +724,33 @@ test(() => {
...
+    assertInstantiateError([importingModule, null], TypeError, "instantiate on importingModule, null throws TypeError");
+    assertInstantiateError([importingModuleBinary, null], TypeError, "instantiate on importingModuleBinary, null throws TypeError");
+    assertInstantiateError([emptyModule, null], TypeError, "instantiate on emptyModule throws TypeError");
+    assertInstantiateError([importingModuleBinary, null], TypeError, "instantiate on importingModuleBinary, null throws TypeError");
+    assertInstantiateError([importingModuleBinary, undefined], TypeError, "instantiate on importingModuleBinary, undefined throws TypeError");
+    assertInstantiateError([importingModuleBinary, {}], TypeError, "instantiate on importingModuleBinary, {} throws TypeError");
+    assertInstantiateError([importingModuleBinary, {"":{g:()=>{}}}], LinkError, "instantiate on importingModuleBinary, {'':{g:()=>{}}} throws LinkError");
```
bnjbvr: How about "instantiate on an module importing a function named f, not provided in the imports object, throws LinkError"? (etc for other tests below)

test/js-api/jsapi.js
```diff
@@ -724,33 +724,33 @@ test(() => {
...
+    assertInstantiateError([new Uint8Array("hi!")], CompileError, "instantiate on Uint8Array('hi!') throws CompileError");
+    assertInstantiateError([new ArrayBuffer("hi!")], CompileError, "instantiate on ArrayBuffer('hi!') throws CompileError");
+    assertInstantiateError([importingModule], TypeError, "instantiate on importingModule throws TypeError");
+    assertInstantiateError([importingModule, null], TypeError, "instantiate on importingModule, null throws TypeError");
+    assertInstantiateError([importingModuleBinary, null], TypeError, "instantiate on importingModuleBinary, null throws TypeError");
+    assertInstantiateError([emptyModule, null], TypeError, "instantiate on emptyModule throws TypeError");
+    assertInstantiateError([importingModuleBinary, null], TypeError, "instantiate on importingModuleBinary, null throws TypeError");
+    assertInstantiateError([importingModuleBinary, undefined], TypeError, "instantiate on importingModuleBinary, undefined throws TypeError");
+    assertInstantiateError([importingModuleBinary, {}], TypeError, "instantiate on importingModuleBinary, {} throws TypeError");
```
bnjbvr: How about "Providing an empty imports object to a module that expects imports throws TypeError"?

test/js-api/jsapi.js
```diff
@@ -768,19 +768,19 @@ test(() => {
...
+    assertInstantiateSuccess(emptyModuleBinary.buffer, undefined, "instantiate on emptyModuleBinary.buffer, undefined succeeds");
+    assertInstantiateSuccess(importingModule, {"":{f:()=>{}}}, 'instantiate on importingModule, {"":{f:()=>{}}} succeeds');
+    assertInstantiateSuccess(importingModuleBinary, {"":{f:()=>{}}}, 'instantiate on importingModuleBinary, {"":{f:()=>{}}} succeeds');
+    assertInstantiateSuccess(importingModuleBinary.buffer, {"":{f:()=>{}}}, 'instantiate on importingModuleBinary.buffer, {"":{f:()=>{}}} succeeds');
     assertInstantiateSuccess(complexImportingModuleBinary, {
         a:{b:()=>{}},
         c:{d:scratch_memory},
         e:{f:scratch_table},
-        g:{'⚡':1}});
+        g:{'⚡':1}}, "instantiate on complexImportingModuleBinary, {...} succeeds");
```
bnjbvr: How about "instantiate complexImportingModuleBinary with all expected imports succeeds"?

test/js-api/jsapi.js
eholk: Done.

test/js-api/jsapi.js
eholk: Done.

test/js-api/jsapi.js
eholk: Done.

test/js-api/jsapi.js
eholk: Done (I got a few more of the tests below, but may have missed some)

test/js-api/jsapi.js
eholk: Done.

eholk:
I just pushed a few more changes, tweaking the names some more. I also added the `Set` to make sure the test names are unique, at least when run from the console.

Do we have an ideal line length for this repo? It looks like the existing code was already beyond 80 columns. This change now has a 206 character line, which is pretty unreasonable :)

test/js-api/jsapi.js
```diff
@@ -714,7 +714,22 @@ test(() => {
...
     assert_equals(instantiate.length, 1);
     assert_equals(instantiate.name, "instantiate");
-    function assertInstantiateError(args, err) {
+
+    // Keep track of all the messages for assertInstantiateError and assertCompileSuccess
+    // to make sure they're all unique.
+    const testNames = new Set();
+    function ensureUniqueTest(message) {
+        if (testNames.has(message)) {
+            if (quit instanceof Function) {
```
bnjbvr: The current code will throw a TypeError if `quit` isn't defined (yay JS!). How about `typeof quit === 'function'` instead?

rossberg-chromium:
I like more informative messages, but I wonder how this approach can scale. The large chunk of the tests that the browser will run are from the core suite, and that contains 13 000+ individual assertions. It doesn't seem realistic to annotate all of them with informative (let alone unique) messages.

bnjbvr:
Indeed, it's not clear for the core tests. For these, a unique sequential number is generated per assertion, at the moment, which is not too helpful to humans, but it allows the WPT harness to automatically intercept errors and generate expectation files.

Maybe a good rule of thumb is to add new descriptive messages at least for new test cases? Usually, a single `test()` (or `promise_test()`) can group several actions and assertions. When seeing a failure in the WPT HTML report, one can see the unique error message, grep for it and find which exact line is affected in the test. (It's possible to hide failures or successes in the WPT HTML report)

rossberg-chromium:
How about we simply report file name/line numbers from the original .wast? The spec interpreter already embeds this information in the generated JS, but currently only as a comment. It would be easy to change this so that it becomes an additional string argument to the harness functions.

Not sure a more heavy-weight approach is worthwhile. Implementers will need to look at the source anyway. And for other random users running the tests I doubt that any form of clear text will be particularly useful.

bnjbvr:
> How about we simply report file name/line numbers from the original .wast?

Would be great indeed.

For the JS tests, we could even get the stack trace (with a caught Error) and generate a description containing the caller's source line as well.

test/js-api/jsapi.js
eholk: Done. I wasn't sure what the idiomatic way to check whether I should or can call something is.

eholk:
I just reflowed this to 100 columns, since that seems to match the rest of the file.

@rossberg-chromium I like the idea of using filename and line numbers from the original .wast, but that approach isn't really relevant here because these tests are for the JS API, which aren't generated from a .wast. Also, I'd be worried that somebody could insert a line in one of the .wasts and then change all the test IDs. 

Ultimately, I'd like something unique and stable to identify tests (#415) so that VMs can list the ones they currently fail on (for example, [in V8](https://cs.chromium.org/chromium/src/v8/test/mjsunit/wasm/jsapi-harness.js?q=jsapi-harness.js+package:%5Echromium$&l=7)). I've been using the human readable description, since it's the best we have so far, but it's not the right solution for the long term.

rossberg-chromium:
Per-assertion IDs may not be too relevant for the core suite, since that's split into many files. In practice, the file names probably provide stable IDs with sufficient granularity (or we can make them so).

(I'd actually think it would be advantageous if the JS tests was split up into smaller pieces as well.)

eholk:
Ah, I like the idea of using filenames. We should move the JS API tests in
that direction too, and I think that will give a reasonable stable ID for
my purposes.

rossberg-chromium:
@eholk, is this PR dead?

eholk:
I still think we should have a way of uniquely identifying name, but this doesn't seem like the most ideal way to do it. Maybe I'll revisit it if and when we get agreement on how to do that.

PR #442, opened 2017-03-10
[document] Describe structure; reorg; infrastructure

rossberg-chromium:
Some progress on the textual spec:

- Some additions to the intro.
- Section on structure, defining and describing abstract syntax.
- Some restructuring of outline.
- Various infrastructure support (math macros, GHP deploy script)
- CSS tweaks.

You can preview the result at http://webassembly.github.io/spec/


document/macros.rst
```diff
@@ -0,0 +1,171 @@
...
+.. |FUNCS| mathmacro:: \mathsf{funcs}
+.. |TABLES| mathmacro:: \mathsf{tables}
+.. |MEMS| mathmacro:: \mathsf{mems}
+.. |GLOBALS| mathmacro:: \mathsf{globals}
+.. |LOCALS| mathmacro:: \mathsf{locals}
+.. |LABELS| mathmacro:: \mathsf{labels}
+.. |IMPORTS| mathmacro:: \mathsf{imports}
+.. |EXPORTS| mathmacro:: \mathsf{exports}
+
+.. |UNREACHABLE| mathmacro:: \mathsf{unreachable}
```
sunfishcode: Other mathmacros are lower-case; is there a reason the instructions are upper-case? It would seem to minimize source versus rendered-text differences to spell the instruction names as lower-case too.

document/macros.rst
```diff
@@ -0,0 +1,171 @@
...
+.. |NOP| mathmacro:: \mathsf{nop}
+.. |BLOCK| mathmacro:: \mathsf{block}
+.. |LOOP| mathmacro:: \mathsf{loop}
+.. |IF| mathmacro:: \mathsf{if}
+.. |ELSE| mathmacro:: \mathsf{else}
+.. |END| mathmacro:: \mathsf{end}
+.. |BR| mathmacro:: \mathsf{br}
+.. |BRIF| mathmacro:: \mathsf{br\_if}
```
sunfishcode: And in any case, preserving the underscores in the instruction names that have them would improve readability of the source, and searchability as well.

document/syntax/values.rst
```diff
@@ -0,0 +1,140 @@
...
+     \by^{N/8} \\
+   \end{array}
+
+The two possible sizes :math:`N` are 32 and 64.
+
+
+Conventions
+...........
+
+* The meta variable :math:`z` ranges over floating point values.
```
sunfishcode: My concern with these single-letter conventions is the use case of jumping into the middle of the document, with a general understanding of wasm, but without full context of the document's conventions. This is something I've found difficult in the interpreter.

It seems to me to make sense to define meta variables in context, rather than globally, so that any subtle issues, like what the specific set of valid values is, or what the consequences are if someone uses such a value outside that set, can be handled in context. For example, it looks like `n` is sometimes used to abstract over `i32` and `i64`. It seems worth saying that `n` can only be those to values, in that context, and in that case, it's fully defined for that context and doesn't need a global convention.

Alternatively, would it be possible to make `n` be a hyperlink to the place where it's defined?

sunfishcode:
It's quite difficult to review changes when the rendered form linked above has changes from multiple PRs as well as apparently changes which are not yet submitted as PRs. Is there an easier way to view the rendered form containing just the changes for a specific PR?

document/syntax/values.rst
```diff
@@ -0,0 +1,140 @@
...
+   \production{signed integers} & \sN &::=&
+     -2^{N-1} ~|~ \dots ~|~ {-}1 ~|~ 0 ~|~ 1 ~|~ \dots ~|~ 2^{N-1}{-}1 \\
+   \production{uninterpreted integers} & \iN &::=&
+     \uN ~|~ \sN \\
+   \end{array}
+
+The latter class defines *uninterpreted* integers, whose signedness interpretation can vary depending on context.
+A 2's complement conversion is assumed for values that are out-of-range for a chosen interpretation.
+That is, semantically, when interpreted as unsigned, negative values :math:`-n` convert to :math:`2^N-n`,
+and when interpreted as signed, positive values :math:`n \geq 2^{N-1}` convert to :math:`n-2^N`.
```
rossberg-chromium: Well, you don't want the AST to contain programs that are not actually expressible in the concrete syntax. Otherwise you gonna run into annoying impedance mismatches. For example, you'd have to specify what happens for out-of-bounds values (e.g. an offset >= 2^32) in places, or impose range restrictions on a meta level, which is more difficult to handle.

Signedness is especially relevant for semantics, think natural vs integer domains. In particular, most numbers in the AST are natural, and defining that directly avoids the need to say anything about negative values all over the place.

You may be right, though, that expressing this in terms of signedness is mixing levels a little. It might make sense to distinguish static (all natural) numbers as they occur in the AST, from dynamic values like used for computation and only occurring in the AST with const instructions. Would naming the former "nat" make that distinction clearer?

The bit about conversion of uninterpreted integers is intended more as an explanation than a definition here. The actual definition will occur in the places of the spec where such values are actually produced/consumed. But I thought it's useful to mention upfront how they will be dealt with. But if the explanation is unclear it probably needs improvement. I reworded it somewhat, not sure it's better now. Also, it should link to the part of the spec that makes this more precise; I added a todo for that.

document/syntax/values.rst
```diff
@@ -0,0 +1,140 @@
...
+
+
+.. _syntax-float:
+.. index:: ! floating-point number
+   pair: abstract syntax; floating-point number
+
+Floating-point Numbers
+~~~~~~~~~~~~~~~~~~~~~~
+
+*Floating-point numbers* are represented as binary values according to the `IEEE-754 <http://ieeexplore.ieee.org/document/4610935/>`_ standard.
```
rossberg-chromium: Done.

document/syntax/values.rst
rossberg-chromium: Agreed for ones that are used rarely, such as this one. Removed this one.

Some others, e.g. n or t, occur very frequently. Their meaning should always be clear from immediate context (and I believe all use sites so far are cross-linked heavily), but pointing out a consistent convention upfront still seems beneficial.

document/syntax/values.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. _syntax-vec:
+.. index:: ! vector
+   pair: abstract syntax; vector
+
+Vectors
+~~~~~~~
+
+*Vectors* are self-contained sequences of the form :math:`A^n` (or :math:`A^\ast`),
+where the :math:`A`-s can either be values or complex constructions.
```
rossberg-chromium: Yes, it wasn't great. But I realised what the actual property is that these have in contrast to arbitrary sequences: their length is bounded to 32 bit. So I reworded to "bounded" and refined the grammar to express that restriction.

rossberg-chromium:
>
> It's quite difficult to review changes when the rendered form linked above
> has changes from multiple PRs as well as apparently changes which are not
> yet submitted as PRs. Is there an easier way to view the rendered form
> containing just the changes for a specific PR?
>

@sunfishcode, acknowledged. Any suggestion where we could upload builds of
multiple PRs without too much hassle? GH pages is a PITA when it comes to
these things.

Right now the version on GH.io should be the accumulation of this and the
other PR. What other changes are you seeing?

rossberg-chromium:
The mathmacro plugin had serious issues. In particular, it created gigantic regexps and overall regexp compile times were quadratic in the number of macros -- builds became superslow.

So I reimplemented a similar thing, but much faster and with more functionality. In particular, it supports macros with params and allows using previously defined macros in expansions. The only thing I couldn't get to work was allowing underscores in def names, because that is in inherent conflict with the special role of underscores in TeX (TeX's own \def mechanism substitutes token sequences instead of just individual names, so can handle it, but that cannot be simulated with simple search+replace).

The new thing also supports a somewhat cleaner handling of crossrefs from math, which now work in the PDF as well. Still not perfect but hopefully good enough.

lukewagner:
Not reading the code, but just having spent some time reading over the rendered document, overall this is looking really good.

I don't know if this is the right PR for it, but some basic UX-level questions/requests:
* when I'm on some page in the rhs content pane, could the associated lhs TOC item be bolded?
* is there any way, at some point in time, to optimize loading so there isn't the visible restyling delay?
* at some point, would each concept/phase in the Overview page should get an anchor and get linked to pervasively?

lukewagner:
Ooh, I also just noticed and like that "Index of Instructions" at the end, that should be quite useful.

distransient:
It seems like what Luke's referring to is MathJax running asynchronously

rossberg-chromium:
>
>
>    - when I'm on some page in the rhs content pane, could the associated
>    lhs TOC item be bolded?
>
>
I like that suggestion, but the theme does not seem to offer that option.
If I find time (later) I might try to investigate if it can be hacked
somehow.


>    - is there any way, at some point in time, to optimize loading so
>    there isn't the visible restyling delay?
>
> As @distransient mentions, that is MathJax in operation, which is dog
slow. The only other option would be KaTeX. Unfortunately, I already tried
that, and while being much faster it had showstopper render bugs (like not
handling super/subscripts correctly, or putting the bar of `\frac` an
entire cm too high, striking out the premise instead). Maybe once that has
matured more we will be able to switch.


>    - at some point, would each concept/phase in the Overview page should
>    get an anchor and get linked to pervasively?
>
>
Yes, that's the idea. There are various dangling xrefs right now as well.
Once all material is there I intend to make a pass to tie up all that.

lukewagner:
Righto, makes sense, lgtm!

rossberg-chromium:
Okay, merging based on LGTM above, some offline discussion, and (I believe) no outstanding comments. Expecting further iterations anyways.

PR #445, opened 2017-03-17
[document] First stab at specifying validation

rossberg-chromium:
Please see test rendering on https://webassembly.github.io/spec/


document/validation/conventions.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Validation checks that a WebAssembly module is well-formed.
+Only valid modules can be :ref:`instantiated <instantiation>`.
+
+Validity is defined by a *type system* over the :ref:`abstract syntax <syntax>` of both instructions and modules.
+For each piece of abstract syntax, there is a typing rule that specifies the constraints that apply to it.
+All rules are given in two *equivalent* forms:
+
+1. In *prose*, describing the meaning in intuitive form.
+2. In *formal notation*, describing the rule in mathematical form.
```
sunfishcode: When we discussed this, we agreed that formal notation would be placed in an appendix, rather than being used within the main document itself.

lukewagner: Having spent some time scanning over this on a few occasions now, I think I really like this current interleaved formatting, for a few reasons:
* it seems easier to keep the two precisely coherent and spot discrepancies when they are juxtaposed
* as a future frequent user of this document, I'll definitely benefit from having the formal rules right next to bulleted explanation
* with the mini-tutorial given in conventions.html#formal-notation, I expect anyone not already familiar with the notation could pick it up by first observing the basic style and then learning by example from reading a few rules
* there's a pretty minimal impact on readability even if you just want to skip over it
* the interleaved heading / summary / notation / explanation form matches the "Structure" section which I also really like as is; in general, I think this is a good high-level style to use throughout the doc

IIRC, this also matches @dherman's recommendation last time we discussed this organizational question.  Happy to have any comments from you too Dave.

lukewagner:
Overall, lgtm.  I'm expecting we'll iterate more as everything comes together.

PR #449, opened 2017-03-30
Add link to gh-pages

jamen:
I built the html locally before I knew the gh-pages existed.  Would good to link it somewhere? 

P.S. The URL on the repository just points to `webassembly.github.io`, instead of with `/spec`.  Might be a typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ specification language.
 It also holds the WebAssembly testsuite, which tests numerous aspects of
 conformance to the spec.
 
-View the HTML rendered spec at [webassembly.github.io/spec](webassembly.github.io/spec/).
+View the spec at [webassembly.github.io/spec](webassembly.github.io/spec/).
 
 At this time, the contents of this repository are under development and known
 to be "incomplet and inkorrect".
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,8 @@ specification language.
 It also holds the WebAssembly testsuite, which tests numerous aspects of
 conformance to the spec.
 
+View the spec at [webassembly.github.io/spec](webassembly.github.io/spec/).
```
shurcooL: This is going to be a broken link. Without a scheme, it gets treated as a relative link to:

https://github.com/WebAssembly/spec/blob/master/webassembly.github.io/spec

To fix that, you should add a scheme, preferably https://.

jamen: Thanks my bad

rossberg-chromium:
Thanks! How about changing that to "work-in-progress spec", for extra clarity?

Can you be a bit more specific about what "URL on the repository" you are referring to in your PS?

shurcooL:
> Can you be a bit more specific about what "URL on the repository" you are referring to in your PS?

I think he's referring to the URL that's next to the repository description. See the bottom right corner of this screenshot:

image

jamen:
Alright I updated the text, and @shurcooL is correct

rossberg-chromium:
Thanks @jamen.

@shurcooL, I see. Seems like I don't have the necessary admin rights to change this. @lukewagner, do you have access?

PR #451, opened 2017-03-30
Correct WebAssembly -> WebAssemblyText global object references

jayphelps:
These examples currently use `WebAssembly` accidentally, instead of `WebAssemblyText`


rossberg-chromium:
Ah, thanks!

PR #452, opened 2017-03-31
Several more tests.

sunfishcode:
In https://github.com/WebAssembly/spec/pull/448 it was noticed that the NaN tests for promote and demote tested for a specific bitpattern; they should the new `assert_return_canonical_nan`.

#414 added several tests using `assert_return_arithmetic_nan` to test various operators on signaling NaN. This PR contains a patch which adds such test coverage for more operators.

With this last change, f32.wast and f64.wast are growing fairly large, so this PR also contains a patch to split out the bitwise floating-point ops (neg, abs, copysign) into their own file.

And, this adds several other miscellaneous tests.


gahaas:
Thanks for writing these tests, lgtm

PR #455, opened 2017-04-05
[document] Specify binary format

rossberg-chromium:
The spec needs to define both the format itself and what abstract syntax it decodes into. After trying various approaches, I arrived at an attribute grammar as what seems to be the most effective and readable way to specify both together.

As before, see https://webassembly.github.io/spec/ for a preview.


lukewagner:
From the parts I read carefully and from a brief scan overall, this looks really good; attribute grammar seems to match pretty well.

A few small nits:
* Pre-existing, but it looks like `block_type` is renamed to `resulttype`.  Being defined right before `functype`, it's a bit confusing that `resulttype` isn't used as the result of `functype`.  The nice thing about "block type" is you didn't expect it to show up anywhere but, well, the block's type.  Is there any other type that doesn't sound like it should be use in `functype`?
* Could perhaps a parameterized grammar rule be used to factor out all the interleaved `customsec*` in the `module` rule so it looks a bit prettier?  The new rule could go in the "Custom Section" section.
* In the note about the module version at the very end, could you add something to the effect that this version is expected to change infrequently if ever; forward-compatible extension is intended to be achieved by adding section, type, op- and other codes without bumping the version.

rossberg-chromium:
>
>
>    - Pre-existing, but it looks like block_type is renamed to resulttype.
>    Being defined right before functype, it's a bit confusing that
>    resulttype isn't used as the result of functype. The nice thing about
>    "block type" is you didn't expect it to show up anywhere but, well, the
>    block's type. Is there any other type that doesn't sound like it should be
>    use in functype?
>
> Two reasons I didn't call it `blocktype` in the abstract syntax:

1. Currently, all block types are result types, but not all result types
are block types. Specifically, the type of a loop label is not the type of
any block, so saying that labels have block type would be sort of wrong.

2. In the future, we may generalise block types to not be result types at
all, but function types. You would still need result types for classifying
labels then.

We could introduce `blocktype` as a separate type, for now defined to be a
result type. But that feels a bit redundant.

(FWIW, I defined `functype` via `resulttype` initially, but it doesn't fit
the binary format or the uses for typing instruction sequences. But once we
have multiple values that impedance mismatch will go away.)


>
>    - Could perhaps a parameterized grammar rule be used to factor out all
>    the interleaved customsec* in the module rule so it looks a bit
>    prettier? The new rule could go in the "Custom Section" section.
>
> Hm, while I agree that it looks a bit ugly, it arguably is the faithful
reflection of the structure. Maybe I misunderstand what you have in mind,
but wouldn't it be a bit strange to fold the occurrence of custom sections
into the other sections?


>    - In the note about the module version at the very end, could you add
>    something to the effect that this version is expected to change
>    infrequently if ever; forward-compatible extension is intended to be
>    achieved by adding section, type, op- and other codes without bumping the
>    version.
>
>
Done.

lukewagner:
> Two reasons I didn't call it `blocktype` in the abstract syntax: [...]

For the binary encoding rules, though, it is a fundamental irregularity between block types and function types that block types have, effectively, "optimized" alternate encodings for ([]->[], []->[i32], []->[i64], ...).  It seems like, for the encoding of generalized blocks, you'd need:
```
blocktype ::= 0x40          => [] -> []
           | vt : valtype   => [] -> [vt]
           | ft : functype  => ft
```
and this would force the abstract syntax to introduce define `blocktype ::= functype` so that blocks could use `blocktype` to "guide" the attribute grammar down the right path.  So that's why, in v.1.0, I was expecting a `blocktype` (but with a suitable v.1 definition).

> Maybe I misunderstand what you have in mind, but wouldn't it be a bit strange to fold the occurrence of custom sections into the other sections?

So right now we have the `section` parameterized grammar rule.  What if we had a higher-level parameterized rule, defined right after `customsec`:
```
sec(N,B) ::= customsec* section(N,B) customsec*
```
and then all the definitions of builtin sections would use `sec` instead of `section` (or some other more sensible pair of names).  This rule would logically match the current Note that custom sections basically can go anywhere in the module.

rossberg-chromium:
On 10 April 2017 at 20:47, Luke Wagner <notifications@github.com> wrote:

> Two reasons I didn't call it blocktype in the abstract syntax: [...]
>
> For the binary encoding rules, though, it is a fundamental irregularity
> between block types and function types that block types have, effectively,
> "optimized" alternate encodings for ([]->[], []->[i32], []->[i64], ...). It
> seems like, for the encoding of generalized blocks, you'd need:
>
> blocktype ::= 0x40          => [] -> []
>            | vt : valtype   => [] -> [vt]
>            | ft : functype  => ft
>
> and this would force the abstract syntax to introduce define blocktype
> ::= functype so that blocks could use blocktype to "guide" the attribute
> grammar down the right path. So that's why, in v.1.0, I was expecting a
> blocktype (but with a suitable v.1 definition).
>

I see what you mean. (I was kind of hoping that at the time we generalise
block types to function types we would also generalise the encoding of all
function types to be able to use the same compression, but admittedly that
should be kept separate.)

Renamed the type in the binary grammar to `blocktype` and added a comment
that this is the only form of resulttype occurring in the binary format.


> Maybe I misunderstand what you have in mind, but wouldn't it be a bit
> strange to fold the occurrence of custom sections into the other sections?
>
> So right now we have the section parameterized grammar rule. What if we
> had a higher-level parameterized rule, defined right after customsec:
>
> sec(N,B) ::= customsec* section(N,B) customsec*
>
> and then all the definitions of builtin sections would use sec instead of
> section (or some other more sensible pair of names). This rule would
> logically match the current Note that custom sections basically can go
> anywhere in the module.
>

Hm, I feel your desire to make the grammar look smaller, but I think this
would give the misleading impression that every custom section somehow
"belongs" to one of the known sections. Instead of reflecting a module's
structure as a flat sequence of independent sections. Moreover, it would
introduce a significant ambiguity (shift/reduce & reduce/reduce conflicts)
in the grammar, for any custom section occurring between two regular ones.

lukewagner:
(Oops, I missed the reply, sorry for the delay)  That's a nice idea about merging `functype` and `blocktype` in the future; that should make the whole irregularity go away.  Also, I guess I see what you mean about not having a section appear to "own" a custom section.  Anyhow, lgtm!

PR #456, opened 2017-04-05
Add br_table test that targets typed loop and void block

jgravelle-google:
This adds a spec test for an unhandled `br_table` cause, found in wabt (https://github.com/WebAssembly/wabt/issues/389) and V8 (https://bugs.chromium.org/p/v8/issues/detail?id=6204)

### DIFF
--- a/test/core/br_table.wast
+++ b/test/core/br_table.wast
@@ -1203,6 +1203,20 @@
       )
     )
   )
+
+  (func (export "nested-br_table-loop-block") (param i32) (result i32)
+    (loop i32
+      (block
+        (br_table 0 1 1
+          (get_local 0)
+        )
+        (return
+          (i32.const 0)
+        )
+      )
+      (i32.const 1)
+    )
+  )
 )
 
 (assert_return (invoke "type-i32"))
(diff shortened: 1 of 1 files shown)


test/core/br_table.wast
```diff
@@ -1203,6 +1203,20 @@
       )
     )
   )
+
+  (func (export "nested-br_table-loop-block") (param i32) (result i32)
+    (loop i32
+      (block
+        (br_table 0 1 1
+          (get_local 0)
```
rossberg-chromium: Nit: no need for line break here

test/core/br_table.wast
```diff
@@ -1203,6 +1203,20 @@
...
+        )
+        (return
+          (i32.const 0)
```
rossberg-chromium: ...or here

rossberg-chromium: On reflection, this return is unreachable, so can just be dropped altogether.

jgravelle-google:
Done in https://github.com/WebAssembly/spec/pull/456/commits/68532e23f4507f959837d3ba9145acb22d9f0ebb

rossberg-chromium:
Thanks!

PR #459, opened 2017-04-13
Memory trap more tests

Cellule:
Add memory trapping tests for lower bound as well and test load as well as stores.
After I found a bug in chakra, I realized the spec test didn't test mem store with negative indexes.
Also, there were no tests for loads either.
Since their implementation can differ, we should test as much as we can just to be on the safe side.

I used a simple script to automatically generate all the tests.
I also changed the store to attempt to store 0 instead of a random value (mostly for line length)


rossberg-chromium:
Nice, thanks!

PR #464, opened 2017-04-18
Fix import-after-global.fail test

AndrewScheidecker:
The test was missing a global variable initializer expression, and so would fail regardless of whether the import-after-global syntax error is detected correctly.

### DIFF
--- a/test/core/import-after-global.fail.wast
+++ b/test/core/import-after-global.fail.wast
@@ -1,1 +1,1 @@
-(module (global i64) (import "" "" (table 0 anyfunc)))
+(module (global i64 (i64.const 0)) (import "" "" (table 0 anyfunc)))


binji:
Ah, interesting. I was wondering why this failed properly in wabt, but it looks like the parser allows for a const_expr to be 0 or more instructions -- this is caught in the validator, but that happens later. The import ordering check happens in the parser. Seems the spec interpreter does the same thing.

jfbastien:
Is that the right behavior (allowing empty global init)? The design repo wording seems to allow is, which I find surprising:
https://github.com/WebAssembly/design/blob/master/Modules.md#initializer-expression

AndrewScheidecker:
Regardless of which errors happens first, the problem is that an implementation that incorrectly accepts an import-after-global-definition may still reject this file because of the empty global initializer, and it's easy to misinterpret that as the implementation passing the test.

binji:
Agreed.

jfbastien:
> Regardless of which errors happens first, the problem is that an implementation that incorrectly accepts an import-after-global-definition may still reject this file because of the empty global initializer, and it's easy to misinterpret that as the implementation passing the test.

Agreed. I'm wondering if "empty global initializer" is also a bug, which we should fix separately and add a test for.

rossberg-chromium:
LGTM, thanks!

rossberg-chromium:
As @binji says, initializer sequences with the wrong number of instructions are validation bugs, mostly for technical reasons (it is the simplest and most modular / future-compatible way to spec the restriction).

See PR #465 for suggested tests.

PR #467, opened 2017-04-26
[spec] Execution: conventions, runtime, instructions

rossberg-chromium:
First stab at of the execution semantics. Describes conventions, runtime structures, and the semantics of instructions. Still missing: the description of the numeric operations and the instantiation/linking of modules. Various other tweaks included.

Preview at https://webassembly.github.io/spec/

### DIFF
--- a/document/execution/instructions.rst
+++ b/document/execution/instructions.rst
@@ -22,7 +22,7 @@ Numeric Instructions
 1. Push the value :math:`t.\CONST~c` to the stack.
 
 .. note::
-   No formal reduction rule is needed for this instruction, since |CONST| instructions coincide with :ref:`values <syntax-val>`.
+   No formal reduction rule is required for this instruction, since |CONST| instructions coincide with :ref:`values <syntax-val>`.
 
 
 .. _exec-unop:
@@ -172,7 +172,7 @@ Parametric Instructions
 
 .. math::
    \begin{array}{lcl@{\qquad}l}
-   \val~\DROP &\stepto& \epsilon
+   \val~~\DROP &\stepto& \epsilon
    \end{array}
 
 
@@ -183,7 +183,7 @@ Parametric Instructions
 
 1. Assert: due to :ref:`validation <valid-select>`, a value :ref:`value type <syntax-valtype>` |I32| is on the top of the stack.
 
-2. Pop the value :math:`\I32.\CONST~n` from the stack.
+2. Pop the value :math:`\I32.\CONST~c` from the stack.
 
 3. Assert: due to :ref:`validation <valid-select>`, two more values (of the same :ref:`value type <syntax-valtype>`) are on the top of the stack.
 
@@ -191,7 +191,7 @@ Parametric Instructions
 
 5. Pop the value :math:`\val_1` from the stack.
 
-6. If :math:`n` is not :math:`0`, then:
+6. If :math:`c` is not :math:`0`, then:
 
    a. Push the value :math:`\val_1` back to the stack.
 
@@ -201,10 +201,10 @@ Parametric Instructions
 
 .. math::
    \begin{array}{lcl@{\qquad}l}
-   \val_1~\val_2~(\I32\K{.}\CONST~n)~\SELECT &\stepto& \val_1
-     & (\mbox{if}~n \neq 0) \\
-   \val_1~\val_2~(\I32\K{.}\CONST~n)~\SELECT &\stepto& \val_2
-     & (\mbox{if}~n = 0) \\
+   \val_1~\val_2~(\I32\K{.}\CONST~c)~\SELECT &\stepto& \val_1
+     & (\mbox{if}~c \neq 0) \\
+   \val_1~\val_2~(\I32\K{.}\CONST~c)~\SELECT &\stepto& \val_2
+     & (\mbox{if}~c = 0) \\
    \end{array}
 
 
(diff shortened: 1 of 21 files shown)


document/execution/runtime.rst
```diff
@@ -0,0 +1,535 @@
...
+.. index:: ! value, constant
+   pair: abstract syntax; value
+
+Values
+~~~~~~
+
+WebAssembly computations manipulate *values* of the four basic :ref:`value types <syntax-valtype>`: :ref:`integers <syntax-int>` and :ref:`floating-point data <syntax-float>` of 32 or 64 bit width each, respectively.
+
+In most places of the semantics, values of different types can occur.
+In order to avoid ambiguities, values are therefor represented with an abstract syntax that makes their type explicit.
```
binji: sp: therefore

document/execution/numerics.rst
```diff
@@ -0,0 +1,54 @@
...
+.. _aux-extend:
+.. _aux-wrap:
+
+Auxiliary Operations
+~~~~~~~~~~~~~~~~~~~~
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \bytes_N(i) &=& \epsilon & (N = 0 \wedge i = 0) \\
+   \bytes_N(i) &=& b~\bytes_{N-8}(j) & (N \geq 8 \wedge i = 8\cdot j + b) \\
```
binji: Should this be `i = 2^8\cdot j + b`?

document/execution/instructions.rst
```diff
@@ -0,0 +1,1053 @@
...
+   \begin{array}{l}
+   \begin{array}{lcl@{\qquad}l}
+   \val^n~(\INVOKE~a) &\stepto& \FRAME_m\{F\}~\BLOCK~[t_2^m]~\instr^\ast~\END~\END
+   \end{array}
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\mbox{if} & S.\FUNCS[a] = f \\
+     \wedge & f.\CODE = \{ \TYPE~x, \LOCALS~t^\ast, \BODY~\instr^\ast~\END \} \\
+     \wedge & f.\MODULE.\TYPES[x] = [t_1^n] \to [t_2^m] \\
+     \wedge & F = \{ \MODULE~f.\MODULE, ~\LOCALS~\val^n~(t.\CONST~0)^\ast \})
```
binji: Seems a bit confusing to use * here since I normally assume they are all unrelated.

document/execution/instructions.rst
```diff
@@ -0,0 +1,1053 @@
...
+
+10. If :math:`i` is not smaller than the length of :math:`\X{tab}.\ELEM`, then:
+
+    a. Trap.
+
+11. If :math:`\X{tab}.\ELEM[i]` is uninitialized, then:
+
+    a. Trap.
+
+12. Let :math:`a` be the :ref:`function address <syntax-funcaddr>` :math:`\X{tab}.\ELEM[i]`.
```
binji: `a` is redefined here (previously defined at step 3). Maybe worth using a new letter?

document/execution/instructions.rst
rossberg-chromium: Well, this pattern of implicit "mapping" over a sequence actually occurs in many places, not just this chapter. That is, when there is a sequence x*, then a sequence (...x...)* is meant to correlate. I added an explanation to the syntactic conventions section.

binji: Ah, I see. The mapping normally tracks for me, but `t* -> (t.const 0)*` didn't as much. In my mind using another symbol would make it clearer (e.g. `t^k -> (t.const 0)^k`), but I'm no formalist so I don't know if that makes any sense. :-)

rossberg-chromium: Makes sense, done.

lukewagner:
Really nice!  lgtm with a few small nits/suggestions:
* conventions.html#textual-notation - Maybe want to rename this section to "Prose conventions" because when I first read "Textual Notation" and have the paragraph say that these are the conventions followed by option (1) above.
* runtime.html#addresses - Might be useful to add note contrasting "addresses" to the indices in syntax/modules.html#syntax-index.  Particularly that indices are static and module-local whereas addreses index a dynamic and index the store.
* t.load/store - I think bounds checking (in prose and formal) needs to be strictly <, not <=.  Also, it might be nice to hyperlink the colon in `data[ea : N]` and others to conventions.html#auxiliary-notation
* call_indirect formal conjuncts' column alignment seems messed up
* instructions.html#instruction-sequences - It'd be nice to prefix Entering/Exiting prose paragraphs with a note that these subroutines are "called" by other instructions.  Similarly for "Invocation" in instructions.html#function-calls.

rossberg-chromium:
@lukewagner, thanks! Addressed comments, except the hyperlinking suggestion (put on my todo list of xref clean-ups/additions) and the bounds checking. I think the bounds check is correct as is -- note that it already includes the width of the data, and reading, say, 2 bytes at address 8 in a 10 byte address space should be fine.

lukewagner:
@rossberg-chromium D'oh, you're right; I missed the data width.  Btw, have you considered any way to visually separate the end of the prose notation from the beginning of the formal rules?  ([e.g.](http://web.archive.org/web/20070629104332im_/http://blood.lith.com/bludrule.gif))  They sortof run together atm.

rossberg-chromium:
@lukewagner, yeah, I don't like that running together either, and I already tried an hline (which is the only thing markdown offers) but it made the grouping really confusing visually (rules looked as if they belonged to the next section title). Tweaking vertical spacing might be better, but markdown makes it impossible to insert less than a full empty line + paragraph space (so similar effect). But I just realised that I could play with putting the space in the formulas themselves. Are you okay with making that a todo?

domenic:
FWIW in the HTML spec we use `<hr>` but style it as extra vertical whitespace. I guess that would only work here if you aren't already using `<hr>` for horizontal lines.

rossberg-chromium:
@domenic, whew, dirty. :)

lukewagner:
> Are you okay with making that a todo?

Oh of course; already lgtm, just a broader question.

PR #471, opened 2017-05-11
[spec] Text format

rossberg-chromium:
This PR specifies the text format, based on earlier discussion between @binji, @lukewagner, @sunfishcode, and myself. Preview at http://webassembly.github.io/spec/

The changes relative to the .wast format currently implemented in the interpreter and other tools are the following.

Removals:

- some of the more baroque forms of sugar for `if`
- binary module bodies
- anything script related (assertions, invokes, etc)
- `infinity` as a secondary spelling for float `inf`
- (we had discussed removing the optional module name as well, but in the light of design PR #1055, we probably want to keep it?)

Additions:

- \u{...} escapes in strings (see below)
- more than just one inline export (that is, you can write (func $f (export "f1") (export "f2") ...), closing a gap in the syntax), and it combines with import
- the toplevel (module ...) is optional

Changes:

I took the liberty to propose one breaking change to make the syntax forward compatible with some of the future extensions that have been discussed:

- non-empty block signatures must now be written (result i32), in order to generalise cleanly to function signatures

Unicode:

- the lexical syntax is defined in terms of Unicode characters (i.e., code points)
- comments and strings may contain mostly arbitrary Unicode, the rest stays within ASCII
- in strings, a Unicode character denotes its UTF-8 encoding
- in strings, Unicode characters can be given explicitly with \u{...} notation
- .wat files are assumed to be encoded in UTF-8

The PR includes changes to the interpreter implementing the Unicode support (but not yet the other changes).

Misc Remarks:

- formatting characters: currently only the minimum set of formatting characters are allowed as white space (\t, \n, \r); we could include more, e.g. the whole set of ASCII "format effectors" (\b, \v, \f), but you quickly get into a lot of Unicode complexity if you want to go further than that

- Unicode in comments: similarly, in order to avoid getting into Unicode specifics, any legal code point is currently allowed in comments; should we be more restrictive?

- binary module bodies: they would seem pretty unusual for a "text" format, so are not included. Is there a reason to keep them?

- abbreviations: to avoid combinatorial complexity in defining the AST to map on, most syntactic sugar is specified in the form of "abbreviations", simple rewritings into the core syntax

- inline function signatures: I tried to come up with a decent way to describe their rewriting into type indices (and the potential insertion of new type definitions) in terms of rules, but ultimately gave up; it's too cumbersome to express succinctly; so this is the one part that is left partially informal (though hopefully still unambiguous)

- formatting: many of the rules do not currently fit the page width; I left them as is for now, and plan to clean up layout issues once the spec is complete, probably tweaking some layout parameters as well

- PDF: currently doesn't build; due to various limitations of MathJax (such as the inability to use packages) I had to resort to some hacks for special characters that apparently don't work in proper LaTeX; will fix later

- tests: lots of stuff we could write tests for, e.g. regarding the Unicode support...


jfbastien:
On Unicode comments: are U+2028 and U+2029 allowed?

rossberg-chromium:
@jfbastien, yes, currently, any code point is allowed, with no special interpretation. The only ones with specific meaning are ; ( ) and \n. (In particular, U+2028/9 would not end a line comment, but nor does ASCII \r, \f, or \v.)

jfbastien:
Gotcha, that makes sense. I'm not sure we want to restrict more than what you have, silly Unicode-isms are silly. Were we to try to fool-proof things there's other semicolons such as U+FF1B ；
❨ and more ❩ ❪ parens than ❫  ⦗ one can  ⦘  ﹙ keep track ﹚ ﹝ of  ﹞ （ because lol Unicode ）.

sunfishcode:
Concerning module names and https://github.com/WebAssembly/design/pull/1055/, the module name in the name section has no semantic effect, and so seems like it could differ from the module name used for linking. I'd suggest this is an interesting enough corner case that it's worth testing, which would mean we'd want distinct syntax for the name-section module name.

Are the baroque forms of sugar mentioned above being removed from the .wast format too? Similarly, are the various Additions and Changes being made to the .wast format too?

What is the spec.exec.1 branch? This PR contains some changes that look like changes already in master, so it's not clear what the specific changes are here.

rossberg-chromium:
@sunfishcode, yes, I plan to make the changes to the interpreter, since .wast should remain a strict super set of .wat that differs only in terms of the additional script constructs.

Branch spec.exec.1 corresponds to PR #467 -- I used that as a baseline here, since there were too many dependent changes.

rossberg-chromium:
One item I forgot: this also doesn't include the binary module body syntax that we use for some tests. Should it?

jfbastien:
> One item I forgot initially: this also doesn't include the binary module body syntax that we use for some tests. Should it?

Can the proposed format generate all interesting invalid inputs without this?

Additionally, can the proposed format generate equivalent modules (say, non-canonical LEBs)?

rossberg-chromium:
@jfbastien, no, without the direct binary notation something like LEB isn't even a thing in the text format. I guess the question we need to answer is whether it is the purpose of the official text format (as opposed to .wast) to enable expressing such tests.

jfbastien:
> @jfbastien, no, without the direct binary notation something like LEB isn't even a thing in the text format. I guess the question we need to answer is whether it is the purpose of the official text format (as opposed to .wast) to enable expressing such tests.

Yes, I think that's the right question to ask. I think we can move forward with your proposal, without this addition, and then add it later if we decide we need it. Correct?

rossberg-chromium:
@jfbastien, correct.

binji:
re: binary module bodies. I think it's pretty natural for an assembler to allow writing binary data directly anywhere. It could be useful for allowing experimenting with future extensions without having to update the parser. The difficulty will be validating a module written this way, however. I agree its probably best to move forward without it for now.

document/math.def
```diff
@@ -377,6 +518,31 @@
...
+
+
+.. Context, textual
+
+.. |TYPEDEFS| mathdef:: \K{typedefs}
+
+
+.. Contexts, meta functions
+
+.. |idfresh| mathdef:: ~\xref{text/conventions}{text-id-fresh}{\mbox{fresh}}
```
binji: This is currently defined in values, not conventions.

document/text/values.rst
```diff
@@ -0,0 +1,260 @@
...
+     s{:}\Tsign~\text{0x}~p{:}\Thexnum~\text{.}~q{:}\Thexfrac
+       &\Rightarrow& s\cdot(p+q) \\ &&|&
+     s{:}\Tsign~\text{0x}~p{:}\Thexnum~(\text{P}~|~\text{p})~t{:}\Tsign~e{:}\Tnum
+       &\Rightarrow& s\cdot p\cdot 2^{t\cdot e} \\ &&|&
+     s{:}\Tsign~\text{0x}~p{:}\Thexnum~\text{.}~q{:}\Thexfrac~(\text{P}~|~\text{p})~t{:}\Tsign~e{:}\Tnum
+       &\Rightarrow& s\cdot(p+q)\cdot 2^{t\cdot e} \\
+   \production{floating-point value} & \TfX{N} &::=&
+     z{:}\Tfloat &\Rightarrow& b^\ast & (\ieee_N(z) = b^\ast) \\ &&|&
+     z{:}\Thexfloat &\Rightarrow& b^\ast & (\ieee_N(z) = b^\ast) \\ &&|&
+     s{:}\Tsign~\text{inf} &\Rightarrow& b^\ast & (\ieeeinf_N(s) = b^\ast) \\ &&|&
```
rossberg-chromium: Ah, right, forgot to list that as a removal. It seems a bit redundant to have two spellings. Added to PR description.

document/math.def
rossberg-chromium: Fixed.

document/text/lexical.rst
```diff
@@ -0,0 +1,100 @@
...
+Characters
+~~~~~~~~~~
+
+The text format assigns meaning to *source text*, which consists of a sequence of *characters*.
+Characters are assumed to be represented as valid `Unicode <http://www.unicode.org/versions/latest/>`_ *code points*.
+
+.. math::
+   \begin{array}{llll}
+   \production{character} & \Tchar &::=&
+     \unicode{00} ~|~ \dots ~|~ \unicode{D800} ~|~ \unicode{E000} ~|~ \dots ~|~ \unicode{10FFFF} \\
```
rossberg-chromium: Indeed!

AndrewScheidecker:
As long as we're making breaking changes to the text format, can we get rid of some syntax sugar?

I really don't like the syntax for allowing inline import/export tags on declarations. It's a pretty substantial complication of the syntax, and I don't see a clear benefit.

I also don't like the syntax for declaring table/memory with an inline segment that implies the size for the same reason, though to a lesser degree.

rossberg-chromium:
@AndrewScheidecker, I can report from my experience with writing thousands of tests that there is a substantial benefit. For example, for exports it's the difference between writing out every function name only once instead of three times in the definition of a module. ;-)

document/text/values.rst
binji: sgtm

binji:
I argued against a lot of the syntax sugar as well. Keeping it seems counter to the premise that WebAssembly is a format that is meant to be generated, not written by hand. Certainly more people will be reading WebAssembly modules than writing them. A better argument for keeping it might be that it is easier to read.

kanaka:
I'm not arguing for or against any specific syntax sugar, but 10 developer hours spent to support easier to read syntax in the textual format might easily save 5 minutes debugging across many thousands of developers in the future. So if it actually does add some benefit to readability (even a little) it still might be worth it at scale. For example, I've already saved a non-trivial amount of time by switching from wasm2wast to wasm-dis due to the friendlier syntactic sugar in the latter. Also, optional syntactic sugar doesn't mean that tools generating wast have to support it, just that assemblers need to and disassemblers can choose to do so right?

jfbastien:
As mentioned readability matters, but we should also consider hand-editing of auto-generated wasm, e.g. for in-browser debugging. It's not that we want to make it easy to generate entire modules, but that we want editing of small parts of a module to be easy.

binji:
@jfbastien fair point.

@kanaka I assume you're referring to the difference between flat (wabt/wasm2wast) and folded (binaryen/wasm-dis) primarily? The trouble with doing folded by default is that it's not possible to represent all modules using the folded format. I believe binaryen works around this by generating output that has the same behavior, but is not actually the same.

Anyway, @jfbastien is right, probably best to continue the conversation in a separate issue.

kanaka:
@binji that is nice, but I actually had in mind the inline types, inline exports/imports, better mnemonic labelling of variables, etc. The type of things that give contextual information and don't force you to go hunting to other distant parts of the code to find out what exactly you're looking at.

rossberg-chromium:
I included the changes necessary to both interpreter and test suite to adjust to the listed grammar modifications. Also a few fixes to the spec, notably including #478 and clarification of the ability to combine inline import/export.

rossberg-chromium:
Anybody opposed to landing this? Anybody willing to review the PR? :)

document/text/conventions.rst
```diff
@@ -0,0 +1,140 @@
...
+Grammar
+~~~~~~~
+
+The following conventions are adopted in defining grammar rules for the text format.
+They mirror the conventions used for :ref:`abstract syntax <grammar>` and for the :ref:`binary format <binary-format>`.
+In order to distinguish symbols of the textual syntax from symbols of the abstract syntax, typewriter font is adopted for the former.
+
+* Terminal symbols are either literal strings of characters enclosed in quotes: :math:`\text{module}`;
+  or expressed as `Unicode <http://www.unicode.org/versions/latest/>`_ code points: :math:`\unicode{0A}`.
+  (All characters written literally are unambguously drawn from the `7-bit ASCII <http://webstore.ansi.org/RecordDetail.aspx?sku=INCITS+4-1986%5bR2012%5d>`_ subset of Unicode.)
```
lukewagner: *unambiguously

document/text/lexical.rst
```diff
@@ -0,0 +1,111 @@
...
+
+The character stream in the source text is divided, from left to right, into a sequence of *tokens*, as defined by the following grammar.
+
+.. math::
+   \begin{array}{llll}
+   \production{token} & \Ttoken &::=&
+     \Tkeyword ~|~ \TuN ~|~ \TsN ~|~ \TfN ~|~ \Tstring ~|~ \Tid ~|~
+     \text{(} ~|~ \text{)} ~|~ \Treserved \\
+   \production{keyword} & \Tkeyword &::=&
+     \mbox{(any terminal symbol in the grammar that is non of the above)} \\
```
lukewagner: *none

document/text/values.rst
```diff
@@ -0,0 +1,260 @@
+.. _text-value:
+.. index:: value
+   pair: text format; value
+
+Values
+------
+
+The grammar produtions in this section define *lexical syntax*,
```
lukewagner: *productions

lukewagner:
Nice work!  lgtm with a few nits above and here:
* lexical.html#characters : the Note seems a bit confusing to me given the preceding para just said all valid unicode code points are characters.  I *think* what's being said is that, across the entire set of rules which define the text format, it is noted to be the case that, outside of comments and string literals, only a subset of 7-bit ASCII characters are used?
* the abstract syntax of a floating constant is defined by values.html#floating-point to be a sequence of (IEEE754-interpreted) bytes however values.html#floating-point seems to produce reals with only a fuzzy note above that they get rounded; could you instead have an explicit `realBytes(...)` function applied to the reals to explicitly produce the bytes?
* types.html#table-types : perhaps add a Note that `elemtype` may be extended with other types in the future

document/text/conventions.rst
rossberg-chromium: Done.

document/text/lexical.rst
rossberg-chromium: Done.

document/text/values.rst
rossberg-chromium: Done.

rossberg-chromium:
On 25 May 2017 at 05:38, Luke Wagner <notifications@github.com> wrote:

> Nice work! lgtm with a few nits above and here:
>
>    - lexical.html#characters : the Note seems a bit confusing to me given
>    the preceding para just said all valid unicode code points are characters.
>    I *think* what's being said is that, across the entire set of rules
>    which define the text format, it is noted to be the case that, outside of
>    comments and string literals, only a subset of 7-bit ASCII characters are
>    used?
>
> Yes, that was the intention. Reworded.


>    - the abstract syntax of a floating constant is defined by
>    values.html#floating-point to be a sequence of (IEEE754-interpreted) bytes
>    however values.html#floating-point seems to produce reals with only a fuzzy
>    note above that they get rounded; could you instead have an explicit
>    realBytes(...) function applied to the reals to explicitly produce the
>    bytes?
>
> Right, that's the purpose of the `ieee` meta functions used in the
attribute expressions. I yet have to define those, though (see the todo),
which I plan to do when I get to the numeric ops.


>    - types.html#table-types : perhaps add a Note that elemtype may be
>    extended with other types in the future
>
> Added.

lukewagner:
@rossberg-chromium Ah, I see now; they all feed in to `fN`; I missed that before.

rossberg-chromium:
Landing with above LGTM and no objections.

PR #474, opened 2017-05-15
Add tests for invalid UTF-8 encoding in the text format

sunfishcode:
This adds a ".fail.wast" test file for each of the 179 invalid UTF-8 byte sequences created for the other UTF-8 tests. This provides some test coverage for validation on the text parsing side (because it matters for some consumers), however it's an awkwardly large number of files, and I don't know if it's worth the hassle. In any case, here's the PR in case anyone's interested.


rossberg-chromium:
I'm torn. The more tests the better, but it _is_ a large number of files. OTOH, once #471 lands, we probably want similar tests checking the UTF-8 encoding of the input itself (i.e., invalid UTF-8 occurring directly, not escaped, in either strings or comments). Perhaps a solution is to introduce a subdirectory for everything Unicode related?

AndrewScheidecker:
Couldn't the need for the .fail.wast files be eliminated by adding an `(assert_unparseable "...")` that embeds the unparseable modules in strings?

rossberg-chromium:
@AndrewScheidecker, ha, interesting idea. Let me think about that.

rossberg-chromium:
PR #475 landed, so these tests can now be written in just one file.

rossberg-chromium:
@sunfishcode, any chance you can adapt this PR?

sunfishcode:
I still do intend to do this, though it's not a high priority right now.

binji:
Pushed a change to refactor these. Hope you don't mind! :-)

rossberg:
When you merge, can you make sure to update the title/description?

sunfishcode:
@binji That looks like all that needs to be done here then. Thanks!

PR #475, opened 2017-05-15
[interpreter] Support quoted module definitions in .wast

rossberg-chromium:
As suggested by @AndrewScheidecker in [this comment](https://github.com/WebAssembly/spec/pull/474#issuecomment-301457926) on #474, this adds the ability to quote a module source (i.e., .wat file) in a string, so that syntax errors for the text format are deferred and thus can be tested in a way analogous to decoding errors for binary format:
```
(assert_malformed (module quote "(module func)"))
(assert_malformed (module binary "\00\00\00\00"))
```
(The PR also changes the syntax of binary modules to require a "binary" keyword, in order to disambiguate.)

This change would make separate `.fail.` tests unnecessary. It was an easy change to the interpreter, but I don't know how simple it would be for other tools. @binji, @kripken, @dschuff, any opinions?


binji:
Hm... I'm pretty sure wabt's parser and lexer are re-entrant. Lemme check. :-)

rossberg-chromium:
@binji, is reentrance actually necessary for this?

binji:
Ah, you're right. The whole point is to defer the module text parsing :-) Yes, this looks like it should be fine in wabt.

interpreter/text/parser.mly
```diff
@@ -160,10 +160,11 @@ let inline_type (c : context) ty at =
 %token CALL CALL_INDIRECT RETURN
 %token GET_LOCAL SET_LOCAL TEE_LOCAL GET_GLOBAL SET_GLOBAL
 %token LOAD STORE OFFSET_EQ_NAT ALIGN_EQ_NAT
-%token CONST UNARY BINARY COMPARE CONVERT
+%token CONST UNARY BINARY TEST COMPARE CONVERT
```
binji: Is `TEST` new?

rossberg-chromium: No, it was just missing from this list. It seems that it actually is unnecessary to declare a token like this when there already is a %token<...> declaration like below, but to be consistent I added this one.

rossberg-chromium:
If everybody is okay with this extension to the wast format then I'll go forward and convert all .fail. tests to quotes.

@binji, would you be okay with reviewing this PR?

rossberg-chromium:
@binji, friendly ping.

interpreter/host/js.ml
```diff
@@ -377,11 +378,12 @@ let of_command mods cmd =
     ":" ^ string_of_int cmd.at.left.line ^ "\n" ^
   match cmd.it with
   | Module (x_opt, def) ->
-    let m =
+    let rec unquote def =
       match def.it with
       | Textual m -> m
       | Encoded (_, bs) -> Decode.decode "binary" bs
-    in bind mods x_opt m;
+      | Quoted (_, s) -> unquote (Parse.string_to_module s)
```
binji: This means you could do a recursive quote, right? `(module quote "(module quote \"...`
It might be nice to disallow that.

rossberg-chromium: Yes. No real use case for that, it just turned out to be simpler and more regular in the interpreter to allow it than to rule it out. But I don't mind making it a syntax error. Will look into it.

rossberg-chromium:
Okay, disallowed extended module syntax in quotes (and also .wat files).

Converted all existing .fail. tests into uses of quotes (with some additional variations, because it's much easier now).

Landing.

PR #482, opened 2017-05-17
A^+ description changed

xor:
A^+ can not be empty, so the description as "...possibly empty sequence of iterations..." was changed to "...non empty sequence of iterations..."


rossberg-chromium:
Thanks!

PR #483, opened 2017-05-19
[spec] Implementation restrictions

rossberg-chromium:
This PR adds an appendix that lists the allowed (mostly numeric) limits that implementations may impose on WebAssembly programs. I also included lazy validation here, since it seems to fit into this context and there is no other obvious place to put it.


document/appendix/implementation.rst
```diff
@@ -0,0 +1,130 @@
...
+* the number of results in a :ref:`function type <syntax-functype>`
+* the number of :ref:`locals <syntax-local>` in a :ref:`function <syntax-func>`
+* the size of a :ref:`function <syntax-func>` body
+* the size of a :ref:`structured control instruction <syntax-instr-control>`
+* the number of :ref:`structured control instructions <syntax-instr-control>` in a :ref:`function <syntax-func>`
+* the nesting depth of :ref:`structured control instructions <syntax-instr-control>`
+* the length of an :ref:`element segment <syntax-elem>`
+* the length of a :ref:`data segment <syntax-data>`
+* the length of a :ref:`name <syntax-name>`
+* the range of :ref:`code points <syntax-codepoint>` in a :ref:`name <syntax-name>`
```
jfbastien: I'm not sure restricting codepoints makes sense. Can you drop from here and discuss in a separate issue?

document/appendix/implementation.rst
```diff
@@ -0,0 +1,130 @@
...
+...........
+
+For a module given in :ref:`text format <text>`, additional limitations may be imposed on the following dimensions:
+
+* the size of the :ref:`source text <source>`
+* the size of any syntactic element
+* the size of an individual :ref:`token <text-token>`
+* the nesting depth of :ref:`folded instructions <text-foldedinstr>`
+* the length of symbolic :ref:`identifiers <text-id>`
+* the range of literal :ref:`characters <text-char>` (code points) allowed in the :ref:`source text <source>`
```
jfbastien: ditto

document/appendix/implementation.rst
```diff
@@ -0,0 +1,130 @@
...
+Validation
+~~~~~~~~~~
+
+An implementation may defer :ref:`validation <valid>` of individual :ref:`functions <syntax-func>` until they are first :ref:`invoked <exec-invoke>`.
+
+If a function turns out to be invalid, then the invocation, and every consecutive call to the same function, results in a :ref:`trap <trap>`.
+
+.. note::
+   This is to allow implementations to use interpretation or just-in-time compilation for functions.
+   The function must still be fully validated before execution of its body begins.
```
jfbastien: @MikeHolman can you confirm this is what you want?

document/appendix/implementation.rst
```diff
@@ -0,0 +1,130 @@
...
+* the number of allocated :ref:`memory instances <syntax-meminst>`
+* the number of allocated :ref:`global instances <syntax-globalinst>`
+* the size of a :ref:`table instance <syntax-tableinst>`
+* the size of a :ref:`memory instance <syntax-meminst>`
+* the number of :ref:`frames <syntax-frame>` on the :ref:`stack <syntax-stack>`
+* the number of :ref:`labels <syntax-label>` on the :ref:`stack <syntax-stack>`
+* the number of :ref:`values <syntax-val>` on the :ref:`stack <syntax-stack>`
+
+If the runtime limits of an implementation are exceeded during execution of a computation,
+then it may terminate that computation by causing a trap or reporting an embedder-specific error to the invoking code.
```
jfbastien: Why trap here? Isn't error sufficient?

document/appendix/implementation.rst
```diff
@@ -0,0 +1,130 @@
...
+
+Some of the above limits may already be verified during instantiation, in which case an implementation may report exceedance in the same manner as for :ref:`syntactic limits <impl-syntax>`.
+
+.. note::
+   Concrete limits are usually not fixed but may be dependent on specifics, interdependent, vary over time, or depend on other implementation- or embedder-specific variables.
```
jfbastien: Add: error at runtime if too much memory is dirtied, or code pages exhausted, or random kill (oom kill, too much cpu, etc.).

document/appendix/implementation.rst
rossberg-chromium: This was the outcome of the discussion on #1016. It allows environments that don't understand Unicode to limit import/export names to ASCII in particular. I added a note explaining the intent.

document/appendix/implementation.rst
rossberg-chromium: Well, technically, a trap is the only runtime error Wasm currently has.
 
But more to the point, this doesn't say that an impl _has to_ manifest resource exhaustion as a trap, but based on previous discussions it should at least be a legal option, if not the preferred one. Remember that we even _required_ that initially for stack overflow, until we found out that that specific case is kind of weird in a JS embedding. Yet it might still be the best choice for other errors or other embeddings.

document/appendix/implementation.rst
rossberg-chromium: Generalised to "implementation- or embedder-specific situations or events". Is that broad enough? I intentionally avoided being more specific or even enumerating examples, since it seems futile trying to compile an open-ended list.

document/appendix/implementation.rst
jfbastien: There wasn't really a discussion or consensus though. I'm not sure restricting versus just ignoring is the right approach. I'd like to discuss this separately, to keep a record of how we came to the decision, and to be really explicit to folks involved. That way we won't revisit the question unless new information comes to light.

So I'd like to tackle this on a follow-up to this PR, which otherwise is pretty straightforward.

document/appendix/implementation.rst
jfbastien: Maybe I'm mistaken, but it seems to me that all traps which can occur are defined as part of individual operation semantics. Separately, we document where embedder-specific exceptions can occur, and these are totally separate from traps though the embedder can manifest them as the same-ish (with some way to differentiate).

Runtime limits aren't caused by specific operations. On the above logic I there don't think they should trap.

document/appendix/implementation.rst
rossberg-chromium: Can you elaborate on the alternatives you're thinking of? Note that you cannot ignore imports. Requiring an implementation to document its restrictions seems reasonable, but is independent of the specific item.

Anyway, let's have the discussion here.

document/appendix/implementation.rst
rossberg-chromium: Some of the above _are_ tied to specific instructions (e.g. calls), some aren't. Thus "trap or reporting an error". Note that a trap currently is the only way in the core semantics to abort execution. Whether and how different traps are distinguished or reported is up to the API spec or the embedder.

We could introduce the notion of "host trap" or something along these lines to the core spec to suggest a distinction, but it would still be indistinguishable from an ordinary trap as far as the spec itself is concerned. Maybe that purpose is better served by a note?

document/appendix/implementation.rst
jfbastien: I'd like to have the discussion in a separate issue, not here. It is desirable because sub-comments are hidden once a patch is updated, and harder to search for. A separate issue for just this calls out the discussion. Once discussed and consensus is clearly reached we won't revisit the discussion unless new information arises. Discussing here leaves open the possibility of re-discussing.

lukewagner: This does seem to have already been called out [quite explicitly](https://github.com/WebAssembly/design/pull/1016#issuecomment-286663852) so I think any further discussion would be revisiting what was already consensus on that issue.

document/appendix/implementation.rst
```diff
@@ -0,0 +1,136 @@
...
+* the size of a :ref:`structured control instruction <syntax-instr-control>`
+* the number of :ref:`structured control instructions <syntax-instr-control>` in a :ref:`function <syntax-func>`
+* the nesting depth of :ref:`structured control instructions <syntax-instr-control>`
+* the length of an :ref:`element segment <syntax-elem>`
+* the length of a :ref:`data segment <syntax-data>`
+* the length of a :ref:`name <syntax-name>`
+* the range of :ref:`code points <syntax-codepoint>` in a :ref:`name <syntax-name>`
+
+If the limits of an implementation are exceeded for a given module,
+then the implementation may reject the :ref:`instantiation <exec-instantiate>` of that module with an embedder-specific error.
```
lukewagner: Could you also add compilation (otherwise it might suggest that these limit errors can only be reported from `instantiate` not `compile`).

document/appendix/implementation.rst
MikeHolman: yup, this is what we want.

document/appendix/implementation.rst
jfbastien: It was a comment on an issue, with no effect on the PR's content.

I think you misunderstand what I'm trying to get at: within the standards process, we can revisit this issue as it stands. By separating it, and polling the group on it, I first expect the outcome to stay as Andreas proposes and I expect we will not revisit the issue for lack of new information (this is a requirement!).

Without separating out the issue we're sneaking it in here among otherwise unsurprising things. That's not the way I want to conduct this effort. I want to call out what *could* be contentious so that people who care clearly see it and can respond. Doing so further forces us to document our thought process.

So, I maintain that I'd like this to be separate.

lukewagner: It was a comment on the same logical issue, though, with eyes on it from everyone involved in the utf8 discussion, so it seemed sufficiently visible.

document/appendix/implementation.rst
rossberg-chromium: Done. Also added validation.

document/appendix/implementation.rst
rossberg-chromium: Ack, but this one is an informal note explaining possible motivation for possible restrictions. It doesn't really make sense to _prescribe_ a Why, that's not checkable or enforceable. Especially when the What itself already is intentionally vague. If we wanted to be less broad then we could do so by stating minimum bounds, but so far we decided against that.

document/appendix/implementation.rst
rossberg-chromium: Split out into separate issue: #488

Issue #453, opened 2017-03-31
[test] build.py fails to build some tests


kmiller68:
Whenever I run `./build.py -js my/favorite/directory` I get the following stderr output:

```
Building JS...
Error when compiling {} to JS: {} /Volumes/Data/wasm-spec/test/core/unwind.wast /Volumes/Data/wasm-spec/test/core/f32.wast:53.15-53.16: syntax error: unknown operator

Error when compiling {} to JS: {} /Volumes/Data/wasm-spec/test/core/unwind.wast /Volumes/Data/wasm-spec/test/core/f64.wast:53.15-53.16: syntax error: unknown operator

Error when compiling {} to JS: {} /Volumes/Data/wasm-spec/test/core/unwind.wast /Volumes/Data/wasm-spec/test/core/float_exprs.wast:49.15-49.16: syntax error: unknown operator

Error when compiling {} to JS: {} /Volumes/Data/wasm-spec/test/core/unwind.wast /Volumes/Data/wasm-spec/test/core/float_misc.wast:61.15-61.16: syntax error: unknown operator

Error when compiling {} to JS: {} /Volumes/Data/wasm-spec/test/core/unwind.wast Result: 2541550819304892160 : i64
Expect: 0 : i64
/Volumes/Data/wasm-spec/test/core/memory_trap.wast:185.1-185.69: assertion failure: wrong return values
```
Those tests don't appear in `my/favorite/directory` so I assume these errors are real.

rossberg-chromium:
@bnjbvr, any idea what could cause this? The tests themselves seem to be fine.

saambarati:
Yeah if we could get a fix for this it'd be great. 

Is there any kind of a temporary workaround to importing these tests?

bnjbvr:
Will take a look.

PR #486, opened 2017-05-23
Automatically recompile the spec interpreter before generating the tests

bnjbvr:
#453 is a case of bad UX: a developer would expect that `build.py` just generates the tests. What's hidden in there is that the test generator uses the wasm spec interpreter, so it needs to be up to date and compiled in the first place.

This PR makes that we recompile the spec interpreter before generating the tests, unless explicitly told not to (with `--dont-recompile`).

@rossberg-chromium, can you have a look, please?

### DIFF
--- a/test/build.py
+++ b/test/build.py
@@ -9,7 +9,8 @@ import shutil
 import multiprocessing as mp
 
 SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
-WASM_EXEC = os.path.join(SCRIPT_DIR, '..', 'interpreter', 'wasm')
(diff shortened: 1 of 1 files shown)


bnjbvr:
It seems all the checks have passed and @rossberg-chromium approved the patch, is this ready to be merged?

rossberg-chromium:
@bnjbvr, you should become a committer. ;)

jfbastien:
He's already in the https://github.com/orgs/WebAssembly/teams/workshop group, doesn't that allow him to merge?

lukewagner:
Only as of today :)

PR #487, opened 2017-05-24
[spec] Module instantiation & handling of host functions

rossberg-chromium:
Specify allocation and instantiation of modules. Also, incorporate host functions as an abstract notion; calls to them are specced as non-deterministic behaviour.


document/execution/instructions.rst
```diff
@@ -1039,24 +1035,75 @@ When the end of a funtion is reached without a jump (|RETURN|) or trap aborting
    \end{array}
 
 
+.. _exec-invoke-host:
+
+Host Functions
+..............
+
+Invoking a :ref:`host function <syntax-hostfunc>` has non-deterministic behaviour.
```
jfbastien: Can you use US spelling consistently? That's what we've been doing elsewhere.

binji:
@jfbastien It looks like he's pushed this version to https://webassembly.github.io/spec. See https://webassembly.github.io/spec/execution/modules.html, for example.

document/execution/instructions.rst
rossberg-chromium: Done. (For some reason this is one that I always get wrong.)

rossberg-chromium:
@jfbastien, @binji, yup, up on github.io. Sorry, forgot to mention it this time.

document/execution/instructions.rst
```diff
@@ -1051,24 +1047,75 @@ When the end of a funtion is reached without a jump (|RETURN|) or trap aborting
...
+.. _exec-invoke-host:
+
+Host Functions
+..............
+
+Invoking a :ref:`host function <syntax-hostfunc>` has non-deterministic behavior.
+It may either terminate with a :ref:`trap <trap>` or return regularly.
+However, the latter case, it is assumed that it consumes and produces the right number and types of WebAssembly :ref:`values <syntax-val>` on the stack,
+according to its type.
+A host function may also modify the :ref:`store <syntax-store>` when invoked.
```
jfbastien: Tables as well? Should we have a provision for Globals when that's possible too?

document/execution/runtime.rst
```diff
@@ -157,13 +158,18 @@ The module instance is used to resolve references to other non-local definitions
...
    \end{array}
 
 Function instances are immutable, and their identity is not observable by WebAssembly code.
 However, the :ref:`embedder <embedder>` might provide implicit or explicit means for distinguishing them.
 
-.. todo:: Host functions?
+A *host function* is a function expressed outside WebAssembly but passed to a :ref:`module <syntax-module>` as an :ref:`import <syntax-import>`.
+The definition and behavior of host functions are outside the scope of this specification.
+For the purpose of this specification, it is assumed that when :ref:`invoked <exec-invoke-host>`,
+a host function behaves non-deterministicaly.
```
jfbastien: "deterministically"

document/execution/instructions.rst
rossberg-chromium: The "store" already encompasses functions, memories, tables, and globals.

document/execution/runtime.rst
rossberg-chromium: Fixed.

lukewagner:
Nice job this all together!  lgtm with some nits:
* In execution/modules.html#external-typing, I think 'store' should link to the 'store' in execution/runtime.html, not in instructions.html.
* In execution/modules.html#alloc-func and #host-functions, in the last lines where `S'` is defined, I think you mean `funcinst` instead of `tableinst`
* In execution/modules.html#alloc-module, I'm not sure that the way `externval`<sup>`*`</sup><sub>`ex`</sub> is defined fully specifies its order.  IIUC, it's saying that, e.g., if you filter out the `func` external values, that they'll be equal to a given list of `funcaddr`s, but it doesn't say where those `func` external values were in the unfiltered original list.
* In the formal rule for Instantiation, I think you can remove the S from S ⊢ module : externtype<sup>n</sup>; the S made me look for a rule inside the execution semantics before realizing this was just the top-level validation rule for modules (which takes no store or context).
* Maybe pre-existing, but in execution/runtime.html#function-instances, could the field `module` of a function instance be named `instance`?  Otherwise, I was confused about whether we were referring to a static module or an instance of a module.
* Pre-existing, but I was wondering: could the syntactic definition of functions store the type directly rather than by type index? I think the exact index used to matter b/c of exact matching of types by index but now that types are matched structurally I think the exact type index used isn't significant.

Also, a question on JS function identity: So iiuc, the JS-wasm binding spec would say: when a function is imported which is not an exported wasm function (and thus does not already have a `funcaddr`), `allochostfunc` is called to allocate a `funcaddr` which is then used for the `externval` of the import.  So then I think what we can say is that the JS embedding behaves as if every `funcaddr` was immediately given a unique JS function object and any time a wasm function (defined by `funcaddr`) flowed out to JS (via export object or `Table.prototype.get`) that `funcaddr`'s unique JS function object was produced.  And the same goes for all the other kinds of objects w.r.t their `*addr`'s.

If that's all right, then I think the note on identity currently in execution/runtime.html#function-instances doesn't belong there (since the instance itself is just a tuple with no inherent notion of identity; it's a value) but rather next to the definition of Addresses in execution/runtime.html#addresses.  Could you expand on this note to explain that host embeddings may choose to assign identity to (func, memory, ...) instances with identities corresponding to unique addresses?

rossberg-chromium:
On 9 June 2017 at 22:23, Luke Wagner <notifications@github.com> wrote:

> Nice job this all together! lgtm with some nits:
>
>    - In execution/modules.html#external-typing, I think 'store' should
>    link to the 'store' in execution/runtime.html, not in instructions.html.
>
> Fixed.

>
>    - In execution/modules.html#alloc-func and #host-functions, in the
>    last lines where S' is defined, I think you mean funcinst instead of
>    tableinst
>
> Fixed.

>
>    - In execution/modules.html#alloc-module, I'm not sure that the way
>    externval*ex is defined fully specifies its order. IIUC, it's saying
>    that, e.g., if you filter out the func external values, that they'll
>    be equal to a given list of funcaddrs, but it doesn't say where those
>    func external values were in the unfiltered original list.
>
> The overall order is induced by the definition of exportinst*, which must
contain the externval*_ex in the same order as the export*.


>
>    - In the formal rule for Instantiation, I think you can remove the S
>    from S ⊢ module : externtypen; the S made me look for a rule inside
>    the execution semantics before realizing this was just the top-level
>    validation rule for modules (which takes no store or context).
>
> Yep, done.

>
>    - Maybe pre-existing, but in execution/runtime.html#function-instances,
>    could the field module of a function instance be named instance?
>    Otherwise, I was confused about whether we were referring to a static
>    module or an instance of a module.
>
> Yeah, I can see that. But currently all fields containing "instances" (of
modules, tables, functions, etc) are named after the kind of entity. If we
want to make the static/dynamic distinction more explicit then it should be
with a naming scheme that works for all of them. I'll try to think of
something.


>    - Pre-existing, but I was wondering: could the syntactic definition of
>    functions store the type directly rather than by type index? I think the
>    exact index used to matter b/c of exact matching of types by index but now
>    that types are matched structurally I think the exact type index used isn't
>    significant.
>
> That would indeed simplify some rules, but would create much more
complication elsewhere. In particular, the AST then would have no way to
represent a module like (module (func (type 1))), which is rejected only by
validation. Effectively, it would require folding parts of validation into
both decoding & parsing.

Also, a question on JS function identity: So iiuc, the JS-wasm binding spec
> would say: when a function is imported which is not an exported wasm
> function (and thus does not already have a funcaddr), allochostfunc is
> called to allocate a funcaddr which is then used for the externval of the
> import. So then I think what we can say is that the JS embedding behaves as
> if every funcaddr was immediately given a unique JS function object and
> any time a wasm function (defined by funcaddr) flowed out to JS (via
> export object or Table.prototype.get) that funcaddr's unique JS function
> object was produced. And the same goes for all the other kinds of objects
> w.r.t their *addr's.
>
Yes, at least currently that would be indistinguishable from the semantics
given in JS.md. However, I think that isn't necessarily so under certain
extensions. If, for example, there was a way to instantiate/link Wasm
modules from Wasm directly, without their mutual ex/imports entering JS
land, then a function may flow through multiple Wasm modules before being
materialised in JS. At that point you'd need to check whether it already
has been materialised through _another_ module. That would seem to require
a persistent global map as opposed to just an instantiation-time per-module
map as of now, which might be rather expensive.

If that's all right, then I think the note on identity currently in
> execution/runtime.html#function-instances doesn't belong there (since the
> instance itself is just a tuple with no inherent notion of identity; it's a
> value) but rather next to the definition of Addresses in
> execution/runtime.html#addresses. Could you expand on this note to
> explain that host embeddings may choose to assign identity to (func,
> memory, ...) instances with identities corresponding to unique addresses?
>
 Right. Added that to Addresses. Turned the previous comment in Function
Instances to a note referencing that.

lukewagner:
> If, for example, there was a way to instantiate/link Wasm modules from Wasm directly, without their mutual ex/imports entering JS land, then a function may flow through multiple Wasm modules before being materialised in JS. At that point you'd need to check whether it already has been materialised through _another_ module.

Hmm, I don't see how that would produce any JS-observable difference from if JS has linked those same wasm modules in the same configuration: the `funcaddr` would in all cases refer to a `funcinst` whose `moduleinst` was the original instance that defined the function, regardless of the path it took to reach JS.

rossberg-chromium:
On 21 June 2017 at 01:31, Luke Wagner <notifications@github.com> wrote:

> If, for example, there was a way to instantiate/link Wasm modules from
> Wasm directly, without their mutual ex/imports entering JS land, then a
> function may flow through multiple Wasm modules before being materialised
> in JS. At that point you'd need to check whether it already has been
> materialised through *another* module.
>
> Hmm, I don't see how that would produce any JS-observable difference from
> if JS has linked those same wasm modules in the same configuration: the
> funcaddr would in all cases refer to a funcinst whose moduleinst was the
> original instance that defined the function, regardless of the path it took
> to reach JS.
>
In the current setup it's a coincidental invariant that every exported
funcinst always gets wrapped into a JS function immediately at
instantiation time. And this relation is then merely maintained when
flowing through other modules. But if instantiation and wrapping are
decoupled, that invariant is gone.

Imagine a module A exporting f, and modules B and C importing and
rexporting it. With the current API, you have to instantiate A from JS, and
that will create a JS wrapper for f. When instantiating B and C, also from
JS, the API then makes sure that the reexported f maintains the identity of
the wrapper. In particular, it is sufficient to maintain a temporary
mapping from function instances to wrappers in the definition of
instantiation, like the API definition currently does.

Now imagine there was some kind of module loader in Wasm itself, and A, B,
and C got instantiated there. No wrappers are constructed. Also imagine
some API for querying that loader from JavaScript, so that you can get
access to B and C _later_. That API would need to materialise wrappers of
functions at that point. When you query C separately from and later than B,
then the only way for it to know that f has already been wrapped is by
somehow maintaining a _global_ mapping from function instances to wrappers
in the API semantics.

In the implementation, that could presumably be achieved by having each
exported funcinst point to its optional wrapper. That might not be too bad
in terms of cost. An actual global map probably would be, though.

lukewagner:
Alternatively, each wasm instance object could maintain a lazily-populated |func-index -> JS wrapper| map.  (That's what we do now in FF so that we don't have to eagerly create JS function wrappers for every function in a `Table`, most of which will never be used.  Because this map is eagerly populated with exported functions, `Table.get` will correctly return the right identity.)

Given this, in the pure-wasm-module-loader scenario you describe, if you have a `funcinst` `f` that you want to reflect in JS, you simply lookup `f.code` (which you'd represent as a function index) in `f.module`.

PR #488, opened 2017-05-29
[spec] Allow impls to limit code point range

rossberg-chromium:
As discussed on https://github.com/WebAssembly/design/pull/1016, make it legal for implementations in environments that do not understand (all of) Unicode to only support smaller character subsets.

### DIFF
--- a/document/appendix/implementation.rst
+++ b/document/appendix/implementation.rst
@@ -51,9 +51,16 @@ An implementation may impose restrictions on the following dimensions of a modul
 * the length of an :ref:`element segment <syntax-elem>`
 * the length of a :ref:`data segment <syntax-data>`
 * the length of a :ref:`name <syntax-name>`
+* the range of :ref:`code points <syntax-codepoint>` in a :ref:`name <syntax-name>`
 
 If the limits of an implementation are exceeded for a given module,
 then the implementation may reject the :ref:`validation <valid>`, compilation, or :ref:`instantiation <exec-instantiate>` of that module with an embedder-specific error.
+
+.. note::
+   The last item allows embedders that operate in limited environments without support for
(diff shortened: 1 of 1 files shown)


tabatkins:
As long as it's a subset (such as ASCII) and not a different charset entirely (like Shift-JIS), yeah, no problem here.

domenic:
I don't really understand the spec text for this restriction, but it seems like other people do, so maybe it's fine.

Reading the other threads, it seems like the actual interpretation is that implementations are free to reject incoming source bytes if certain bits in those bytes are set to certain values? E.g. an implementation is free to reject incoming source bytes if the most-significant-bit is set, effectively only allowing ASCII names?

It would be a lot clearer to me if things were stated that way, but as I said, it seems like others aren't having this comprehension problem, so maybe it's fine.

RyanLamansky:
@tabatkins @domenic The terms "code point" and "common subsets" are clear enough to me that we're still talking about Unicode values, not binary bytes/bits.  It might be helpful to be more explicit about this distinction, though.

domenic:
How are we talking about Unicode values? Isn't this spec discussing implementation-specific limitations on the inputs, which are definitely bytes?

RyanLamansky:
@domenic I have to look at the whole file; not shown in he GitHub "Files changed" feature are the section headings, which add more context to the changes.

domenic:
Right, I guess I don't understand what the first change applies to, i.e. the "Syntactic Limits" heading.

rossberg-chromium:
@domenic, it's described in terms of the abstract syntax, which defines names as sequences of Unicode code points. That makes it independent of the concrete input format (e.g. binary or text format).

rossberg-chromium:
Seems like there is approval and no objections, so I'll merge.

PR #491, opened 2017-06-02
Test Harness wasm-module-builder.js allow long sections

Cellule:
Spreading the section into arguments can become too long for some engines.


test/harness/wasm-module-builder.js
```diff
@@ -74,7 +74,9 @@ class Binary extends Array {
     // Emit section length.
     this.emit_u32v(section.length);
     // Copy the temporary buffer.
-    this.push(...section);
+    for (let i = 0; i < section.length; ++i) {
```
rossberg-chromium: Nit: use a for-of loop instead?

PR #497, opened 2017-06-06
Fix result syntax

rossberg-chromium:
Fix spec deviations:

- allow multiple results in function defs
- disallow multiple results in block sig

This is consistent with the spec document and the binary format.

### DIFF
--- a/interpreter/Makefile
+++ b/interpreter/Makefile
@@ -26,12 +26,13 @@ JS =		# set to JS shell command to run JS tests
 
 # Main targets
 
-.PHONY:		unopt opt libunopt libopt all land zip
+.PHONY:		default opt unopt libopt libunopt all land zip
 
-unopt:		$(UNOPT)
+default:	opt
 opt:		$(OPT)
-libunopt:	_build/$(LIB).cmo
+unopt:		$(UNOPT)
 libopt:		_build/$(LIB).cmx
+libunopt:	_build/$(LIB).cmo
 all:		unopt opt libunopt libopt test
 land:		all $(WINMAKE)
 zip: 		$(ZIP)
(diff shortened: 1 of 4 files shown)


binji:
Probably worth adding some parse tests for this. Is that possible without trapping in the validator?

rossberg-chromium:
Added tests.

AndrewScheidecker:
What's the motivation for allowing multiple `result` declarations in the syntax? Isn't it sufficient to support multiple results with multiple value types per result declaration, e.g. `(result f32 f64)`?

rossberg-chromium:
Mainly syntactic consistency. All declarations define the singular form as primary syntax, and merged variants are merely syntactic sugar. Results don't currently allow names, so it could have been done differently in this particular case, but what would such deviation buy?

AndrewScheidecker:
The fact that parameters may be named vs unnamed makes both parameter declaration forms useful in different cases, but I can't think of a case where you would want to use the multiple declaration form for results. The symmetry with parameters also doesn't seem directly beneficial, so I would say that the extra form isn't worthwhile.

To be clear, I'm suggesting that the grammar is simplified from:
`('(' RESULT value_type* ')')*`
to:
`('(' RESULT value_type* ')')?`

cretz:
@rossberg, @binji, et al - Commenting here to keep from opening a new issue. I don't see the result-sans-type as supported in the interpreter readme or the spec. For example, this commit added:

```
  (func $complex
    (param i32 f32) (param $x i64) (param) (param i32)
    (result) (result i32) (result)
    (local f32) (local $y i32) (local i64 i32) (local) (local f64 i32)
    (unreachable) (unreachable)
  )
```

Where is `(result)` (i.e. sans type) acceptable in the grammar? Am I misreading it? What do I do with a no-type result?

PR #498, opened 2017-06-07
[test] Remove expected output

rossberg-chromium:
Testing expected output doesn't really make sense for Wasm, since it has no I/O facilities of its own. Almost all tests use assertions instead. This PR removes the handful of early expected-output files and respective logic from the runner, and converts the relevant tests to rely on assertions.


gahaas:
lgtm

PR #499, opened 2017-06-08
Update wasm-module-builder

hammacher:
This brings the wasm-module-builder.js in sync again with the version
we currently have in v8, from which this clone was initially created.

- the name section uses the new spec'ed format
- {addBody()} automatically appends the end opcode
- support custom sections
- more type checking
- use !== to check for undefined
- emit i64 global inits as varint
- {toBuffer()} now returns the ArrayBuffer instead of a Uint8Array
- minor fixes / formatting

### DIFF
--- a/test/harness/wasm-module-builder.js
+++ b/test/harness/wasm-module-builder.js
@@ -99,6 +99,17 @@ class WasmFunctionBuilder {
   }
 
   addBody(body) {
+    for (let b of body) {
+      if (typeof b != 'number')
+        throw new Error('invalid body (entries have to be numbers): ' + body);
+    }
+    this.body = body.slice();
+    // Automatically add the end for the function block to the body.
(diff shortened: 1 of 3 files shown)


test/harness/wasm-module-builder.js
```diff
@@ -99,6 +99,17 @@ class WasmFunctionBuilder {
...
+      if (typeof b != 'number')
+        throw new Error('invalid body (entries have to be numbers): ' + body);
+    }
+    this.body = body.slice();
+    // Automatically add the end for the function block to the body.
+    this.body.push(kExprEnd);
+    return this;
+  }
+
+  addBodyWithEnd(body) {
```
rossberg-chromium: Is this actually used for anything? If not I'd suggest removing it.

hammacher: Done.

PR #500, opened 2017-06-08
[interpreter] Directory reorg

rossberg-chromium:
Reorganise directory structure of interpreter sources into smaller parts that match structure of spec document. The prior spec/ directory is split into

* `syntax`: the definition of abstract syntax; corresponds to the "Structure" section
* `valid`: validation of code and modules; corresponds to the "Validation" section
* `exec`: execution and module instantiation; corresponds to the "Execution" section
* `binary`: encoding and decoding of binary format; corresponds to the "Binary Format" section
* `text`: parsing and printing of text format; corresponds to the "Text Format" section

Similarly, the former host/ directory is split into

* `script`: abstract syntax and execution of the extended script language
* `main`: main program

The former host/imports becomes just host/.

Also, bump version of the interpreter to 1.0.


gahaas:
lgtm

PR #503, opened 2017-06-09
Remove odd tests

jfbastien:
These two tests don't make much sense to me because the string argument invokes the `length` constructor of `TypedArray`:
https://tc39.github.io/ecma262/#sec-typedarray-length
This then goes ToIndex / ToInteger etc, and the string becomes an integer.

What's the point of this test?

### DIFF
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -14,6 +14,8 @@
  * limitations under the License.
 */
 
+const kC0DEFEFE = new Uint8Array([0xC0, 0xDE, 0xFE, 0xFE]);
+
 (function testJSAPI() {
 
 const WasmPage = 64 * 1024;
@@ -686,7 +688,7 @@ assertCompileError([1], TypeError);
 assertCompileError([{}], TypeError);
 assertCompileError([new Uint8Array()], CompileError);
 assertCompileError([new ArrayBuffer()], CompileError);
-assertCompileError([new Uint8Array([0xC0, 0xDE, 0xFE, 0xFE]), CompileError);
+assertCompileError([kC0DEFEFE], CompileError);
 
 num_tests = 1;
 function assertCompileSuccess(bytes) {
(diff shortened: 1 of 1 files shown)


jfbastien:
@bnjbvr @lukewagner FWIW my current SM shell build does the wrong thing with `new Uint8Array("hi!")`, so you should be failing that test 😁

jfbastien:
To be clear, this coerces the string to integer, it becomes 0, empty array happens.

mtrofin:
Oh... I won't hide my surprise (betraying my JS naivety) that "hi!" becomes the integer 0.

The intent was to test with arrays populated with garbage. @jfbastien , thanks for identifying the bug in those tests!

Could you please replace them with the JS matching the intent?

Thanks!

jfbastien:
I'm not sure what garbage you want. Neutering would be what I can think of, but `postMessage` isn't available in shells. We use a JSC builtin for that in our tests.

jfbastien:
OK so just whatever integers then.

jfbastien:
@mtrofin like this?

test/js-api/jsapi.js
```diff
@@ -686,8 +686,7 @@ assertCompileError([1], TypeError);
 assertCompileError([{}], TypeError);
 assertCompileError([new Uint8Array()], CompileError);
 assertCompileError([new ArrayBuffer()], CompileError);
-assertCompileError([new Uint8Array("hi!")], CompileError);
-assertCompileError([new ArrayBuffer("hi!")], CompileError);
+assertCompileError([new Uint8Array([0xC0, 0xDE, 0xFE, 0xFE]), CompileError);
```
mtrofin: Mind defining a const kInvalidWasmBuffer = [0xC0, 0xDE, 0xFE, 0xFE], and then we can just reuse this? We may have a few other places we're doing this, too.

jfbastien: Done.

PR #505, opened 2017-06-12
Test harness: fix assert_invalid, improve error messages

jfbastien:
assert_invalid uses module() and checks:
            assert_true(e instanceof WebAssembly.CompileError, "expected invalid failure:");

But module() catches and re-throws a plain Error, so instanceof cannot succeed.

While I'm here, improve the error messages a bit.


jfbastien:
Ping.

PR #510, opened 2017-06-19
Add test in traps.wast

Cellule:
Add test for dead-code with i32.div_s & i64.div_s (min_int, -1)

### DIFF
--- a/test/core/traps.wast
+++ b/test/core/traps.wast
@@ -17,8 +17,8 @@
 (assert_trap (invoke "no_dce.i32.div_u" (i32.const 1) (i32.const 0)) "integer divide by zero")
 (assert_trap (invoke "no_dce.i64.div_s" (i64.const 1) (i64.const 0)) "integer divide by zero")
 (assert_trap (invoke "no_dce.i64.div_u" (i64.const 1) (i64.const 0)) "integer divide by zero")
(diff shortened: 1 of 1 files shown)


test/core/traps.wast
```diff
@@ -17,6 +17,8 @@
 (assert_trap (invoke "no_dce.i32.div_u" (i32.const 1) (i32.const 0)) "integer divide by zero")
 (assert_trap (invoke "no_dce.i64.div_s" (i64.const 1) (i64.const 0)) "integer divide by zero")
 (assert_trap (invoke "no_dce.i64.div_u" (i64.const 1) (i64.const 0)) "integer divide by zero")
+(assert_trap (invoke "no_dce.i32.div_s" (i32.const 0x80000000) (i32.const -1)) "interger overflow")
```
rossberg-chromium: Typo: "integer" (which is why it fails on Travis)

test/core/traps.wast
```diff
@@ -17,6 +17,8 @@
...
+(assert_trap (invoke "no_dce.i64.div_s" (i64.const 0x8000000000000000) (i64.const -1)) "interger overflow")
```
rossberg-chromium: Same here.

PR #512, opened 2017-06-27
[interpreter/test] Fix symbolic forward references and change inline type expansion

rossberg-chromium:
This PR fixes/changes identifier resolution for the text format:

1. Correctly handle forward references to type definitions, and add tests.
2. Tweak the expansion of inline function types by inserting type definitions at the _end_ of the module.

The latter should hardly affect any existing programs, because it's only observable if a program uses raw numeric type indices _and_ forward-references later type definitions with them.

See [comment below](https://github.com/WebAssembly/spec/pull/512#issuecomment-311778787) for an example.

DETAILS

All module-level symbolic names are in scope everywhere, but the interpreter did not handle forward references to type definitions correctly, and there were no tests. This PR fixes that, and handles name resolution and AST construction more consistently in 2 passes (apologies that the implementation of this in the parser has become a bit messy over time, I should probably separate it out at some point).

In this context it turned out that the expansion of inline function types was a big issue, because they were inserted in place. That's problematic for at least the following reasons:

- Inline types may or may not create a type definition, so in-place insertion of _implicit_ types has very brittle effects on the effective indices of _explicit_ type definitions following later, which become hard to predict.

- It makes it unnecessarily difficult to correctly define & implement the resolution of forward references to explicit type definitions in general, e.g. when they cross function definitions.

- It would be almost impossible to extend this behaviour cleanly to mutually recursive type definitions, as needed e.g. for GC types (prototyping those were how I stumbled over this issue).

The new behaviour implemented in this PR -- inserting all implicit type definitions, in order, at the end of the module -- is actually consistent with what is stated in the spec draft (which was more by accident than by design). However, the change required adjusting one existing test (in func.wast) that is using raw indices and was relying on the old behaviour, so clearly, other implementations will be affected.

### DIFF
--- a/interpreter/text/parser.mly
+++ b/interpreter/text/parser.mly
@@ -59,8 +59,8 @@ module VarMap = Map.Make(String)
 type space = {mutable map : int32 VarMap.t; mutable count : int32}
 let empty () = {map = VarMap.empty; count = 0l}
 
-type types = {mutable tmap : int32 VarMap.t; mutable tlist : type_ list}
-let empty_types () = {tmap = VarMap.empty; tlist = []}
+type types = {space : space; mutable list : type_ list}
+let empty_types () = {space = empty (); list = []}
 
 type context =
   { types : types; tables : space; memories : space;
@@ -74,14 +74,11 @@ let empty_context () =
 let enter_func (c : context) =
   {c with labels = VarMap.empty; locals = empty ()}
 
-let type_ (c : context) x =
-  try VarMap.find x.it c.types.tmap
-  with Not_found -> error x.at ("unknown type " ^ x.it)
-
 let lookup category space x =
   try VarMap.find x.it space.map
   with Not_found -> error x.at ("unknown " ^ category ^ " " ^ x.it)
 
+let type_ (c : context) x = lookup "type" c.types.space x
 let func (c : context) x = lookup "function" c.funcs x
 let local (c : context) x = lookup "local" c.locals x
 let global (c : context) x = lookup "global" c.globals x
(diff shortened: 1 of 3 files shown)


binji:
Sorry -- I don't really understand the ML changes, so I can't really comment on them.

rossberg-chromium:
@binji, don't worry about the details of the implementation, that is less interesting. What I thought might be relevant to you are the added tests and the change to function signature expansion.

Being pedantic, the latter changes the meaning of the text format as previously implemented. However, the difference is only observable by programs that use raw type indices, so it arguably is very minor. Yet, I wanted to make sure you are aware.

Specifically,
```
(module
  (func (result f64) unreachable)
  (func (param i32))
  (type (func (param i32)))
)
```
now desugars into
```
(module
  (func (type 1) unreachable)
  (func (type 0))
  (type $0 (func (param i32)))
  (type $1 (func (result f64)))
)
```
whereas previously it desugared into
```
(module
  (type $0 (func (result f64)))
  (func (type 0) unreachable)
  (type $1 (func (param i32)))
  (func (type 1))
  (type $2 (func (param i32)))
)
```
(note the different index order and the failure to reuse the later type definition).

binji:
OK, thanks for this simplified example. It may be nice to have something like this in the testsuite as well, for clarity.

Just so I understand the proposed behavior:

```
(module
  (func (result f64) unreachable)
  (func (param i32))
  (type (func (param i32))))
```

1. The explicit type `[i32]->[]` is given type index 0
1. The implicit type `[]->[i64]` of function 0 is given type index 1
1. The implicit type of function 1 matches type index 0, so it is reused.

Is that correct?

If so, this seems fine to me.

AndrewScheidecker:
What's the argument for allowing `type` declarations after `func` declarations? The text format already requires that imports occur before definitions; why not also require that `type` declarations occur before both?

rossberg-chromium:
@binji, yup. Added a simpler test and split the existing one out of the larger module to make it clearer as well.

@AndrewScheidecker, yes, under that restriction both semantics would be indistinguishable. But it would be a bit artificial and user-unfriendly. More importantly, once we add the ability to import types (which e.g. will be necessary to support host types or GC types), such a restriction would not help anyway.

PR #513, opened 2017-06-28
Use official WebAssembly logo

### LINKED ISSUE #477: [spec] Use official WebAssembly logo
https://github.com/WebAssembly/spec/blob/master/document/placeholder.jpg is an older proposed logo, the spec should instead use the [official logo](https://github.com/WebAssembly/web-assembly-logo/tree/master/dist).

chicoxyzzy:
**Attention**: PDF build failed on my laptop so I'm not sure it works as expected (I use SVG logo here).
Part of log:
```bash
latexmk -pdf -dvi- -ps-  'WebAssembly.tex'
make[1]: latexmk: No such file or directory
make[1]: *** [WebAssembly.pdf] Error 1
make: *** [latexpdf] Error 2
```

Screenshot of HTML page:
screen shot 2017-06-28 at 13 08 40

fixes #477


chicoxyzzy:
Can I check builded files on [Travis](https://travis-ci.org/WebAssembly/spec/builds/247875801) somehow?

rossberg-chromium:
No, the spec build hasn't been set up for Travis, though that might be worth looking into.

I tried it on my machine, and unfortunately I get "! LaTeX Error: Unknown graphics extension: .svg.". But I suppose a JPG or PNG should be good enough for this purpose.

One question: would it be possible to make the logo appear somewhat smaller, with more space around it? As is it looks rather overblown and cramped on the page.

chicoxyzzy:
Updated to PNG instead of SVG. Also I install MacTex instead of BasicTex and was able to sure PDF was built without any issue.

@rossberg-chromium AFAIK Sphinx theme used here [doesn't support](http://alabaster.readthedocs.io/en/latest/customization.html#theme-options) such option and logo size is aligned to navigation bar width. I think it's possible to use css rules but I'm not good at CSS.

rossberg-chromium:
Thanks! I will look into tweaking the CSS (though I'm not at all good at that either).

Issue #508, opened 2017-06-15
[spec] Nondescript anchor names


binji:
I was trying to link to specific sections of the spec document and found that some have unexpected names. For example, in the execution runtime document, the "Stack" anchor is https://webassembly.github.io/spec/execution/runtime.html#index-11. In the same document, the "values" anchor is https://webassembly.github.io/spec/execution/runtime.html#id4.

It seems that https://webassembly.github.io/spec/execution/runtime.html#stack actually does work, so perhaps ReST is just choosing a poor example for the hover anchor? The values example is trickier, since it works too but goes to a different values section.

rossberg-chromium:
These and all similar issues should be fixed by #514, although Sphinx's underlying logic is unclear, so there may still be other problems of this sort.

PR #514, opened 2017-06-29
[spec] Fix cross references and anchors

rossberg-chromium:
This fixes all missing or duplicate xref labels.

It also massages the order of label definitions and index markup to address #508. Unfortunately,  the underlying rules that Sphinx uses to assign label names are undocumented and less than obvious, so there may be other occurrences of this.

Another issue remains with the generated LaTeX: when several labels are defined without text inbetween, Sphinx's Latex generator will erroneously insert these labels multiple times, creating lots of warnings from Latex. All my attempts to work around this bug have failed.


rossberg-chromium:
Landing, since this solely contains fixes and doesn't change content.

PR #519, opened 2017-07-06
[spec] Clean-up pass over main body of spec

rossberg-chromium:
- Fix various layout issues
- Introduce yet more cross-references, especially to record definitions
- Move some definitions to more suitable sections
- Fix bugs and typos
- Misc wording tweaks and additions


rossberg-chromium:
Mainly clean-up with only marginal changes in contents, so taking the liberty to land directly.

PR #522, opened 2017-07-11
[spec] Specify name section

rossberg-chromium:
Preview: http://webassembly.github.io/spec/appendix/custom.html

### DIFF
--- a/document/core/appendix/custom.rst
+++ b/document/core/appendix/custom.rst
@@ -69,7 +69,7 @@ Name Maps
 
 A *name map* assigns :ref:`names <syntax-name>` to :ref:`indices <syntax-index>` in a given :ref:`index space <syntax-index>`.
 It consists of a :ref:`vector <binary-vec>` of index/name pairs in order of increasing index value.
-Each index is expected to be unique, but the assigned names need not be.
+Each index must be unique, but the assigned names need not be.
 
 .. math::
    \begin{array}{llclll}
@@ -80,8 +80,8 @@ Each index is expected to be unique, but the assigned names need not be.
    \end{array}
 
 An *indirect name map* assigns :ref:`names <syntax-name>` to a two-dimensional :ref:`index space <syntax-index>`, where secondary indices are *grouped* by primary indices.
-It consists of a vector of primary index/name map pairs in order of incresing index value, where each name map in turn maps secondary indices to names.
-Each primary index is expected to be unique, as is each secondary index per individual name map.
+It consists of a vector of primary index/name map pairs in order of increasing index value, where each name map in turn maps secondary indices to names.
+Each primary index must be unique, and likewise each secondary index per individual name map.
 
 .. math::
    \begin{array}{llclll}
--- a/document/core/appendix/custom.rst
+++ b/document/core/appendix/custom.rst
@@ -1,0 +1,140 @@
+.. index:: custom section, section, binary format
+
+Custom Sections
+---------------
+
+This appendix defines dedicated :ref:`custom sections <binary-customsec>` for WebAssembly's :ref:`binary format <binary>`.
+Such sections do not contribute to, or otherwise affect, the WebAssembly semantics, and like any custom section they may be ignored by an implementation.
+However, they provide useful meta data that implementations can make use of to improve user experience or take compilation hints.
+
+Currently, only one dedicated custom section is defined, the :ref:`name section<binary-namesec>`.
+
+
+.. index:: ! name section, name, Unicode UTF-8
+.. _binary-namesec:
+
+Name Section
+~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


document/core/appendix/custom.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. index:: ! name map, index, index space
+.. _binary-indirectnamemap:
+.. _binary-namemap:
+
+Name Maps
+.........
+
+A *name map* assigns :ref:`names <syntax-name>` to :ref:`indices <syntax-index>` in a given :ref:`index space <syntax-index>`.
+It consists of a :ref:`vector <binary-vec>` of index/name pairs in arbitrary order.
```
binji: The design doc says that they must be sorted by index: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#name-map.

rossberg-chromium: Oh right, missed that. Fixed.

document/core/appendix/custom.rst
```diff
@@ -80,8 +80,8 @@ Each index is expected to be unique, but the assigned names need not be.
    \end{array}
 
 An *indirect name map* assigns :ref:`names <syntax-name>` to a two-dimensional :ref:`index space <syntax-index>`, where secondary indices are *grouped* by primary indices.
-It consists of a vector of primary index/name map pairs, where each name map in turn maps secondary indices to names.
-Each primary index is expected to be unique, as is each secondary index in each individual name map.
+It consists of a vector of primary index/name map pairs in order of incresing index value, where each name map in turn maps secondary indices to names.
```
jfbastien: typo "increasing"

document/core/appendix/custom.rst
```diff
@@ -68,7 +68,7 @@ Name Maps
 .........
 
 A *name map* assigns :ref:`names <syntax-name>` to :ref:`indices <syntax-index>` in a given :ref:`index space <syntax-index>`.
-It consists of a :ref:`vector <binary-vec>` of index/name pairs in arbitrary order.
+It consists of a :ref:`vector <binary-vec>` of index/name pairs in order of increasing index value.
```
jfbastien: Monotonically increasing?

document/core/appendix/custom.rst
rossberg-chromium: Fixed.

document/core/appendix/custom.rst
rossberg-chromium: All increasing sequences are monotonic, so it is usually implied: http://tutorial.math.lamar.edu/Classes/CalcII/MoreSequences.aspx

jfbastien: That's not the definition I'm used to. Mine was more in line with:
http://mathworld.wolfram.com/topics/IncreasingandDecreasing.html
Even if monotonic is implied, I'd rather have the spec be (slightly) redundantly correct than minimally and ambiguously correct.

rossberg-chromium: See Wikipedia, too. But I'm not sure how that is even relevant to the wording here -- "in order of increasing value" is common and unambiguous usage, while nobody says "in order of monotonically increasing value".

RyanLamansky:
I'd like to see a note somewhere encouraging placement of this section as early in the WASM file as possible.  In a streaming compilation scenario, it may not be possible to set names after certain stages of the process.

The .NET-based compiler I'm making, for example, where I'm essentially directly mapping WASM to .NET equivalents as the bytes are received, a name section that comes after the Function section means I can't name the generated methods any more; they've already been committed.  Any kind of deferral will add latency.

lukewagner:
The current text in BinaryEncoding.md#name-section actually says the name section (to be valid) must occur after the Data section (so, at the end).  That's the only place FF recognizes it and it's also what binaryen emits.  Given that the names section is large, placement at the end is beneficial since it allows streaming compilation of functions to start immediately and overlap download of the names section.

RyanLamansky:
@lukewagner Over in .NET land (and this is a general property of the tech, not related to my little hobby) native compilation doesn't actually happen until the method is first called, so a large assembly that's not fully utilized has significantly lower startup time, at the obvious expense of being somewhat lethargic initially.  Combined with unbuffered streaming compilation, the time from last byte to the first function call being possible is a few milliseconds, regardless of WASM size.

Oh, well.  Given the different sizes of our users bases (my zero vs. your billions), your needs are more important 🙂

I'll put it behind an option flag.  When set, it will look for a name section before CIL conversion, at the expense of memory usage and startup latency.

rossberg-chromium:
@RyanLamansky, @lukewagner, the text in this PR also states explicitly that the name section shall appear after the data section.

RyanLamansky:
@rossberg-chromium Yeah, I somehow missed that when I initially scrolled through.  I still would have commented, though, but phrased differently..,

PR #523, opened 2017-07-11
[spec] Explain validation algorithm

rossberg-chromium:
Preview: http://webassembly.github.io/spec/appendix/algorithm.html

### DIFF
--- a/document/core/appendix/algorithm.rst
+++ b/document/core/appendix/algorithm.rst
@@ -1,6 +1,193 @@
+.. index:: validation, algorithm, instruction, module, binary format, opcode
+.. _algo-valid:
+
 Validation Algorithm
 --------------------
 
-.. todo::
+The specification of WebAssembly :ref:`validation <valid>` is purely *declarative*.
+It describes the constraints that must be met by a :ref:`module <valid-module>` or :ref:`instruction <valid-instr>` sequence to be valid.
 
-   Describe algorithm, state correctness properties (soundness, completeness)
+This section sketches the skeleton of a sound and complete *algorithm* for validating code, i.e., sequences of :ref:`instructions <syntax-instr>`, effectively.
+(Other aspects of validation are straightforward to implement.)
+
+In fact, the algorithm is expressed over the flat sequence of opcodes as occurring in the :ref:`binary format <binary>`, and performs only a single pass over it.
+Consequently, it can be integrated directly into a decoder.
+
+The algorithm is expressed in typed pseudo code whose semantics is intended to be self-explanatory.
+
+
+.. index:: value type, stack, label, frame, instruction
+
+Data Structures
+~~~~~~~~~~~~~~~
+
+The algorithm uses two separate stacks: the *operand stack* and the *control stack*.
+The former tracks the :ref:`types <syntax-valtype>` of operand values on the :ref:`stack <stack>`,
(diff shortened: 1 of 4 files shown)


document/core/appendix/algorithm.rst
```diff
@@ -1,6 +1,193 @@
...
+
+       case (br_if n)
+         pop_opd(I32)
+         error_if(ctrls.size() < n)
+         pop_opds(ctrls[n].label_types)
+         push_opds(ctrls[n].label_types)
 
-   Describe algorithm, state correctness properties (soundness, completeness)
+       case (br_table n* m)
+         error_if(ctrls.size() < m)
```
binji: Need to `pop_opd(I32)` here too.

document/core/appendix/algorithm.rst
```diff
@@ -1,6 +1,193 @@
...
+     let frame = ctrls.pop()
+     pop_opds(frame.end_types)
+     error_if(opds.size() =/= frame.height)
+     return frame.end_types
+
+   func unreachable() =
+     opds.resize(ctrl[0].height)
+     ctrl[0].unreachable := true
+
+Pushing a control frame takes the types of the label and result values.
```
binji: Might want to mention that `push` prepends. Otherwise the `br` steps below and `unreachable` above don't make sense.

document/core/appendix/algorithm.rst
```diff
@@ -1,6 +1,193 @@
...
+
+Validation of Opcode Sequences
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following function shows the validation of a number of representative instructions that manipulate the stack.
+Other instructions are checked in a similar manner.
+
+.. note::
+   Various instructions not shown here will additionally require the presence of a validation :ref:`context <context>` for checking uses of :ref:`indices <syntax-index>`.
+   That is an easy addition and therefor omitted from this presentation.
```
binji: sp: therefore

document/core/appendix/algorithm.rst
```diff
@@ -1,6 +1,193 @@
...
+         pop_opd()
+
+       case (select)
+         pop_opd(I32)
+         let t1 = pop_opd()
+         let t2 = pop_opd(t1)
+         push_opd(t2)
+
+       case (unreachable)
+         unreachable()
```
binji: May want to include `return` here, since that also calls `unreachable`.

document/core/appendix/algorithm.rst
rossberg-chromium: Done.

document/core/appendix/algorithm.rst
rossberg-chromium: Done.

document/core/appendix/algorithm.rst
rossberg-chromium: Done.

document/core/appendix/algorithm.rst
rossberg-chromium: Hm, `return` requires the context, which is why I left it out. It wasn't actually the intention to cover all uses of unreachable. You think it is easy enough to extrapolate that from the rules for branches?

binji: Seems fine to omit.

PR #525, opened 2017-07-17
Memory.grow(0) should use strict equality

dtig:
Specifically when memory is initialized with 0 memory, and then grown by zero, use strict equality as the buffers are different only by reference.

### DIFF
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -485,7 +485,7 @@ test(() => {
     var buf = mem.buffer;
     assert_equals(buf.byteLength, WasmPage);
     assert_equals(mem.grow(0), 1);
-    assert_not_equals(buf, mem.buffer)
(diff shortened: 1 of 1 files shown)


aheejin:
Sorry, I clicked something by mistake...

bnjbvr:
Thanks for the PR. `testharness.js`'s `assert_not_equals(a,b)` asserts `!same_value(a, b)`, which ends up doing a [strict equality comparison](https://github.com/WebAssembly/spec/blob/master/test/harness/testharness.js#L867-L877). So I'm a bit unclear of what we'd get more from this change. Can you explain a bit further please, @dtig?

dtig:
@bnjbvr Thanks for the clarification, this looks like an implementation detail in the engine where it uses deep equality for comparison instead of strict equality. Closing this pull request.

PR #527, opened 2017-07-20
[spec/test/interpreter] Allow underscores in numbers

rossberg-chromium:
Allow literals like 1_000, 0xfff_ffff, 3.145_692 etc, as discussed earlier.

### DIFF
--- a/document/core/text/values.rst
+++ b/document/core/text/values.rst
@@ -27,6 +27,7 @@ Integers
 ~~~~~~~~
 
 All :ref:`integers <syntax-int>` can be written in either decimal or hexadecimal notation.
+In both cases, digits can optionally be separated by underscores.
 
 .. math::
    \begin{array}{llclll@{\qquad}l}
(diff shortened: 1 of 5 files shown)


binji:
FYI: Found this as a useful reference for prior art: https://www.python.org/dev/peps/pep-0515/#prior-art.

Seems as though most languages allow underscores only between digits (like you've done here). It also seems that most languages allow multiple consecutive underscores (unlike here). I don't think it's a big deal either way, though.

rossberg-chromium:
I'd be fine with allowing multiple consecutive underscores. I merely went for the most conservative option.

samuelgoto:
FWIW, this is a similar discussion in JS land that may help inform your design (e.g. tradeoffs and prior art).

https://github.com/tc39/proposal-numeric-separator

samuelgoto:
> I'd be fine with allowing multiple consecutive underscores. I merely went for the most conservative option.

FWIW, the most conservative option is to allow non consecutive underscores. You can always change that later (we haven't found enough evidence that it adds anything) but not retract otherwise. This is the direction that JS is going to with that regards.

PR #530, opened 2017-07-26
Update JavaScript WasmModuleBuilder with new functionality.

titzer:
The original version of this file was a fork of the V8 module builder. This is just an update with the additional functionality that we have seen fit to add in our tests, including, e.g., the new names section format.

### DIFF
--- a/test/harness/wasm-module-builder.js
+++ b/test/harness/wasm-module-builder.js
@@ -107,8 +107,8 @@ class WasmFunctionBuilder {
 
   addBody(body) {
     for (let b of body) {
-      if (typeof b != 'number')
-        throw new Error('invalid body (entries have to be numbers): ' + body);
+      if (typeof b !== 'number' || (b & (~0xFF)) !== 0 )
+        throw new Error('invalid body (entries must be 8 bit numbers): ' + body);
     }
     this.body = body.slice();
     // Automatically add the end for the function block to the body.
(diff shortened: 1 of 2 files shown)


test/harness/wasm-module-builder.js
```diff
@@ -100,17 +107,23 @@ class WasmFunctionBuilder {
 
   addBody(body) {
     for (let b of body) {
-      if (typeof b !== 'number' || (b & (~0xFF)) !== 0 )
```
hammacher: It seems that in this case, the version in the spec repo was updated to do even stricter tests. Hence we should merge this change back to v8.

test/harness/wasm-module-builder.js
```diff
@@ -365,10 +378,9 @@ class WasmModuleBuilder {
       if (debug) print("emitting memory @ " + binary.length);
       binary.emit_section(kMemorySectionCode, section => {
         section.emit_u8(1);  // one memory entry
-        const has_max = wasm.memory.max !== undefined;
```
hammacher: Also here, the spec version looks more reasonable than the v8 version.

test/harness/wasm-module-builder.js
titzer: Good catch. I had just blindly copied over the v8 version instead of merging the two.

test/harness/wasm-module-builder.js
titzer: Done.

titzer:
This is ready to merge.

gahaas:
lgtm

PR #535, opened 2017-07-29
Add test for unreachable tee_local value

binji:
See https://github.com/WebAssembly/wabt/issues/586.

### DIFF
--- a/test/core/unreached-invalid.wast
+++ b/test/core/unreached-invalid.wast
@@ -627,7 +627,7 @@
 (assert_invalid
   (module (func $tee-local-unreachable-value
     (local i32)
-    tee_local 0 (unreachable))
-  )
+    (tee_local 0 (unreachable))
+  ))
   "type mismatch"
 )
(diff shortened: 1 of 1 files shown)


test/core/unreached-invalid.wast
```diff
@@ -623,3 +623,11 @@
   ))
   "type mismatch"
 )
+
+(assert_invalid
+  (module (func $tee-local-unreachable-value
+    (local i32)
+    tee_local 0 (unreachable))
```
rossberg-chromium: Nit: parens for consistency :)

rossberg-chromium: Actually, the bug in Wabt seems to be for `(unreachable) (tee_local 0)`, while this test is the other way round. Is that intentional?

binji: Oops

binji:
Travis seems to be wedged, maybe from the GH outage this morning? The other bot passed, so I'm assuming it is OK.

PR #540, opened 2017-08-01
Fix small issues in the spec document

gahaas:
This PR fixes some small issues I noticed when reading through the spec. Please take a special look at the changes to the floating point parts.


document/core/syntax/values.rst
```diff
@@ -108,7 +108,7 @@ No distinction is made between signalling and quiet NaNs.
      {+} \fNmag ~|~ {-} \fNmag \\
    \production{floating-point magnitude} & \fNmag &::=&
      (1 + \uM\cdot 2^{-M}) \cdot 2^e & (\iff -2^{E-1}+2 \leq e \leq 2^{E-1}-1) \\ &&|&
-     (0 + \uM\cdot 2^{-M}) \cdot 2^e & (\iff e = -2^{E-1}+2) \\ &&|&
+     (0 + \uM\cdot 2^{-M}) \cdot 2^e & (\iff e = -2^{E-1}+1) \\ &&|&
```
rossberg-chromium: I believe this was correct before: in the abstract syntax, subnormals are distinguished by the leading 0, while the exponent has the same value as the smallest possible normal one (it is, however, encoded specially, see the binary format). This may be worth adding a note for.

gahaas: Thanks for noticing it. I did not realize that the exponent is interpreted differently for subnormals. I removed this change again.

PR #544, opened 2017-08-09
Fix decoding of f64 binary values

### LINKED ISSUE #543: Spec interpreter invalid binary format float const reading
Consider
````
(module
 (export "calc" (func $calc))
 (func $calc (result f64)
  (f64.convert_u/i32
   (i32.const 4294967249)
  )
 )
)
````
and
````
(module
 (export "calc" (func $calc))
 (func $calc (result f64)
  (f64.const 4294967249)
 )
)
````
The only difference is that in the first we have the const as an int, then convert to a float, and in the second the conversion has been done already. But running these files leads to different results:
````
$ wasm a.wasm -e '(invoke "calc")' && wasm b.wasm -e '(invoke "calc")'
4294967249. : f64
4294965201. : f64
````
This is the same no matter who translates the text to binary (binaryen, wabt, the spec interpreter itself), so it looks like the bug is in the reading of the float const in the spec interpreter.

### PR DESCRIPTION
**Pull request by binji**
This loads the u64 value in two u32 pieces, but was using
`Int64.of_int32` which sign-extends the i32 value. This causes the f64
value `4294967249` to be read as `4294965201`, for example.

Fixes #543.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -72,9 +72,8 @@ let u32 s =
   Int32.(add lo (shift_left hi 16))
 
 let u64 s =
-  let u64_of_u32 x = Int64.logand (Int64.of_int32 x) 0x00000000ffffffffL in
-  let lo = u64_of_u32 (u32 s) in
-  let hi = u64_of_u32 (u32 s) in
+  let lo = I64_convert.extend_u_i32 (u32 s) in
+  let hi = I64_convert.extend_u_i32 (u32 s) in
   Int64.(add lo (shift_left hi 32))
 
 let rec vuN n s =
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: interpreter/binary/decode.ml [hunk 1]
```diff
@@ -72,8 +72,9 @@ let u32 s =
   Int32.(add lo (shift_left hi 16))
 
 let u64 s =
-  let lo = Int64.of_int32 (u32 s) in
-  let hi = Int64.of_int32 (u32 s) in
+  let u64_of_u32 x = Int64.logand (Int64.of_int32 x) 0x00000000ffffffffL in
```
rossberg-chromium: This function already exists, you can simply use `I64_convert.extend_u_i32`.

binji: Done.

PR #549, opened 2017-08-23
Fix invalid module names in test

gahaas:
The module names in the tests in elem.wast contained a "-", which is an invalid character in JavaScript variable names. Thereby the JavaScript test generated from elem.wast could not be executed anymore. This CL removes the "-" from the module names and thereby fixes the problem.

### DIFF
--- a/test/core/elem.wast
+++ b/test/core/elem.wast
@@ -164,7 +164,7 @@
 
 ;; Test element sections across multiple modules change the same table
 
-(module $module-1
+(module $module1
   (type $out-i32 (func (result i32)))
   (table (export "shared-table") 10 anyfunc)
   (elem (i32.const 8) $const-i32-a)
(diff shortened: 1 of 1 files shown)


RyanLamansky:
A couple things:

1. WebAssembly isn't intrinsically constrained by JavaScript limitations.  If it were, it wouldn't be possible to use 64-bit integers, for instance.
2. You can access module names with any Unicode characters using object["module-1"] syntax.

gahaas:
@RyanLamansky The problem here is not a problem with WebAssembly itself but with the test script test/core/run.py. This test script generates .js files from the .wast files in the test/core directory. The .js files can then be run directly in JavaScript runtimes like SpiderMonkey or V8 to test their WebAssembly implementation. With the current module names the test script generates invalid JavaScript code which cannot be executed in JavaScript runtimes. With this PR the generated JavaScript code is valid again. Alternatively the test script could be changed, but I do not think that this would be worth the effort, especially since the tests I changed do not test the names used within WebAssembly modules but the correctness of the handling of the elements section.

PR #553, opened 2017-08-29
[test] Add tests for unreachable br_if

rossberg-chromium:
Adds a few tests and fixes a bug in interpreter error message.

See https://bugs.chromium.org/p/v8/issues/detail?id=6651


titzer:
lgtm

PR #567, opened 2017-09-19
[spec] Add embedder interface

rossberg-chromium:
This PR adds an appendix defining an "embedder interface", i.e., entry points that other specs like the JS API can reference to define the operations that they provide to a host environment. The intention is that this interface is complete in the sense that no other parts of the core spec (except pieces of the abstract syntax, like types and values) need to be directly referenced from an embedder spec.

Preview at: https://webassembly.github.io/spec/appendix/embedding.html

### DIFF
--- a/document/core/appendix/embedding.rst
+++ b/document/core/appendix/embedding.rst
@@ -1,0 +1,526 @@
+.. index:: ! embedding, embedder, implementation, host
+.. _embed:
+
+Embedding
+---------
+
+A WebAssembly implementation will typically be *embedded* into a *host* environment.
+An *embedder* implements the connection between such a host environment and the WebAssembly semantics as defined in the main body of this specification.
+An embedder is expected to interact with the semantics in well-defined ways.
+
+This section defines a suitable interface to the WebAssembly semantics in the form of entry points through which an embedder can access it.
+The interface is intended to be complete, in the sense that an embedder does not need to reference other functional parts of the WebAssembly specification directly.
+
+.. note::
+   On the other hand, an embedder does not need to provide the host environment with access to all functionality defined in this interface.
+   For example, an implementation may not support :ref:`parsing <embed-parse-module>` of the :ref:`text format <text>`.
+
+In the description of the embedder interface, syntactic classes from the :ref:`abstract syntax <syntax>` and the :ref:`runtime's abstract machine <syntax-runtime>` are used as names for variables that range over the possible objects from that class.
+Hence, these syntactic classes can also be interpreted as types.
+
+.. _embed-error:
+
+Failure of an interface operation is indicated by an auxiliary syntactic class:
+
+.. math::
+   \begin{array}{llll}
+   \production{(error)} & \error &::=& \ERROR \\
+   \end{array}
+
+In addition to the error conditions specified explicitly in this section, implementations may also return errors when specific :ref:`implementation limitations <impl>` are reached.
+
+
+.. note::
+   Errors are abstract and unspecific with this definition.
+   Implementations can refine it to carry suitable classifications and diagnostic messages.
+
+
+
+.. index:: allocation, store
+.. _embed-store:
+
+Store
+~~~~~
+
+.. _embed-init-store:
+
+:math:`\F{init\_store}() : \store`
+..................................
+
+* Return the empty :ref:`store <syntax-store>`.
+
+.. math::
+   \begin{array}{lclll}
+   \F{init\_store}() &=& \{ \SFUNCS~\epsilon,~ \SMEMS~\epsilon,~ \STABLES~\epsilon,~ \SGLOBALS~\epsilon \} \\
+   \end{array}
+
+
+
+.. index:: module
+.. _embed-module:
+
+Modules
+~~~~~~~
+
+.. index:: binary format
+.. _embed-decode-module:
+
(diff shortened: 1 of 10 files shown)


document/core/appendix/embedding.rst
```diff
@@ -0,0 +1,526 @@
...
+:math:`\F{validate\_module}(\module) : \error^?`
+................................................
+
+* If :math:`\module` is :ref:`valid <valid-module>`, then return nothing.
+
+* Else, return :math:`\ERROR`.
+
+.. math::
+   \begin{array}{lclll}
+   \F{validate\_module}(m) &=& \epsilon && (\iff {} \vdashmodule m : \externtype^\ast) \\
```
binji: According to https://webassembly.github.io/spec/appendix/index-rules.html, this should be `m : externtype* -> externtype*`, right?

document/core/appendix/embedding.rst
```diff
@@ -0,0 +1,526 @@
...
+:math:`\F{decode\_module}(\byte^\ast) : \module ~|~ \error`
+...........................................................
+
+* If there exists a derivation for the :ref:`byte <syntax-byte>` sequence :math:`\byte^\ast` as a :math:`\Bmodule` according to the :ref:`binary grammar for modules <binary-module>`, yielding a :ref:`module <syntax-module>` :math:`m`, then return :math:`m`.
+
+* Else, return :math:`\ERROR`.
+
+.. math::
+   \begin{array}{lclll}
+   \F{decode\_module}(b^\ast) &=& m && (\iff \Bmodule \longrightarrow m{:}b^\ast) \\
```
binji: What does the longrightarrow mean? I can guess, but I don't see where it's defined.

document/core/appendix/embedding.rst
```diff
@@ -0,0 +1,526 @@
...
+
+Exports
+~~~~~~~
+
+.. _embed-get-export:
+
+:math:`\F{get\_export}(\moduleinst, \name) : \externval ~|~ \error`
+...................................................................
+
+* For each :math:`\exportinst_i` in :math:`\moduleinst.\MIEXPORTS`, do:
```
binji: Is it necessary to assert that the name is unique?

document/core/appendix/embedding.rst
```diff
@@ -0,0 +1,526 @@
...
+:math:`\F{alloc\_func}(\store, \functype, \hostfunc) : (\store, \funcaddr)`
+...........................................................................
+
+* Let :math:`\funcaddr` be the result of :ref:`allocating a host function <alloc-func>` in :math:`\store` with :ref:`function type <syntax-functype>` :math:`\functype` and host function code :math:`\hostfunc`.
+
+* Return the new store paired with :math:`\funcaddr`.
+
+.. math::
+   \begin{array}{lclll}
+   \F{alloc\_func}(S, \X{ft}, \X{code}) &=& (S', \X{a}) && (\iff \alloctable(S, \X{ft}, \X{code}) = S', \X{a}) \\
```
binji: \allochostfunc

document/core/appendix/embedding.rst
```diff
@@ -0,0 +1,526 @@
...
+.. math::
+   \begin{array}{lclll}
+   \F{write\_mem}(S, a, i, b) &=& S' && (\iff S' = S \with \SMEMS[a].\MIDATA[i] = b) \\
+   \F{write\_mem}(S, a, i, b) &=& \ERROR && (\otherwise) \\
+   \end{array}
+
+
+.. _embed-grow-mem:
+
+:math:`\F{grow\_mem}(\store, \memaddr, n) : \store ~|~ \error`
```
binji: Is it possible/worthwhile to share this somehow with the [grow_memory instruction](https://webassembly.github.io/spec/exec/instructions.html#exec-grow-memory)?

Also, according to the note that instruction can fail for unspecified reasons. Should this function have similar non-determinism?

document/core/appendix/embedding.rst
binji: Yeah I think this is fine, I'm just not accustomed to reading stuff like this so I usually like to make sure I actually know what the symbols mean (which is why [Index of Semantic Rules](https://webassembly.github.io/spec/appendix/index-rules.html) is so useful). You're right, it's probably OK without further explanation since the text is there.

document/core/exec/modules.rst
```diff
@@ -371,6 +371,46 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
...
+
+.. index:: memory, memory instance, memory address, grow, limits
+.. _grow-mem:
+
+Growing :ref:`memories <syntax-meminst>`
+........................................
+
+1. Let :math:`\meminst` be the :ref:`memory instance <syntax-meminst>` to grow and :math:`n` the number of :ref:`pages <page-size>` by which to grow it.
+
+2. If :math:`\meminst.\MIMAX` is not empty and smaller than :math:`n` added to the length of :math:`\meminst.\MIDATA` divided by the :ref:`page size <page-size>` :math:`64\,\F{Ki}`, then fail.
```
binji: It's a little odd that the text here describes dividing by page size, but the formula below multiplies instead. It's symmetric since `meminst.data` always has size that is a multiple of page size, but I wonder if it would be clearer to use the same operation.

rossberg-chromium: True, changed.

PR #569, opened 2017-09-22
[interpreter] Editorial changes to match spec more closely

rossberg-chromium:
This PR should not imply any functional change. It merely does a bunch of renamings and smaller refactorings to match the spec document more closely. The only observable difference are somewhat more generic error messages for import type mismatches.

(This is in preparation for releasing a new 1.0 Opam package.)


binji:
Mostly just scanned, but lgtm

PR #576, opened 2017-10-10
[test] Add tests for control instructions in unreachable code

hammacher:
The current tests allowed an implementation to either assume an else
block to be reachable, or inherit the reachability from it's parent
(which would be wrong).
This adds more tests for different control instructions in unreachable
code.

### DIFF
--- a/test/core/unreached-invalid.wast
+++ b/test/core/unreached-invalid.wast
@@ -196,6 +196,18 @@
   "type mismatch"
 )
 (assert_invalid
+  (module (func $type-unary-num-vs-void-in-loop-after-unreachable
+    (unreachable) (loop (drop (i32.eqz (nop))))
+  ))
+  "type mismatch"
+)
+(assert_invalid
+  (module (func $type-unary-num-vs-void-in-i32-loop-after-unreachable
+    (unreachable) (loop (result i32) (i32.eqz (nop)))
+  ))
+  "type mismatch"
+)
+(assert_invalid
   (module (func $type-unary-num-vs-num-after-unreachable
     (unreachable) (drop (i32.eqz (f32.const 1)))
   ))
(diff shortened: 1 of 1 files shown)


test/core/unreached-invalid.wast
```diff
@@ -249,6 +249,42 @@
   ))
   "type mismatch"
 )
+(assert_invalid
+  (module (func $type-unary-num-vs-void-in-block-after-unreachable
+    (unreachable) (block (drop (i32.eq (nop))))
```
rossberg-chromium: Isn't this the same as line 204?
Also, did you mean `eqz`?.

test/core/unreached-invalid.wast
```diff
@@ -249,6 +249,42 @@
...
+  ))
+  "type mismatch"
+)
+(assert_invalid
+  (module (func $type-unary-num-vs-void-in-loop-after-unreachable
+    (unreachable) (loop (drop (i32.eq (nop))))
```
rossberg-chromium: `eqz`?

test/core/unreached-invalid.wast
```diff
@@ -249,6 +249,42 @@
...
+  ))
+  "type mismatch"
+)
+(assert_invalid
+  (module (func $type-unary-num-vs-void-in-i32-loop-after-unreachable
+    (unreachable) (loop (result i32) (i32.eq (nop)))
```
rossberg-chromium: `eqz`?

test/core/unreached-invalid.wast
```diff
@@ -249,6 +249,42 @@
...
+  ))
+  "type mismatch"
+)
+(assert_invalid
+  (module (func $type-unary-num-vs-void-in-if-after-unreachable
+    (unreachable) (if (i32.const 0) (then (i32.eqz (nop))))
```
rossberg-chromium: You may want to add a `drop` around the `i32.eqz`, so that it fires for the right reason.

test/core/unreached-invalid.wast
hammacher: Dropped this one.

test/core/unreached-invalid.wast
hammacher: Done, and moved the loop tests up after line 197.

test/core/unreached-invalid.wast
hammacher: Done

test/core/unreached-invalid.wast
hammacher: Done

PR #588, opened 2017-10-18
Needed for the install target (with @protz)

beurdouche:


beurdouche:
@rossberg It would be nice if you can yes :)

rossberg:
Done.

Issue #583, opened 2017-10-16
Can we state wasm's MIME type next to the recommended file extension?


lukewagner:
Specifically, in §5.1.

jfbastien:
Wouldn't that fall in the Web embedding spec that @littledan is working on?

lukewagner:
It would make sense to mention it there too, since it's part of the semantics of `compileStreaming`/`instantiateSTreaming`, but I was thinking it makes sense here too, as a general "if you are serving one of these, here's the MIME type to use" note (which is technically broader than the Web).

littledan:
Putting it in the web spec seems appropriate to me. As for §5.1, I'm surprised that remark exists. The JS spec doesn't say anything about file extensions. "What's a file  anyway?" is not really the most off-topic question when file extensions generally aren't significant on the web (aside from web servers guessing mimetypes). I think the right "normalized" place to put the file extension is in the IANA media type registry, and you find that by the Web spec saying that the mimetype is application/wasm. We can have other places where these things are indicated, as notes.

rossberg:
I don't have a strong opinion, but I agree that both should be mentioned in the same place, while also agreeing that the main body of the spec may be a somewhat strange place for that. It probably would fit better into an Appendix, but can't find an obvious place. Any suggestion?

littledan:
If you want to put it in the main spec, an appendix with all the mimetypes and file extensions sounds good to me. It'd be nice if whichever place says it points to the IANA media type registry as the source of truth. I don't quite understand why this should be in the main spec while the JS interface is left out, though, when not all Wasm use has file extensions or mimetypes (e.g., the current JS binding doesn't even support either).

lukewagner:
That's true, but it is a somewhat general and non-web-specific convention that "when you need to pick an extension or MIME type, use these".

lukewagner:
So ok if I put up a PR adding 'application/wasm' next to the mention of '.wasm'?

rossberg:
Fine by me.

lukewagner:
Ok, merged.  @rossberg could you perhaps merge into the gh-pages branch?

PR #589, opened 2017-10-18
Mention suggested MIME type in Binary Format | Conventions

lukewagner:
As discussed in #583.  Reading https://www.iana.org/assignments/media-types/media-types.xhtml it seems like the more modern name is "Media Type".

### DIFF
--- a/document/core/binary/conventions.rst
+++ b/document/core/binary/conventions.rst
@@ -20,7 +20,8 @@ Except for a few exceptions, the binary grammar closely mirrors the grammar of t
    Implementations of decoders must support all possible alternatives;
    implementations of encoders can pick any allowed encoding.
 
-The recommended extension for files containing WebAssembly modules in binary format is ":math:`\T{.wasm}`".
+The recommended extension for files containing WebAssembly modules in binary format is ":math:`\T{.wasm}`"
+and the recommended Media Type is ":math:`\T{application/wasm}`".
 
 .. [#compression]
    Additional encoding layers -- for example, introducing compression -- may be defined on top of the basic representation defined here.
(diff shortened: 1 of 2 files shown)


document/core/binary/conventions.rst
```diff
@@ -20,7 +20,8 @@ Except for a few exceptions, the binary grammar closely mirrors the grammar of t
    Implementations of decoders must support all possible alternatives;
    implementations of encoders can pick any allowed encoding.
 
-The recommended extension for files containing WebAssembly modules in binary format is ":math:`\T{.wasm}`".
+The recommended extension for files containing WebAssembly modules in binary format is ":math:`\T{.wasm}`"
+and the recommended Media Type is ":math:`\T{application/wasm}`".
```
rossberg: Nit: "media type" lower case?

lukewagner: https://www.iana.org/assignments/media-types/media-types.xhtml seems to capitalize it everywhere as a proper noun.  I was actually thinking of linking Media Type to https://www.iana.org/assignments/media-types/media-types.xhtml, if you could tell me the right incantation...

rossberg: Interesting, Wikipedia doesn't do that. But that's okay. For linking you can follow the examples in util/macros.def, and then just write `|MimeType|`.

lukewagner:
lgty?

lukewagner:
Oh, I see it was already approved.  I did a test build locally and it worked.

rossberg:
@lukewagner, yes, sorry for the confusion. I updated gh-pages, but in fact, if you can already build locally then `make publish` is all that is required. I updated the README to document that.

PR #599, opened 2017-11-07
 [spec] Use typeuse syntax for call_indirect

rossberg:
This is a late but small change/extension to the text format that I floated with some people.

Instead of a simple type index, the call signature for `call_indirect` is specified using the existing `typeuse` syntax used e.g. in function declarations:
```
call_indirect (type $t)
call_indirect (param i32 i32)
call_indirect (param i32 i64) (result i32)
```
That makes it consistent with how types are/will be referenced elsewhere (besides function declarations, e.g. in block signatures in the [multi-value proposal](https://github.com/WebAssembly/multi-value/)). Furthermore, as the example shows, it implies that you can write the call signature inline, like for functions and blocks.

The expansion rules are like for function declarations. However, inline types for call_indirect must not include parameter names.

### DIFF
--- a/document/core/text/instructions.rst
+++ b/document/core/text/instructions.rst
@@ -87,7 +87,7 @@ The same label identifier may optionally be repeated after the corresponding :ma
 All other control instruction are represented verbatim.
 
 .. math::
-   \begin{array}{llclll}
+   \begin{array}{llcllll}
    \production{plain instruction} & \Tplaininstr_I &::=&
      \text{unreachable} &\Rightarrow& \UNREACHABLE \\ &&|&
      \text{nop} &\Rightarrow& \NOP \\ &&|&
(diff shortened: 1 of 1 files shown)


rossberg:
Merging based on above approval and off-line feedback.

cretz:
I have a request for clarification here as I'm just getting around to implementing this. On https://github.com/WebAssembly/spec/blob/98b90e2ab22053559ded143768d6559d2ba8d2fd/test/core/call_indirect.wast#L59 I am not seeing an existing function type this is referencing. According to [the spec](https://github.com/WebAssembly/spec/blob/98b90e2ab22053559ded143768d6559d2ba8d2fd/document/core/text/modules.rst#abbreviations), "If no such index exists, then a new type definition of the form [...] is inserted at the end of the module."

So, just to confirm, the module-level function type def list can be altered while parsing instructions? I can ask this in a new issue or in another place if this issue isn't where it should happen.

binji:
> So, just to confirm, the module-level function type def list can be altered while parsing instructions?

Yes, the behavior is the same as defining a function without an explicit function type.

cretz:
Thanks @binji, and just one more confirmation since order is important, if the type for the func that contains this `call_indirect` insn of an unforeseen type has also never been seen before, the outer func's type is added to the def list *before* the `call_indirect` ones within, right?

binji:
Not sure if I quite understand, but here's an example:
```
(module
  (type (func (param f64)))
  (import "spectest" "table" (table 10 anyfunc))
  (func (param i32)
    (call_indirect (param f32) (f32.const 1) (get_local 0))))
```

The type section should have:
```
type 0: [f64] -> []
type 1: [i32] -> []
type 2: [f32] -> []
```

rossberg:
Yes, note the sentence at the end of the section, which states that "abbreviations are expanded in the order they appear".

cretz:
Perfect, thanks! It's just taking me a bit of extra work because not only did I never have the type def list passed in as contextual state to my insn parser, but I hadn't needed any module-level mutable state in there until now (granted I could just return a bunch of immutable things from every recursive insn-parse call and shore up, but same annoyance applies). Thanks again.

Issue #600, opened 2017-11-07
Textual spec should indicate that memory alignments are stored as log2


eholk:
Apologies if this is written somewhere else and I missed it, but https://webassembly.github.io/spec/binary/instructions.html#binary-memarg does not seem to mention that alignments are stored as the log2 of the alignment (e.g. 3 instead of 8). The spec interpreter does the left-shift though, so it looks like log2 is what is actually used. The textual spec should mention this too.

rossberg:
Good point, see #601. In the spec the AST itself uses log2, so adding the note there instead.

PR #601, opened 2017-11-08
[spec] Add note on log2 representation for alignment

rossberg:
To address #600.


eholk:
lgtm (I don't think I have rights to give an official review on this)

PR #605, opened 2017-11-22
Use the old subprocess API in test/build.py

gahaas:
The old subprocess API also works for python version below 3.5, and the new API is actually not needed in this script.


gahaas:
@bnjbvr, could you please take a look?

PR #609, opened 2017-11-29
[spec] Fix block values

### LINKED ISSUE #608: specification of labels for loops and exiting from blocks
I have some problems with the spec. Namely, spec states for the [`loop`](https://github.com/WebAssembly/spec/blob/master/document/core/exec/instructions.rst#looptinstrastend): 

- Let L be the label whose arity is 0 and whose continuation is the start of the loop.

And then in the Blocks section, in [Exiting inst* with Label L](https://github.com/WebAssembly/spec/blob/master/document/core/exec/instructions.rst#exiting-instrast-with-label-l) (I assume this explains how `end` works?) it states:

- Let n be the arity of L
- ...
- Pop the results valⁿ from the stack

But is it correct? I'm interpreting this as "`loop..end` block can't return any values, because when executing `loop` we pushing label with arity=0 and when we execute `end` we pop 0 values". Which seems incorrect for me (e.g.  [counter-example](https://github.com/WebAssembly/spec/blob/c4774b47d326e4114f96232f1389a555639d7348/test/core/labels.wast#L69)).

Could someone clarify this to me?

rossberg:
Fix #608.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -964,24 +964,22 @@ Exiting :math:`\instr^\ast` with label :math:`L`
 
 When the end of a block is reached without a jump or trap aborting it, then the following steps are performed.
 
-1. Let :math:`n` be the arity of :math:`L`.
+1. Let :math:`m` be the number of values on the top of the stack.
 
-2. Assert: due to :ref:`validation <valid-instr-seq>`, there are :math:`n` values on the top of the stack.
+2. Pop the values :math:`\val^m` from the stack.
 
-3. Pop the results :math:`\val^n` from the stack.
+3. Assert: due to :ref:`validation <valid-instr-seq>`, the label :math:`L` is now on the top of the stack.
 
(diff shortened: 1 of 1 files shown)


document/core/exec/instructions.rst
```diff
@@ -964,24 +964,22 @@ Exiting :math:`\instr^\ast` with label :math:`L`
 
 When the end of a block is reached without a jump or trap aborting it, then the following steps are performed.
 
-1. Let :math:`n` be the arity of :math:`L`.
+1. Let :math:`m` be the number of values on the top of the stack.
```
binji: I know what you mean by "values on top of the stack", but is that mentioned anywhere in the spec?

rossberg: The runtime section defines the stack. Although never defined explicitly, the spec mentions "the top of the stack" in various places. Do you think it might be too unclear?

binji: I think it's probably fine when it's just a few values. This case is perhaps a little less clear since you're only counting up from the label. That said, the formula below makes it explicit.

PR #614, opened 2017-12-02
[spec] read_table returns an optional function address

HugoGuiroux:
The formal definition is correct, but the return type of the `read_table` embedding function in the prose misses the optional sign.


HugoGuiroux:
Done.

PR #623, opened 2017-12-13
Some editorial improvements to the Wasm/JS spec

littledan:
See commit messages for details

cc @rossberg @lukewagner

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -150,6 +150,7 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
         text: 𝗆𝖾𝗆
         text: 𝗀𝗅𝗈𝖻𝖺𝗅
     text: global type; url: syntax/types.html#syntax-globaltype
+    text: address; url: exec/runtime.html#addresses
 </pre>
 
 <pre class='link-defaults'>
@@ -208,17 +209,17 @@ fetch('demo.wasm').then(response =>
 
 <h3 id="store">Interaction of the WebAssembly Store with JavaScript</h3>
 
-Note: WebAssembly semantics are defined in terms of a [=store=], representing the state of the world, or "memory". WebAssembly operations take a store and return a new store.
+Note: WebAssembly semantics are defined in terms of an abstract [=store=], representing the state of the WebAssembly abstract machine. WebAssembly operations take a store and return an updated store.
 
 Each [=agent=] has an <dfn>associated store</dfn>. When a new agent is created, its associated store is set to the result of [=init_store=]().
 
-Note: In this specification, no WebAssembly-related objects, memory or addresses can be shared among agents in an [=agent cluster=]. In a future version of WebAssembly, the <a href="https://github.com/WebAssembly/threads">threads proposal</a> will add such sharing. For the JS binding, this will likely mean that there will be a store per agent cluster, rather than per agent. This specification uses a store per agent in order to limit semantics to how a store operates in a single-threaded manner, as the WebAssembly specification without threads only deals with that case.
+Note: In this specification, no WebAssembly-related objects, memory or addresses can be shared among agents in an [=agent cluster=]. In a future version of WebAssembly, this may change.
 
 Elements of the WebAssembly store may be <dfn>identified with</dfn> JavaScript values. In particular, each WebAssembly [=memory instance=] with a corresponding {{Memory}} object is identified with a JavaScript [=Data Block=]; modifications to this Data Block are identified to updating the agent's store to a store which reflects those changes, and vice versa.
 
 <h3 id="object-caches">WebAssembly JS Object Caches</h3>
 
-Note: There are several WebAssembly objects that may have a corresponding JavaScript object. The correspondence is stored in a per-agent mapping from WebAssembly addresses to JavaScript objects. This mapping is used to ensure that, for a given [=agent=], there exists at most one JavaScript object for a particular WebAssembly address.
+Note: There are several WebAssembly objects that may have a corresponding JavaScript object. The correspondence is stored in a per-agent mapping from WebAssembly [=address=]es to JavaScript objects. This mapping is used to ensure that, for a given [=agent=], there exists at most one JavaScript object for a particular WebAssembly address.
 
 Each [=agent=] is associated with the following three [=ordered map=]s:
     * The <dfn>Memory object cache</dfn>, mapping [=memory address=]es to {{Memory}} objects.
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -365,9 +365,7 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
     1. If |ret| is [=error=], throw an exception. This exception should be a WebAssembly {{RuntimeError}} exception, unless otherwise indicated by <a href="#errors">the WebAssembly error mapping</a>.
     1. Assert: |ret| is empty.
     1. Let |exportsObject| be ! [=ObjectCreate=](null).
-    1. For each pair (|nameBytes|, |externtype|) in [=module_exports=](|module|),
-        1. Let |name| be |nameBytes| [=UTF-8 decode without BOM or fail|decoded as UTF-8=].
-        1. Assert: |name| is not failure (|module| is [=valid=]).
+    1. For each pair (|name|, |externtype|) in [=module_exports=](|module|),
         1. Let |externval| be [=get_export=](|instance|, |name|).
         1. If |externtype| is of the form [=𝖿𝗎𝗇𝖼=] |functype|,
             1. If |funcs| contains an entry |func| where 𝖿𝗎𝗇𝖼 |func|.\[[FunctionAddress]] is |externval|, let |value| be |func|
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -325,74 +317,46 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
...
-                Note: The table and function objects created by the above steps are only observable for tables that are either imported or exported.
-
-    1. Let |startfuncaddr| be the start function of |instance|.
-    1. Let (|store|, |ret|) be [=invoke_func=](|store|, |startfuncaddr|, empty).
-    1. If |ret| is [=error=], throw an exception. This exception should be a WebAssembly {{RuntimeError}} exception, unless otherwise indicated by <a href="#errors">the WebAssembly error mapping</a>.
-    1. Assert: |ret| is empty.
+    1. Let (|store|, |instance|) be [=instantiate_module=](|store|, |module|, |imports|).
+    1. If |instance| is [=error=], throw an appropriate exception type:
+        * A {{LinkError}} exception for most cases which occur during linking.
+        * If the error came when running the start function, throw a {{RuntimeError}} for most errors which occur from WebAssembly, or the error object propagated from inner ECMAScript code.
```
rossberg: You can remove the "most" here, I think.

document/js-api/index.bs
```diff
@@ -325,74 +317,46 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
...
+    1. For each pair (|name|, |externtype|) in [=module_exports=](|module|),
         1. Let |externval| be [=get_export=](|instance|, |name|).
+        1. Assert: |externval| is not [=error=].
         1. If |externtype| is of the form [=𝖿𝗎𝗇𝖼=] |functype|,
-            1. If |funcs| contains an entry |func| where 𝖿𝗎𝗇𝖼 |func|.\[[FunctionAddress]] is |externval|, let |value| be |func|
-            1. Otherwise,
-                1. Let |func| be the result of [=a new Exported Function=] from |externval|.
-                1. [=Append=] |func| to |funcs|.
-                1. Let |value| be |func|.
+            1. Then, |externval| is of the form [=external value|𝖿𝗎𝗇𝖼=] |funcaddr|.
```
rossberg: Turn this into an assertion?

document/js-api/index.bs
```diff
@@ -325,74 +317,46 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
...
+            1. Let |func| be the result of [=a new Exported Function=] from |funcaddr|.
```
rossberg: s/of a new/of creating a new/ ?

document/js-api/index.bs
```diff
@@ -325,74 +317,46 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
...
-                1. Let |memory| be [=create a memory object|a new Memory object=] from |externval|.
-                1. [=Append=] |memory| to |memories|.
-                1. Let |value| be |memory|.
+            1. Then, |externval| is of the form [=external value|𝗆𝖾𝗆=] |memaddr|.
+            1. Let |memory| be [=create a memory object|a new Memory object=] from |memaddr|.
+            1. Let |value| be |memory|.
         1. Otherwise, |externtype| is of the form [=𝗍𝖺𝖻𝗅𝖾=] |tabletype|,
-            1. Let |value| be the unique |table| in |tables| such that 𝗍𝖺𝖻𝗅𝖾 |table|.\[[Table]] is |externval|.
+            1. Then, |externval| is of the form [=external value|𝗍𝖺𝖻𝗅𝖾=] |tableaddr|.
+            1. Let |table| be [=create a Table object|a new Table object=] from |tableaddr|.
```
rossberg: s/from/created from/ ?

document/js-api/index.bs
```diff
@@ -556,22 +515,31 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
 
     * \[[Memory]] : a [=memory address=]
     * \[[BufferObject]] : an {{ArrayBuffer}} whose [=Data Block=] is [=identified with=] the above memory address
+
+Each [=agent=] has an associated [=ordered map=] known as the <dfn>Memory object cache</dfn> mapping [=memory address=]es to {{Memory}} objects.
```
rossberg: Maybe collect the introduction of these caches in a single place in some earlier section?

document/js-api/index.bs
littledan: I was initially thinking an assertion, but is it OK for an assertion to introduce a binding? I'm not sure what the proper way to phrase this is for typical web spec conventions (cc @tobie).

rossberg: Hm, fair question. I think I have been using assertions for that in the core spec, but it's a fair point. Leaving it up to you.

tobie: So an assertion doesn't do anything itself, it just asserts that at this point, that statement is true and would be a spec bug if it wasn't. See: https://infra.spec.whatwg.org/#assertions. In this particular context, I'm not too sure what "is of the form" actually means. (Maybe link it to a dfn?)

littledan: The Wasm spec uses a sort of "algebraic data type" notation that I'm trying to "destructure" here. This differs from JS Records. I'm not sure what you would recommend. I could also imagine using "Let func funcaddr be externval" here.

tobie: Oh, are you introducing |funcaddr| here for the first time? Then you should def to a `Let |funcaddr| be…` step.

littledan: But, what would it be?

tobie: So I'm not sure I understand precisely what's going on here (and I'm lacking context), which is precisely why I suggested using something like pr-preview to help with this.

I'm not sure I really understand how external values work well enough to comment.

rossberg: The cleanest solution probably is to split this into 2 lines: first the assert about the structure, then a destructuring let. I just checked through the core spec: destructuring let is used in a few places, but no cases exist where it might fail.

document/js-api/index.bs
littledan: Well, if it's a stack overflow or OOM from WebAssembly, it will be different. Probably there is a better way to word this, though.

document/js-api/index.bs
littledan: Any idea for where? Currently, all stuff with agents happens at the end. I imagine it'd be less clear if I moved these things to the end, though.

rossberg: There are plenty of forward refs to the "associated store" already, and these are no different. They are part of the same mapping mechanism, so should best be defined in one place. Maybe the section about store interaction could be moved to the beginning?

lukewagner: A "store" section makes sense.

littledan: OK, moved it to a store section, PTAL.

document/js-api/index.bs
```diff
@@ -204,6 +204,27 @@ fetch('demo.wasm').then(response =>
 );
 ```
 
+<h2 id="webassembly-storage">Internal storage</h2>
+
+<h3 id="store">Interaction of the WebAssembly Store with JavaScript</h3>
+
+Note: WebAssembly semantics are defined in terms of a [=store=], representing the state of the world, or "memory". WebAssembly operations take a store and return a new store.
```
rossberg: Wording suggestion:

* Say "an abstract store" to clarify its role,
* Replace "state of the world" with "state of the WebAssembly abstract machine",
* Remove the "or 'memory'" to avoid confusion with Memory instances,
* Say "and return an updated store".

document/js-api/index.bs
```diff
@@ -204,6 +204,27 @@ fetch('demo.wasm').then(response =>
...
+
+Each [=agent=] has an <dfn>associated store</dfn>. When a new agent is created, its associated store is set to the result of [=init_store=]().
+
+Note: In this specification, no WebAssembly-related objects, memory or addresses can be shared among agents in an [=agent cluster=]. In a future version of WebAssembly, the <a href="https://github.com/WebAssembly/threads">threads proposal</a> will add such sharing. For the JS binding, this will likely mean that there will be a store per agent cluster, rather than per agent. This specification uses a store per agent in order to limit semantics to how a store operates in a single-threaded manner, as the WebAssembly specification without threads only deals with that case.
```
rossberg: I'd avoid any references to specific proposals. Just saying "This may change in future version of WebAssembly" is enough.

document/js-api/index.bs
```diff
@@ -204,6 +204,27 @@ fetch('demo.wasm').then(response =>
...
+
+Elements of the WebAssembly store may be <dfn>identified with</dfn> JavaScript values. In particular, each WebAssembly [=memory instance=] with a corresponding {{Memory}} object is identified with a JavaScript [=Data Block=]; modifications to this Data Block are identified to updating the agent's store to a store which reflects those changes, and vice versa.
+
+<h3 id="object-caches">WebAssembly JS Object Caches</h3>
+
+Note: There are several WebAssembly objects that may have a corresponding JavaScript object. The correspondence is stored in a per-agent mapping from WebAssembly addresses to JavaScript objects. This mapping is used to ensure that, for a given [=agent=], there exists at most one JavaScript object for a particular WebAssembly address.
```
rossberg: Link "addresses" to https://webassembly.github.io/spec/core/exec/runtime.html#addresses

littledan:
I made all of @rossberg 's suggested changes; PTAL.

littledan:
Merging this patch as it's gotten an LGTM.

PR #626, opened 2017-12-16
Adding single page bikeshed wrapper build of core spec.

flagxor:
This produces a single page bikeshed wrapper around the core spec.
A bunch of tex seems to croak with MathJax if it's not broken apart, so did that.

Tab, I had to work around |foo| getting swallowed by bikeshed into <var>foo</var>, does that rise to the level of a bug?

Eric, the resulting doc nominally looks ok, but perhaps we could just incorporate the appropriate text from within Spinx ?


document/core/conf.py
```diff
@@ -129,24 +129,28 @@
...
-  'sidebar_width': '260px',
-  'sidebar_collapse': True,
-  'show_powered_by': False,
-  'extra_nav_links': {
-    'Index': 'BASEDIR/genindex.html',
-    'Download as PDF': 'BASEDIR/_download/' + name + '.pdf'
-  },
+  'nosidebar': True,
+
+#  'logo': logo,
```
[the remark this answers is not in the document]
flagxor: Ah, yeah, not thinking.
Splitting out into an alternate config.

flagxor:
Ok, this seems to be able to build everything.
See result here:
http://flagxor.github.io/spec/

PTAL

littledan:
@flagxor Great job on this document. It's fast to load and comfortable to read!

document/core/katex_fix.patch
```diff
@@ -0,0 +1,29 @@
+diff --git a/dist/katex.css b/dist/katex.css
```
rossberg: Can you move this file to core/util?

document/core/mathjax2katex.py
```diff
@@ -0,0 +1,197 @@
+#! /usr/bin/env python
```
rossberg: Also move to util?

flagxor:
PTAL
Was able to drop the rst changes as katex doesn't choke (only mathjax).
Did each of the comments.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "document/core/katex"]
```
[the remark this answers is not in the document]
rossberg: Don't you need to adjust the above path as well?

document/core/bikeshed/conf.py
```diff
@@ -0,0 +1,470 @@
+# -*- coding: utf-8 -*-
```
rossberg: Nit: move this to /util as well.

flagxor: Moved just the bikeshed one for now.
Probably should consider the other one too at some point. Although it's kind of a gray area whether the config is part of the document or generic.

PR #632, opened 2018-01-01
[spec] Clarify exec/instructions: `ea` in load/store cannot overflow

swasey:


swasey:
Note the Travis CI failure seems to be a temporary problem having nothing to do with the patch.

PR #637, opened 2018-01-04
[spec] Fix evaluation contexts and structural rules

swasey:
Replace frame instructions in evaluation contexts with a
context-sensitive structural rule.

Clarify the side-condition in the trap propagation rule.

### DIFF
--- a/document/core/exec/runtime.rst
+++ b/document/core/exec/runtime.rst
@@ -576,12 +576,12 @@ Finally, the following definition of *evaluation context* and associated structu
    S; F_0; \FRAME_n\{F\}~\instr^\ast~\END &\stepto& S'; F_0; \FRAME_n\{F'\}~\instr'^\ast~\END
    \end{array}
    \\ \qquad
-     (\iff S; F; \instr^\ast \stepto S'; F'; {\instr'}^\ast) \\
+     (\iff S; F; \instr^\ast \stepto S'; F'; {\instr'}^\ast) \\[1ex]
    \begin{array}{lcl@{\qquad}l}
    S; F; E[\TRAP] &\stepto& S; F; \TRAP
-   \end{array}
-   \\ \qquad
-     (\iff E \neq [\_] \wedge E \neq \epsilon~[\_]~\epsilon) \\
+     &(\iff E \neq [\_]) \\
+   S; F; \FRAME_n\{F'\}~\TRAP~\END &\stepto& S; F; \TRAP
+   \end{array} \\
    \end{array}
 
 Reduction terminates when a thread's instruction sequence has been reduced to a :ref:`result <syntax-result>`,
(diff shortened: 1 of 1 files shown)


document/core/exec/runtime.rst
```diff
@@ -562,16 +562,26 @@ Finally, the following definition of *evaluation context* and associated structu
...
 .. math::
+   \begin{array}{l}
    \begin{array}{lcl@{\qquad}l}
    S; F; E[\instr^\ast] &\stepto& S'; F'; E[{\instr'}^\ast]
-     & (\iff S; F; \instr^\ast \stepto S'; F'; {\instr'}^\ast) \\
+   \end{array}
+   \\ \qquad
+     (\iff S; F; \instr^\ast \stepto S'; F'; {\instr'}^\ast) \\
+   \begin{array}{lcl@{\qquad}l}
+   S; F_0; \FRAME_n\{F\}~\instr^\ast~\END &\stepto& S'; F_0; \FRAME_n\{F'\}~\instr'^\ast~\END
```
rossberg: I think you'll need a second rule to propagate a trap out of a frame?

document/core/exec/runtime.rst
```diff
@@ -562,16 +562,26 @@ Finally, the following definition of *evaluation context* and associated structu
...
+   \end{array}
+   \\ \qquad
+     (\iff S; F; \instr^\ast \stepto S'; F'; {\instr'}^\ast) \\
+   \begin{array}{lcl@{\qquad}l}
    S; F; E[\TRAP] &\stepto& S; F; \TRAP
-     & (\iff E \neq [\_]) \\
+   \end{array}
+   \\ \qquad
+     (\iff E \neq [\_] \wedge E \neq \epsilon~[\_]~\epsilon) \\
```
rossberg: Yeah, I think this change to the side condition isn't really meaningful. On one hand, the way these are actually specified as overlapping syntactic productions, both cases are equivalent and thus just confusing. On the other hand, if the grammar was to be reinterpreted as defining disjoint constructions, you'd need a recursive formulation to also cover "\eps (\eps [_] \eps) \eps" and so on.

So I'd prefer to keep the condition as is and rather add a note saying that this also covers cases like the second.

swasey: I hadn't see the problem you point out (and wanted to avoid two notes in a row). Will tweak.

document/core/exec/runtime.rst
swasey: Whoops.

document/core/exec/runtime.rst
```diff
@@ -589,3 +599,7 @@ that is, either a sequence of :ref:`values <syntax-val>` or to a |TRAP|.
       E = (\F64.\CONST~x_1)~[\_]~(\F64.\CONST~x_3)~\F64.\ADD~\F64.\MUL
 
    Moreover, this is the *only* possible choice of evaluation context where the contents of the hole matches the left-hand side of a reduction rule.
+
+.. note::
```
rossberg: Nit: I think it's fine to merge these into one note.

document/core/exec/runtime.rst
rossberg: One more nit: can we name F consistently across rules? Just use F' and F'' for the inner.

rossberg:
Thanks!

PR #638, opened 2018-01-05
Travis-CI integration for building the core HTML spec

littledan:


document/travis-deploy.sh
```diff
@@ -13,13 +13,18 @@ function doCompile {
   make
   cd ../web-api
   make
+  cd ../core
+  make html
```
rossberg: Can you add a TODO here to use plain make?

document/travis-deploy.sh
```diff
@@ -13,13 +13,18 @@ function doCompile {
...
   cd ../../out
   if [[ ! -e js-api ]]; then mkdir js-api; fi
   mv ../document/js-api/_build/html/index.html js-api/index.html
   git add js-api/index.html
   if [[ ! -e web-api ]]; then mkdir web-api; fi
   mv ../document/web-api/_build/html/index.html web-api/index.html
   git add web-api/index.html
+  if [[ ! -e core ]]; then mkdir core; fi
```
rossberg: I'm increasingly confused about this script. The entire doCompile function essentially duplicates the existing Makefile, and in a non-modular manner. Can't we avoid that? The script also shouldn't need to git add things individually.

The existing Makefile creates one _build directory that contains the entire updated contents of the gh-pages branch and then invokes the existing deploy script to update the branch in a single atomic step.

Is there a reason why we cannot simply invoke `make publish` with the right variables instead of replicating everything manually in this script?

document/travis-deploy.sh
littledan: Switched to plain make.

document/travis-deploy.sh
littledan: Deploying this way requires less account privileges than `make publish` does by the use of the ssh key. See https://gist.github.com/domenic/ec8b0fc8ab45f39403dd#get-encrypted-credentials for details. I changed the script here to use `make`; to reduce duplication, maybe we can remove the deployment parts of the Makefile and `deploy.sh` and leave that to this script, unless you still want to do manual deploys sometimes (which aren't so easy with this script)

document/travis-deploy.sh
```diff
@@ -9,18 +9,10 @@ TARGET_BRANCH="gh-pages"
 
 function doCompile {
   # TODO(littledan): Integrate with document/deploy.sh
-  cd document/js-api
+  cd document
```
rossberg: Nit: `(cd document; make)`, removing the `cd ..`

document/travis-deploy.sh
```diff
@@ -9,18 +9,10 @@ TARGET_BRANCH="gh-pages"
...
-  make
-  cd ../../out
-  if [[ ! -e js-api ]]; then mkdir js-api; fi
-  mv ../document/js-api/_build/html/index.html js-api/index.html
-  git add js-api/index.html
-  if [[ ! -e web-api ]]; then mkdir web-api; fi
-  mv ../document/web-api/_build/html/index.html web-api/index.html
-  git add web-api/index.html
   cd ../
+  mv document/_build/* out/
```
rossberg: Change out to _build below? Alternatively, add out to the clean target in the Makefile. But I'd prefer producing just one dir and the _xxx convention for generated dirs.

document/travis-deploy.sh
rossberg: I see, IIUC, this needs checking out as git@github to work? Still would like to unify the two scripts (and keep the ability to manually deploy, especially useful for code reviews on forks). But that isn't high priority -- the make change was the more important bit.

document/travis-deploy.sh
littledan: Fixed

document/travis-deploy.sh
littledan: Done

littledan:
This needs to pass an SSH deploy key up to GitHub to make the commit. At this point, only Travis has the means to decrypt this key; I've deleted the original.

I see why `make` locally is useful, but why would you want to deploy from a fork manually, as opposed to setting up similar auto-deployment on those forks?

rossberg:
> This needs to pass an SSH deploy key up to GitHub to make the commit.

Sure, but that part could be made to work with the other script. But it seems you cannot avoid doing a new clone.

> I see why make locally is useful, but why would you want to deploy from a fork manually, as opposed to setting up similar auto-deployment on those forks?

E.g., to publish changes for code reviews. It’s difficult to review non-trivial stuff merely in source.

littledan:
I see, if the author of the change wants to put it on their personal gh-pages so that reviewers don't need to make locally? V
Got it.

I don't think this script needs a fresh clone. It just needs to have the ssh scheme as a remote.

PR #644, opened 2018-01-09
[spec,test] Host function name specification; tests for function names and caching

littledan:
Normative: Re-exported host functions are numbered

The first patch patch makes the names of exported functions which come from
imported host functions be the index of the import. Although it
was a bit implicit in the original JS.md specification, this are
probably the originally indended behavior.

These semantics seem to be implemented by V8 and SpiderMonkey, but
JSC supports different semantics for exported function names.

----

[test] JS API tests for caching and function names

The second patch add tests against the WebAssembly JS API for
- Memory, tables and functions, including JS FFI functions, are
  cached and reused between imports and exports.
- The name property of functions is determined by the index where
  they are defined; for JS functions, this is the index where
  they are imported.

These tests pass on SpiderMonkey but have certain failures
in JSC and V8.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -306,7 +306,7 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
     1. Return |promise|.
 </div>
 
-<div algorithm>
+<div algorithm="instantiate">
   To <dfn>instantiate a WebAssembly module</dfn> from a {{Module}} |moduleObject| and imports |importObject|, perform the following steps:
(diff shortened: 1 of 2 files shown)


littledan:
I've grouped these two patches together in a single PR in order to follow the pattern of a single commit for spec changes and tests. However, the spec change is just sort of going back to what JS.md did.

littledan:
I was thinking about making the storage more explicit, but it seemed more trouble than it was worth--it didn't seem like that would make things easier for anyone reading the specification. If you're OK with this wording, I'll land the patches as is. Thanks for the quick review!

PR #645, opened 2018-01-09
[spec] Editorial changes

littledan:
- Remove warning that the JS API spec isn't the real spec
- Add appendix listing (known) differences from the previous JS.md spec
- Fix a couple misnamed parameters

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -596,8 +596,8 @@ dictionary TableDescriptor {
 [LegacyNamespace=WebAssembly, Constructor(TableDescriptor descriptor), Exposed=(Window,Worker,Worklet)]
 interface Table {
   unsigned long grow([EnforceRange] unsigned long delta);
-  Function? get([EnforceRange] unsigned long delta);
-  void set([EnforceRange] unsigned long delta, Function? value);
+  Function? get([EnforceRange] unsigned long index);
+  void set([EnforceRange] unsigned long index, Function? value);
   readonly attribute unsigned long length;
 };
 </pre>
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -855,3 +846,9 @@ same class of exception is thrown as for out of memory conditions in JavaScript.
 The particular exception here is implementation-defined in both cases.
 
 Note: ECMAScript doesn't specify any sort of behavior on out-of-memory conditions; implementations have been observed to throw OOMError and to crash. Either is valid here.
+
+<h2 id="appendix-changes">Appendix A: Changes from previous versions</h2>
```
rossberg: I suggest to omit this section. JS.md wasn't an official spec yet, just a design doc, and was itself changing constantly. It seems odd to bother describing the last set of diffs in this document.

littledan: I thought this appendix was based on @lukewagner 's advice, though now I can't find the comment asking for it.

I'm not sure how relevant it is that JS.md wasn't an "official spec". It was used as the basis for implementations in four engines and shared tests between them, so it seems relevant.

rossberg: All the design repo was used to build early implementations. That's normal, but not a concern of the finalised standard. Would e.g. Ecma have bothered listing all the differences between ES2015 and, say, the earlier proxy spec?

It might be useful to track the entire evolution of the API design up to the final spec. But the uniform place to do that would be JS.md. How about adding this section there, for example?

littledan: I think JS.md is a bit different since this was used for four shipping implementations. Anyway, if @lukewagner thinks it is not helpful, I'll just drop this patch from the PR.

flagxor:
As this isn't "editorial" though, you should also change the commit message.
In fact, in the interest of making sure we don't forget to flag this, can you be sure to label this explicitly as a substantive change + conformance change.

littledan:
If we don't care about the appendix, than this patch is subsumed by other editorial patches that already landed.

PR #652, opened 2018-01-26
Rename spectest imports to avoid depending on overloading.

sunfishcode:
This just updates the testsuite, and does not update the interpreter.

### DIFF
--- a/interpreter/host/spectest.ml
+++ b/interpreter/host/spectest.ml
@@ -34,8 +34,10 @@ let lookup name t =
   match Utf8.encode name, t with
   | "print", _ -> ExternFunc (func print (FuncType ([], [])))
   | "print_i32", _ -> ExternFunc (func print (FuncType ([I32Type], [])))
-  | "print_i32_f32", _ -> ExternFunc (func print (FuncType ([I32Type; F32Type], [])))
-  | "print_f64_f64", _ -> ExternFunc (func print (FuncType ([F64Type; F64Type], [])))
+  | "print_i32_f32", _ ->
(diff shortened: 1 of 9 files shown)


rossberg:
Well, Travis should never go red on master, so it'd be good if you could also adjust the interpreter (should be a straightforward change to function host/spectest.ml:lookup).

Also, the suite now no longer tests that the "overloading" pattern at least has to validate. In order to avoid regressing coverage, could you add a simple test that checks that? (Could be done with `assert_unlinkable`, which verifies that linking fails, not validation.)

sunfishcode:
I've now added a test for the "overloading" pattern, including overloading on arity, return types, and kinds. The interpreter code is not straightforward to me.

binji:
@sunfishcode I can take a look at the interpreter part if you like.

sunfishcode:
@binji Thanks!

icefoxen:
Thank you as well; I just ran into this hiccup in my own implementation.  Any idea when this might get merged?

sunfishcode:
I just added two more minor patches here; one is a trivial rename to synchronize the testsuite names with the names now used in the interpreter, and the other updates the JS test harness, which I just noticed also has an implementation of the `spectest` module.

sunfishcode:
This has [unanimous consent](https://github.com/WebAssembly/meetings/blob/master/2018/CG-01-26.md#should-the-spectest-module-in-the-wasm-testsuite-be-implementable-in-wasm) and various review approvals, and I'm not aware of any remaining outstanding issues. Merging, as discussed in the CG meeting today.

PR #653, opened 2018-01-27
Fix typo in JS API index.bs

binji:


document/js-api/index.bs
```diff
@@ -615,7 +615,7 @@ interface Table {
 <div>
 A {{Table}} object represents a single [=table instance=]
 which can be simultaneously referenced by multiple {{Instance}} objects. Each
-{{Table}} object has one internal slots:
+{{Table}} object has two internal slots:
```
jfbastien: "has the following internal slots" so it'll always be right?

binji: True, probably should change the other occurrences in the file if we go that route.

PR #657, opened 2018-01-29
Remove explicit counts in JS API spec

binji:
This way the counts won't go out-of-sync with their associated lists.

See comment here: https://github.com/WebAssembly/spec/pull/653#discussion_r164284778.


jfbastien:
count

littledan:
lgtm

PR #661, opened 2018-02-04
[spec] Clean up instantiation and expression evaluation.

swasey:
Change expression evaluation to reduction. The use of `\stepto^\ast` in the evaluation rule's conclusion had been confusing, being syntax rather than the usual RTC operator.

Fix an incidental typo in `_index-rules`. The instruction reduction judgment lacked an `^\ast`.

Remove unused details from the rules for `\instantiate`. Neither `\globaladdr^\ast` nor `k` was used.

Make `\instantiate` prose and rules agree on when to trap by dropping the generic "otherwise" rule that trapped. The rules no longer relate inputs to outputs where the prose specified "Fail". (Aside: nothing had to change in the embedding, where instantiation is used.)

Make `\instantiate` check, rather than assume, validity of inputs.

Update `\instantiate` prose to use `allocmodule`.

Note the mutual recursion between instantiation's module allocation and execution of global initializers. (Left a sketch of the pre-pass needed to break the recursion in prose step (5) because it's easy to miss or forget that global initializers depend only on imported globals.)

Prefer `\val` to `v` in the rule for `\instantiate`.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1140,5 +1140,6 @@ An :ref:`expression <syntax-expr>` is executed relative to a :ref:`current <exec
    }
 
 .. note::
-   This formal reduction rule is used during :ref:`instantiation <exec-instantiation>` to execute :ref:`global <syntax-global>` initializer expressions as well as :ref:`element <syntax-elem>` and :ref:`data <syntax-data>` segment offset expressions.
-   Expressions constituting :ref:`function <syntax-func>` bodies, by contrast, are executed during function :ref:`invocation <exec-invoke>`.
+   By iterating this reduction rule, we can *evaluate* some expressions to values.
+   This is always possible, for example, with :ref:`constant expressions <valid-constant>`.
+   Expressions constituting :ref:`function <syntax-func>` bodies, by contrast, need not terminate with a value are executed during function :ref:`invocation <exec-invoke>`.
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1126,11 +1126,17 @@ All these notions are made precise in the :ref:`Appendix <soundness>`.
 Expressions
 ~~~~~~~~~~~
 
-An :ref:`expression <syntax-expr>` is executed relative to a :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>` pointing to its containing :ref:`module instance <syntax-moduleinst>`.
+An :ref:`expression <syntax-expr>` is *evaluated* relative to a :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>` pointing to its containing :ref:`module instance <syntax-moduleinst>`.
 
 1. Jump to the start of the instruction sequence :math:`\instr^\ast` of the expression.
 
 2. Execute the instruction sequence.
+
+3. Assert: due to :ref:`validation <valid-expr>`, the top of the stack contains a :ref:`value <syntax-val>`.
+
+4. Pop the :ref:`value <syntax-val>` :math:`\val` from the stack.
+
+The value :math:`\val` is the result of the evaluation.
 
 .. math::
    \frac{
@@ -1140,6 +1146,5 @@ An :ref:`expression <syntax-expr>` is executed relative to a :ref:`current <exec
    }
 
 .. note::
-   By iterating this reduction rule, we can *evaluate* some expressions to values.
-   This is always possible, for example, with :ref:`constant expressions <valid-constant>`.
-   Expressions constituting :ref:`function <syntax-func>` bodies, by contrast, need not terminate with a value are executed during function :ref:`invocation <exec-invoke>`.
+   Evaluation iterates this reduction rule until reaching a value.
+   Expressions constituting :ref:`function <syntax-func>` bodies need not terminate with a value are executed during function :ref:`invocation <exec-invoke>`.
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1147,4 +1147,4 @@ The value :math:`\val` is the result of the evaluation.
 
 .. note::
    Evaluation iterates this reduction rule until reaching a value.
-   Expressions constituting :ref:`function <syntax-func>` bodies need not terminate with a value are executed during function :ref:`invocation <exec-invoke>`.
+   Expressions constituting :ref:`function <syntax-func>` bodies are executed during function :ref:`invocation <exec-invoke>`.
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1126,21 +1126,19 @@ All these notions are made precise in the :ref:`Appendix <soundness>`.
 Expressions
 ~~~~~~~~~~~
 
-An :ref:`expression <syntax-expr>` is *evaluated* relative to a :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>` pointing to its containing :ref:`module instance <syntax-moduleinst>`.
+An :ref:`expression <syntax-expr>` is executed relative to a :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>` pointing to its containing :ref:`module instance <syntax-moduleinst>`.
 
 1. Jump to the start of the instruction sequence :math:`\instr^\ast` of the expression.
 
-2. Execute of the instruction sequence.
-
-3. Assert: due to :ref:`validation <valid-expr>`, the top of the stack contains a :ref:`value <syntax-val>`.
-
-4. Pop the the :ref:`value <syntax-val>` :math:`\val` from the stack.
-
-The value :math:`\val` is the result of the evaluation.
+2. Execute the instruction sequence.
 
 .. math::
    \frac{
-     S; F; \instr^\ast \stepto^\ast S'; F'; v
+     S; F; \instr^\ast \stepto S'; F'; \instr'^\ast
    }{
-     S; F; \instr^\ast~\END \stepto^\ast S'; F'; v
+     S; F; \instr^\ast~\END \stepto S'; F'; \instr'^\ast~\END
    }
+
+.. note::
+   This formal reduction rule is used during :ref:`instantiation <exec-instantiation>` to execute :ref:`global <syntax-global>` initializer expressions as well as :ref:`element <syntax-elem>` and :ref:`data <syntax-data>` segment offset expressions.
+   Expressions constituting :ref:`function <syntax-func>` bodies, by contrast, are executed during function :ref:`invocation <exec-invoke>`.
--- a/document/core/exec/modules.rst
+++ b/document/core/exec/modules.rst
@@ -563,85 +563,95 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
 
 4. For each :ref:`external value <syntax-externval>` :math:`\externval_i` in :math:`\externval^n` and :ref:`external type <syntax-externtype>` :math:`\externtype'_i` in :math:`\externtype_{\F{im}}^n`, do:
 
-   a. Assert: :math:`\externval_i` is :ref:`valid <valid-externval>` with :ref:`external type <syntax-externtype>` :math:`\externtype_i` in store :math:`S`.
+   a. If :math:`\externval_i` is not :ref:`valid <valid-externval>` with an :ref:`external type <syntax-externtype>` :math:`\externtype_i` in store :math:`S`, then:
+
+      i. Fail.
 
    b. If :math:`\externtype_i` does not :ref:`match <match-externtype>` :math:`\externtype'_i`, then:
 
       i. Fail.
 
-5. Let :math:`\moduleinst_{\F{im}}` be the auxiliary module instance :math:`\{\MIGLOBALS~\evglobals(\externval^\ast)\}` that only consists of the imported globals.
+.. _exec-initvals:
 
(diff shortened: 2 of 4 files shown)


document/core/exec/modules.rst
```diff
@@ -679,25 +689,21 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
...
+     &\wedge& (S'; F; \data.\DOFFSET \stepto^\ast S'; F; \I32.\CONST~\X{do}~\END)^\ast \\[1ex]
      &\wedge& (\X{eo} + |\elem.\EINIT| \leq |S'.\STABLES[\tableaddr].\TIELEM|)^\ast \\
      &\wedge& (\X{do} + |\data.\DINIT| \leq |S'.\SMEMS[\memaddr].\MIDATA|)^\ast
    \\[1ex]
-     &\wedge& \globaladdr^\ast = \moduleinst.\MIGLOBALS[|\moduleinst.\MIGLOBALS|-k \slice k] \\
      &\wedge& (\tableaddr = \moduleinst.\MITABLES[\elem.\ETABLE])^\ast \\
      &\wedge& (\memaddr = \moduleinst.\MIMEMS[\data.\DMEM])^\ast \\
      &\wedge& (\funcaddr = \moduleinst.\MIFUNCS[\start.\SFUNC])^?)
-   \\[1ex]
-   \instantiate(S, \module, \externval^n) &=& S; \{\AMODULE~\{\}, \ALOCALS~\epsilon\}; \TRAP
```
rossberg: Not sure I understand how it can work to just eliminate this rule. How is failure turned into a trap now?

document/core/exec/modules.rst
```diff
@@ -563,85 +563,95 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
...
+   c. Push the frame :math:`F_{\F{im}}` to the stack.
 
-7. Push the frame :math:`F` to the stack.
+   d. For each :ref:`global <syntax-global>` :math:`\global_i` in :math:`\module.\MGLOBALS`, do:
 
-8. Let :math:`\globalidx_{\F{new}}` be the :ref:`global index <syntax-globalidx>` that corresponds to the number of global :ref:`imports <syntax-import>` in :math:`\module.\MIMPORTS` (i.e., the index of the first non-imported global).
+      i. :ref:`Execute <exec-expr>` the initializer expression :math:`\global_i.\GINIT`.
 
-9. For each :ref:`global <syntax-global>` :math:`\global_i` in :math:`\module.\MGLOBALS`, do:
+      ii. Assert: due to :ref:`validation <valid-global>`, an expression :math:`\val_i~\END` is now on the top of the stack.
```
rossberg: s/an expression/a value/   (the stack only has values)
s/~\END//  (I'd gloss over "executing" the `END` itself)

Okay, I now see why I had turned evaluation in its own judgement: it's to encapsulate the tedious repetition in the prose. I suggest that we keep the notion of evaluation that returns a value and the prose for it, but still just give the single-step formal rule -- it doesn't require an explicit rule for iterating that, since we can immediately express that with `^\ast`. A simple note might be enough. WDYT?

document/core/exec/modules.rst
```diff
@@ -563,85 +563,95 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
...
 
-   a. Let :math:`\val_i` be the result of :ref:`evaluating <exec-expr>` the initializer expression :math:`\global_i.\GINIT`.
+      iii. Pop the expression from the stack.
 
-   b. Let :math:`\globalidx_i` be the :ref:`global index <syntax-globalidx>` :math:`\globalidx_{\F{new}} + i`.
+   e. Assert: due to :ref:`validation <valid-module>`, the frame :math:`F_{\F{im}}` is now on the top of the stack.
 
-   c. Assert: due to :ref:`validation <valid-global>`, :math:`\moduleinst.\MIGLOBALS[\globalidx_i]` exists.
+   f. Pop the frame from the stack.
```
rossberg: Nit: name the frame :math:`F_{\F{im}}`

document/core/exec/modules.rst
```diff
@@ -563,85 +563,95 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
...
+7. Let :math:`F` be the :ref:`frame <syntax-frame>` :math:`\{ \AMODULE~\moduleinst, \ALOCALS~\epsilon \}`.
 
-   f. Let :math:`\globalinst_i` be the :ref:`global instance <syntax-globalinst>` :math:`S.\SGLOBALS[\globaladdr_i]`.
+8. Push the frame :math:`F` to the stack.
 
-10. For each :ref:`element segment <syntax-elem>` :math:`\elem_i` in :math:`\module.\MELEM`, do:
+9. For each :ref:`element segment <syntax-elem>` :math:`\elem_i` in :math:`\module.\MELEM`, do:
 
-    a. Let :math:`\X{eoval}_i` be the result of :ref:`evaluating <exec-expr>` the expression :math:`\elem_i.\EOFFSET`.
+	a. :ref:`Execute <exec-expr>` the expression :math:`\elem_i.\EOFFSET`.
```
rossberg: Indentation off?

document/core/exec/modules.rst
```diff
@@ -563,85 +563,95 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
 
 4. For each :ref:`external value <syntax-externval>` :math:`\externval_i` in :math:`\externval^n` and :ref:`external type <syntax-externtype>` :math:`\externtype'_i` in :math:`\externtype_{\F{im}}^n`, do:
 
-   a. Assert: :math:`\externval_i` is :ref:`valid <valid-externval>` with :ref:`external type <syntax-externtype>` :math:`\externtype_i` in store :math:`S`.
+   a. If :math:`\externval_i` is not :ref:`valid <valid-externval>` with an :ref:`external type <syntax-externtype>` :math:`\externtype_i` in store :math:`S`, then:
```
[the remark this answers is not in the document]
swasey: I don't think the connection to type soundness at this point is what you describe.

These rules are concerned with the host-WebAssembly boundary. One of the points, I thought, was to show to a host language implementor what (possibly dynamic) checks they must make to benefit from WebAssembly type soundness.

With that view, dynamic type checks in these rules make sense. A dynamically typed host langauge will tag (external) values and can implement the checks. A statically typed host language may be able to discharge the typing side-conditions statically (and avoid tagging values at runtime).

Minor points:

1. The rule has such a check, and I wanted to eliminate the mismatch.

2. `instantiate_module` does no type-checking on its own, so eliminating the checks here would be, I think, a non-local change. What typing assumptions does the `instantiate_module` you envision make? (The current version makes no assumptions and relies on the checks in `instantiate`, unless I am missing something.)

document/core/exec/instructions.rst
```diff
@@ -1126,21 +1126,19 @@ All these notions are made precise in the :ref:`Appendix <soundness>`.
...
    \frac{
-     S; F; \instr^\ast \stepto^\ast S'; F'; v
+     S; F; \instr^\ast \stepto S'; F'; \instr'^\ast
    }{
-     S; F; \instr^\ast~\END \stepto^\ast S'; F'; v
+     S; F; \instr^\ast~\END \stepto S'; F'; \instr'^\ast~\END
    }
+
+.. note::
+   This formal reduction rule is used during :ref:`instantiation <exec-instantiation>` to execute :ref:`global <syntax-global>` initializer expressions as well as :ref:`element <syntax-elem>` and :ref:`data <syntax-data>` segment offset expressions.
```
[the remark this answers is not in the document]
swasey: Ack. I'll make both changes.

document/core/exec/modules.rst
swasey: Regarding evaluation, are you describing

	S; F; instr* ~~>* S'; F'; v
	-------------------------------
	S; F; instr* end ~~> S'; F'; v

If so, I agree. I liked the old evaluation judgment for the reason you give. I simply thought having `^\ast` part of the judgment's syntax was confusing.

Sanity check: Your `sed` script makes things a little less accurate but a little more clear, and will be moot if we return to an evaluation judgment.

Aside: I'd like to add and use an `\evals` macro that looks like `\stepto` but that links to this section. (Analogous to the various `\vdash*` macros.)

document/core/exec/modules.rst
swasey: It is not. The only traps that arise in the prose are those from executing any start function. This change make the rules work the same way.

document/core/exec/modules.rst
swasey: Ack.

document/core/exec/modules.rst
swasey: Ack.

document/core/exec/modules.rst
swasey: I should add that where the prose speaks of "Fail"ing, the judgment simply isn't derivable.

document/core/exec/modules.rst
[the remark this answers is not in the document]
rossberg: Okay, fair enough.

document/core/exec/modules.rst
rossberg: My suggestion is to keep the change you did to the formal rule, but keep the old version of the textual rules. The idea being that in text we have to unfold the `^\ast`, but in the formal rules we don't need to because we can slap it on directly. IOW, "evaluation" is iterated expression execution.

document/core/exec/modules.rst
rossberg: Oh, right, I see. Yeah, the trap expansion was a leftover from a time when this still was an administrative statement with a reduction rule.

document/core/exec/modules.rst
swasey: Ack.

swasey: I addressed this and your other comments in cde59d8. (The changes at this point simply resurrect the old prose, now that there's a note about evaluation.)

document/core/exec/instructions.rst
```diff
@@ -1126,21 +1126,20 @@ All these notions are made precise in the :ref:`Appendix <soundness>`.
 Expressions
 ~~~~~~~~~~~
 
-An :ref:`expression <syntax-expr>` is *evaluated* relative to a :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>` pointing to its containing :ref:`module instance <syntax-moduleinst>`.
+An :ref:`expression <syntax-expr>` is executed relative to a :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>` pointing to its containing :ref:`module instance <syntax-moduleinst>`.
 
 1. Jump to the start of the instruction sequence :math:`\instr^\ast` of the expression.
 
-2. Execute of the instruction sequence.
```
rossberg: I think you need to leave in the other steps, otherwise its not defined at the use sites what the "result" of evaluation is. Likewise, reinstantiate the `*evaluated*` above, because that's the name of the definition used elsewhere.

document/core/exec/instructions.rst
```diff
@@ -1126,21 +1126,20 @@ All these notions are made precise in the :ref:`Appendix <soundness>`.
...
-     S; F; \instr^\ast \stepto^\ast S'; F'; v
+     S; F; \instr^\ast \stepto S'; F'; \instr'^\ast
    }{
-     S; F; \instr^\ast~\END \stepto^\ast S'; F'; v
+     S; F; \instr^\ast~\END \stepto S'; F'; \instr'^\ast~\END
    }
+
+.. note::
+   By iterating this reduction rule, we can *evaluate* some expressions to values.
+   This is always possible, for example, with :ref:`constant expressions <valid-constant>`.
```
rossberg: Not sure what information this sentence conveys, maybe remove it?

document/core/exec/instructions.rst
```diff
@@ -1126,21 +1126,20 @@ All these notions are made precise in the :ref:`Appendix <soundness>`.
...
    \frac{
-     S; F; \instr^\ast \stepto^\ast S'; F'; v
+     S; F; \instr^\ast \stepto S'; F'; \instr'^\ast
    }{
-     S; F; \instr^\ast~\END \stepto^\ast S'; F'; v
+     S; F; \instr^\ast~\END \stepto S'; F'; \instr'^\ast~\END
    }
+
+.. note::
+   By iterating this reduction rule, we can *evaluate* some expressions to values.
```
rossberg: Nit: use passive form instead of "we". Also, move definitional emphasis on "evaluate" to the intro paragraph.

document/core/exec/instructions.rst
swasey: Okay, I see what you have in mind.

document/core/exec/instructions.rst
swasey: I restored the original intro (modulo typos) and updated the note (6d618ce).

document/core/exec/instructions.rst
```diff
@@ -1130,17 +1130,21 @@ An :ref:`expression <syntax-expr>` is *evaluated* relative to a :ref:`current <e
...
-     S; F; \instr^\ast \stepto^\ast S'; F'; v
+     S; F; \instr^\ast \stepto S'; F'; \instr'^\ast
    }{
-     S; F; \instr^\ast~\END \stepto^\ast S'; F'; v
+     S; F; \instr^\ast~\END \stepto S'; F'; \instr'^\ast~\END
    }
+
+.. note::
+   Evaluation iterates this reduction rule until reaching a value.
+   Expressions constituting :ref:`function <syntax-func>` bodies need not terminate with a value are executed during function :ref:`invocation <exec-invoke>`.
```
rossberg: Word missing? But I would just drop the "need not terminate with a value part".

swasey: Fixed. Thanks!

rossberg:
Thanks!

PR #665, opened 2018-02-12
Update wasm-module-builder.js

gahaas:
Pull in the latest changes from the V8 repository.

### DIFF
--- a/test/harness/wasm-module-builder.js
+++ b/test/harness/wasm-module-builder.js
@@ -69,14 +69,12 @@ class Binary extends Array {
     // Emit section name.
     this.emit_u8(section_code);
     // Emit the section to a temporary buffer: its full length isn't know yet.
-    const section = new Binary;
+    let section = new Binary;
     content_generator(section);
     // Emit section length.
     this.emit_u32v(section.length);
     // Copy the temporary buffer.
-    for (const b of section) {
-      this.push(b);
-    }
+    for (var i = 0; i < section.length; i++) this.push(section[i]);
   }
 }
 
@@ -123,9 +121,25 @@ class WasmFunctionBuilder {
     return this;
   }
 
+  getNumLocals() {
+    let total_locals = 0;
+    for (let l of this.locals || []) {
+      for (let type of ["i32", "i64", "f32", "f64", "s128"]) {
+        total_locals += l[type + "_count"] || 0;
+      }
+    }
+    return total_locals;
+  }
+
   addLocals(locals, names) {
-    this.locals = locals;
-    this.local_names = names;
+    const old_num_locals = this.getNumLocals();
+    if (!this.locals) this.locals = []
+    this.locals.push(locals);
+    if (names) {
+      if (!this.local_names) this.local_names = [];
+      const missing_names = old_num_locals - this.local_names.length;
+      this.local_names.push(...new Array(missing_names), ...names);
+    }
     return this;
   }
 
@@ -155,9 +169,11 @@ class WasmModuleBuilder {
     this.imports = [];
     this.exports = [];
     this.globals = [];
+    this.exceptions = [];
     this.functions = [];
     this.function_table = [];
-    this.function_table_length = 0;
+    this.function_table_length_min = 0;
+    this.function_table_length_max = 0;
     this.function_table_inits = [];
     this.segments = [];
     this.explicit = [];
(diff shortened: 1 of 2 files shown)


test/harness/wasm-module-builder.js
```diff
@@ -69,14 +69,12 @@ class Binary extends Array {
...
-    const section = new Binary;
+    let section = new Binary;
     content_generator(section);
     // Emit section length.
     this.emit_u32v(section.length);
     // Copy the temporary buffer.
-    for (const b of section) {
-      this.push(b);
-    }
+    for (var i = 0; i < section.length; i++) this.push(section[i]);
```
rossberg: Why can't we use for-of? Nit: at least use let instead of var.

test/harness/wasm-module-builder.js
```diff
@@ -155,9 +169,11 @@ class WasmModuleBuilder {
     this.imports = [];
     this.exports = [];
     this.globals = [];
+    this.exceptions = [];
```
rossberg: I don't think we should push code concerned with proposed extensions into spec master.

test/harness/wasm-module-builder.js
```diff
@@ -171,8 +187,8 @@ class WasmModuleBuilder {
     return this;
   }
 
-  addMemory(min, max, exp) {
-    this.memory = {min: min, max: max, exp: exp};
+  addMemory(min, max, exp, shared) {
```
rossberg: Same here, this is still a proposal.

test/harness/wasm-module-builder.js
```diff
@@ -198,8 +214,9 @@ class WasmModuleBuilder {
   }
 
   addType(type) {
-    // TODO: canonicalize types?
     this.types.push(type);
+    var pl = type.params.length;  // should have params
```
rossberg: Nit: let

test/harness/wasm-module-builder.js
```diff
@@ -210,6 +227,13 @@ class WasmModuleBuilder {
     return glob;
   }
 
+  addException(type) {
```
rossberg: Same here.

test/harness/wasm-module-builder.js
```diff
@@ -232,9 +256,9 @@ class WasmModuleBuilder {
     return this.num_imported_globals++;
   }
 
-  addImportedMemory(module = "", name, initial = 0, maximum) {
+  addImportedMemory(module = "", name, initial = 0, maximum, shared) {
```
rossberg: Same here.

test/harness/wasm-module-builder.js
```diff
@@ -336,7 +364,12 @@ class WasmModuleBuilder {
             section.emit_u8(imp.mutable);
           } else if (imp.kind == kExternalMemory) {
             var has_max = (typeof imp.maximum) != "undefined";
-            section.emit_u8(has_max ? 1 : 0); // flags
+            var is_shared = (typeof imp.shared) != "undefined";
```
rossberg: Same here.

test/harness/wasm-module-builder.js
```diff
@@ -381,7 +416,13 @@ class WasmModuleBuilder {
       binary.emit_section(kMemorySectionCode, section => {
         section.emit_u8(1);  // one memory entry
         const has_max = wasm.memory.max !== undefined;
-        section.emit_u32v(has_max ? kResizableMaximumFlag : 0);
+        const is_shared = wasm.memory.shared !== undefined;
```
rossberg: And here.

test/harness/wasm-module-builder.js
```diff
@@ -489,6 +530,20 @@ class WasmModuleBuilder {
       });
     }
 
+    // Add exceptions.
```
rossberg: And here.

test/harness/wasm-module-builder.js
```diff
@@ -513,6 +566,9 @@ class WasmModuleBuilder {
             if (l.f64_count > 0) {
               local_decls.push({count: l.f64_count, type: kWasmF64});
             }
+            if (l.s128_count > 0) {
```
rossberg: And here.

test/harness/wasm-module-builder.js
gahaas: This PR would bring the wasm-module-builder back in sync with the V8 version of it. I can clean up the wasm-module-builder in V8 and do the sync afterwards, but I cannot remove the support of exceptions because we need it to test exceptions in V8.
If you say that there is no value in keeping the two versions in sync I will just close this PR.

rossberg: Well, I think there *is* value in keeping them in sync, but only up to experimental extensions. For reasons of process, those shouldn't be in the spec repo, even though I understand that that makes it harder to merge desirable changes.

xtuc:
This seems outdated and will eventually be replaced by #944

binji:
Agreed, let's close this.

PR #666, opened 2018-02-13
[spec] Cite mechanisation paper

rossberg:
Cite Conrad Watt's CPP 2018 paper describing the Isabelle mechanisation of the soundness proof.


rossberg:
@conrad-watt, can you have a look?

conrad-watt:
LGTM!

PR #672, opened 2018-02-13
Fixing broken formatting in katex version.

flagxor:


flagxor:
Not sure what's going on with https://webassembly.github.io/spec/core/bikeshed/index.html
Weirdly gh-pages checked out locally seems to be correct.
But I figure dropping dead links + clean build can't hurt.

Issue #684, opened 2018-02-14
[spec] TypeError during instantiation


gahaas:
Hi,
I found an inconsistency between the JS API spec and the core spec tests and wonder how it can be resolved properly.
In the JS API spec in https://webassembly.github.io/spec/js-api/index.html#instantiate-a-webassembly-module it says in Step 4.2 that "If Type(o) is not Object, throw a TypeError exception." 
The spec test in https://github.com/WebAssembly/spec/blob/master/test/core/imports.wast#L563, however, requires a LinkError. First I thought it's an issue with the spec test, and I wanted to change it to expect a TypeError. However, the core spec tests don't even have the concept of TypeErrors. In my opinion this makes sense, so I wonder why the JS API spec requires a TypeError in Step 4.2.

I think it would be best if we change the JS API and require a LinkError in Step 4.2 instead of the TypeError.

rossberg:
This was specced as a TypeError because it is a JS-side issue: the import object that you pass to instantiate must have objects as its properties, to represent the inner namespaces. The core semantics does not care how the two-level names are interpreted, a LinkError only occurs if the module is given something of the wrong type. Arguably that is a subtle distinction, and another design choice might have been just as appropriate.

The problem actually is with the JS test harness that the interpreter generates, and how it simply uses the registry as the import object. An easy fix would be to wrap a proxy around it that raises TypeError instead of returning undefined for an unknown property.

gahaas:
I don't understand what you mean with this easy fix. Did you actually want to write 
"An easy fix would be to wrap a proxy around it that returns an EMPTY OBJECT instead of returning undefined for an unknown property."? I don't see how it would help to raise a TypeError.

rossberg:
Yes, you are right, sorry. Specifically, try adding these two lines after the declaration of `registry`:
```
let handler = {get(name) { return (name in registry) ? registry[name] : {}}
registry = new Proxy(registry, handler)
```

gahaas:
I created a PR: https://github.com/WebAssembly/spec/pull/685

littledan:
I'd be fine with making this a LinkError rather than a TypeError. JS always throws TypeErrors for these sorts of things because there just aren't many error classes. I don't think there's any sort of general convention that JS things which are outside of the ES specification always throw one of the built-in error  types.

I don't quite understand why it's OK to use the JS API to create a module in these tests, but then it's not OK to expect a TypeError in these error cases.

rossberg:
[...]

PR #685, opened 2018-02-14
[test] Wrap the default imports in a proxy to avoid TypeErrors

gahaas:
This PR fixes the problem discussed in https://github.com/WebAssembly/spec/issues/684. We wrap the default imports in the core spec tests to avoid TypeErrors, because in the core WebAssembly spec TypeErrors have no place.

Drive-by change: add newlines I forgot in the last PR.

### DIFF
--- a/interpreter/script/js.ml
+++ b/interpreter/script/js.ml
@@ -22,12 +22,12 @@ let harness =
   "  table: new WebAssembly.Table({initial: 10, maximum: 20, element: 'anyfunc'}),\n" ^
   "  memory: new WebAssembly.Memory({initial: 1, maximum: 2})\n" ^
   "};\n" ^
-  "let spectest_handler = {\n" ^
+  "let handler = {\n" ^
   "  get(target, prop) {\n" ^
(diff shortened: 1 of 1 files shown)


interpreter/script/js.ml
```diff
@@ -10,20 +10,24 @@ let harness =
...
+  "  print_f64: console.log.bind(console),\n" ^
+  "  global_i32: 666,\n" ^
+  "  global_f32: 666,\n" ^
+  "  global_f64: 666,\n" ^
+  "  table: new WebAssembly.Table({initial: 10, maximum: 20, element: 'anyfunc'}),\n" ^
+  "  memory: new WebAssembly.Memory({initial: 1, maximum: 2})\n" ^
   "};\n" ^
-  "\n" ^
-  "let registry = {spectest};\n" ^
+  "let spectest_handler = {\n" ^
```
rossberg: Nit: just name it handler, it wraps the registry, not the spectest module in it.

gahaas: Done.

Issue #622, opened 2017-12-12
Multiple exports under same name


pepyakin:
[Spec states](https://webassembly.github.io/spec/valid/modules.html#valid-module) that each export should have a different name. 

However, it looks like spec tests are relying on that module `spectest` publishes several functions with different signatures under the same name.

https://github.com/WebAssembly/spec/blob/fb7e7e1e381ffc283c923a87fdfea5ebbd213737/test/core/imports.wast#L36-L39

Why is that? Is `spectest` module is sort of magic?

binji:
It's only as magic as JavaScript :-)

`spectest.print` is just one export, but it happens to be a function that can print as many arguments as it is given with whichever types it is given. You could provide it in the import object as:

```
let importObj = {spectest: {print: console.log}};
```

eholk:
You can only have one export of a given name, but you can something as many times as you want.

This is useful especially in the case of importing JavaScript functions, as these will accept values of different types. For example, `console.info` can log any type object, but Wasm requires the types to match when you call a function. That means if you want to call `console.info` at different types, you must import it for each type you want.

That's what's going on here with `spectest` as well.

pepyakin:
Ah, I see.

I'm working on a WebAssembly interpreter implementation and it is not related to JS at all. This implementation enforces all the exports that is provided by any module (including "host" modules). I thought that I can implement `spectest` as ordinary module, but apparently I can't.

It seems that I had a wrong model of wasm: I thought that every import should be satisfied by somekind of a module (even if derived from `importObj` in JS), but it seems that is not the case.

jfbastien:
Internal to wasm you're correct, all named functions have to be strictly typed. That's not the case for imports from the embedder, as evidenced by the spec interpreter and JS, but your embedder could force that restriction if you wanted. It just seems fairly restrictive to do so, even for a C++ embedding I'd allow variadic things like this by passing arguments / returns arrays along with a signature.

pepyakin:
That makes sense, thank you!

Currently API I working on roughly works like

```
vm.add_host_module("env", env_host_module);
vm.add_module("main", main_module);
```

Now it became clear to me, that I probably want something similar to JS:

```
vm.instantiate(
  main_module,
  {
    "env": env_host_module_instance
  }
);
```

I'm wondering why spec avoids name matching stuff?

binji:
[...]

Issue #650, opened 2018-01-18
spectest module not implementable in wasm


sunfishcode:
The testsuite uses functions named `print` which are overloaded by their argument type. For example:

https://github.com/WebAssembly/spec/blob/a7e226a92e660a3d5413cfea4269824f513259d2/test/core/imports.wast#L36

This limits the ways in which the testsuite can be used. A non-Web wasm engine may not otherwise need a JS interpreter, function overloading (which would require name mangling in some implementations), or a `spectest` builtin module (which would be extra needless surface area in production). It would be useful to have the option to implement `spectest` in wasm itself, however that's currently not possible with the current interface (as discussed in https://github.com/WebAssembly/spec/issues/622).

Replacing `spectest`'s `print` with `print_i32`, `print_i64`, and so on would eliminate these issues.

binji:
Agreed this would be nice. Also, as currently defined [here](https://github.com/WebAssembly/spec/blob/master/interpreter/host/spectest.ml#L10), `"global"` is also overloaded. It seems that the tests don't actually make use of this, however.

rossberg:
Hm, the idea was that this allows testing that engines can actually handle that case, because it's legal from the Wasm perspective -- at least it needs to validate, even if it cannot be linked in all embeddings.

Any suggestion how we can serve both purposes? Perhaps isolate the overloaded case in a separate test?

sunfishcode:
Requiring that in core wasm would require an otherwise needless feature in a wasm engine that only intends to run wasm modules.

It seems like this is something that should apply to specific embeddings, such as the Web embedding, and it can be tested for there as well.

rossberg:
Well, validation is supposed to be the same everywhere.

rossberg:
But I guess we could turn that into a validation-only test.

jfbastien:
This forces an embedder to call functions with an explicit signature. Is the objection that this is more than what's needed, or that it can cause a double-dispatch, or can't use native calling conventions? I implemented some of this and it doesn't seem overly burdensome.

sunfishcode:
@rossberg I think that's right; a module with two imports with the same name and different signatures can still pass validation, but engines can issue a link error if they don't support it. A validation-only test for this sounds fine, and then we can remove the overloading in `print` and `global`.

sunfishcode:
@jfbastien Could I ask you to clarify what you're responding to here, and what you implemented?

jfbastien:
[...]

PR #687, opened 2018-02-14
[spec] Clarify overloaded imports

rossberg:
Add a note clarifying that import names are not unique, and remove a contradicting statement. Addresses #622.

### DIFF
--- a/document/core/syntax/modules.rst
+++ b/document/core/syntax/modules.rst
@@ -367,6 +367,6 @@ In each index space, the indices of imports go before the first index of any def
    Unlike export names, import names are not necessarily unique.
    It is possible to import the same |IMODULE|/|INAME| pair multiple times;
    such imports may even have different type descriptions, including different kinds of entities.
-   A module with such imports can still be instantiated as long as the :ref:`embedder <embedder>` is either agnostic to a specific typing difference or it takes the expected types into account when resolving imports.
+   A module with such imports can still be instantiated depending on the specifics of how an :ref:`embedder <embedder>` allows resolving and supplying imports.
    However, embedders are not required to support such overloading,
    and a WebAssembly module itself cannot implement an overloaded name.
--- a/document/core/syntax/modules.rst
+++ b/document/core/syntax/modules.rst
@@ -307,7 +307,7 @@ The |MEXPORTS| component of a module defines a set of *exports* that become acce
      \EDGLOBAL~\globalidx \\
    \end{array}
 
-Each export is identified by a unique :ref:`name <syntax-name>`.
+Each export is labeled by a unique :ref:`name <syntax-name>`.
 Exportable definitions are :ref:`functions <syntax-func>`, :ref:`tables <syntax-table>`, :ref:`memories <syntax-mem>`, and :ref:`globals <syntax-global>`,
 which are referenced through a respective descriptor.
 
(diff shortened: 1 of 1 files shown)


document/core/syntax/modules.rst
```diff
@@ -354,7 +354,7 @@ The |MIMPORTS| component of a module defines a set of *imports* that are require
      \IDGLOBAL~\globaltype \\
    \end{array}
 
-Each import is identified by a two-level :ref:`name <syntax-name>` space, consisting of a |IMODULE| name and a unique |INAME| for an entity within that module.
+Each import is identified by a two-level :ref:`name <syntax-name>` space, consisting of a |IMODULE| name and a |INAME| for an entity within that module.
```
sunfishcode: I also suggest changing "is identified by" to "utilizes", or something, because two imports with the same names may still have distinct identities: they have different indices and/or kinds, and they may be resolved to different entities within the same instantiation.

document/core/syntax/modules.rst
```diff
@@ -363,3 +363,9 @@ In each index space, the indices of imports go before the first index of any def
 
 .. note::
    In the current version of WebAssembly, only *immutable* globals may be imported.
+
+   Unlike export names, import names are not necessarily unique.
+   It is possible to import the same |IMODULE|/|INAME| pair multiple times.
+   Even if such imports have different types a module can still be instantiated by :ref:`embedders <embedder>` for which the type difference is irrelevant or that take the expected types into account when resolving imports.
```
sunfishcode: I find this wording difficult to understand. I think there are two surprising things to highlight: it's surprising that embedders *can* overload on types, since wasm itself can't do that. And it's surprising that embedders can overload on anything *other* than types too, since that's somewhat esoteric.

document/core/syntax/modules.rst
```diff
@@ -363,3 +363,9 @@ In each index space, the indices of imports go before the first index of any def
...
+   However, embedders are not required to support such "overloading",
```
sunfishcode: For my reference, do the quotes here mean this word is being misused?

document/core/syntax/modules.rst
rossberg: I assume by the second thing you mean being different kinds of entities, like function vs global? Semantically, that still is simply a difference in type, but I changed wording to call out that case explicitly.

document/core/syntax/modules.rst
rossberg: Changed to "labeled".

document/core/syntax/modules.rst
rossberg: Removed the quotes. It was more because it all depends on the host's interpretation. For example, with JavaScript it isn't really overloading, because the name denotes the same single object (and it fails in all cases where it doesn't). Similarly, another hypothetical embedder could interpret the names as a reference to a single generic definition, with the types as different instantiation parameters.

document/core/syntax/modules.rst
sunfishcode: I also mean like your comment [here](https://github.com/WebAssembly/spec/issues/650#issuecomment-362879847) about resolving imports "positionally". My understanding of core wasm here is that import resolution is just unconstrained; there are no rules about what information embedder may consult to resolve an import. Two imports in a module could be identical in every way, except their position in the import sequence, and still be resolved to different entities, in the same instantiation.

rossberg: Yes, I see. Okay, I dropped the attempt to enumerate possible interpretations/implementations in favour of a more general statement. PTAL

PR #691, opened 2018-02-15
Make katex conversion progress visible regardless of terminal.

flagxor:


flagxor:
Committing TBR (ensures katex conversion works even on travis).

PR #695, opened 2018-02-15
[spec] ~ abstract and href="index.html#" links

ericprud:
~ make ref to sybling docs be a complete sentence
~ s/href="index.html#..."/href="#..."/g


rossberg:
@ericprud, is this PR still relevant?

Ms2ger:
Not in this form, definitely – it's editing the generated spec documents, so any change would be overwritten at the next edit.

I proposed the change to the abstract at https://github.com/tabatkins/bikeshed/pull/1660. I haven't tried to figure out where the "index.html" change would need to be made, but that document is due to be removed anyway.

PR #705, opened 2018-02-17
[spec] Editorial: Simplify introduction

littledan:
- Remove irrelevant issues and notes
- Clean up wording to indicate more clearly what it is for

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -603,8 +603,8 @@ dictionary TableDescriptor {
 [LegacyNamespace=WebAssembly, Constructor(TableDescriptor descriptor), Exposed=(Window,Worker,Worklet)]
 interface Table {
   unsigned long grow([EnforceRange] unsigned long delta);
-  Function? get([EnforceRange] unsigned long delta);
-  void set([EnforceRange] unsigned long delta, Function? value);
+  Function? get([EnforceRange] unsigned long index);
+  void set([EnforceRange] unsigned long index, Function? value);
   readonly attribute unsigned long length;
 };
 </pre>
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -166,7 +166,7 @@ spec:ecmascript; type:interface; for:ECMAScript; text:ArrayBuffer
 spec:promises-guide-1; type:dfn; text:resolve
 </pre>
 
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -166,16 +166,7 @@ spec:ecmascript; type:interface; for:ECMAScript; text:ArrayBuffer
...
-
-This API is, initially, the only API for accessing WebAssembly [[WEBASSEMBLY]] from the web platform, through a bridge to explicitly construct modules from ECMAScript [[ECMASCRIPT]].
-
-Issue: In future versions, WebAssembly may
-be loaded and run directly from an HTML `<script type='module'>` tag—and
-any other Web API that loads ES6 modules via URL—as part of
-[ES6 Module integration](https://github.com/WebAssembly/design/issues/1087).)
-
-Note: WebAssembly JS API declaration file for TypeScript can be found [here](https://github.com/01alchemist/webassembly-types/blob/master/webassembly.d.ts) which enable autocompletion and make TypeScript compiler happy.
+This API privides a way to access WebAssembly [[WEBASSEMBLY]] from the web platform or other JavaScript environments such as Node.js, through a bridge to explicitly construct modules from ECMAScript [[ECMASCRIPT]].
```
rossberg: Nit: Is there even a need to call out individual JS platforms? Why not simply say "to access WebAssembly modules and objects from ECMAScript"?

littledan: This is an introduction, which seems like an appropriate place to provide a little context. The Web and Node.js have been influential in the WebAssembly design process, including the decision to make a JS API in the first place. What is the downside of including this explanation?

rossberg: There are many environments in which JS runs nowadays. While the web stands out for historical reasons, a singleton enumeration of other clients makes them look very privileged, which a standard document should avoid. Not even the ECMAScript spec itself contains such a list.

littledan: The ECMAScript spec has tons of references to the web platform. Anyway, shortened the introduction to remove a mention of the Web or Node.js.

rossberg: Yes, as I said, the web stands out, what I meant is it doesn't mention any other. Anyway, LGTM.

PR #706, opened 2018-02-17
[spec] Normative: Clarify that JSAPI i32 conversions are signed

### LINKED ISSUE #647: Clarify signed/unsigned interpretation for ToJSValue with i32 arguments
This is more of a question than an issue:
When calling an exported wasm function from JavaScript that returns an `i32`, is it guaranteed to be interpreted as signed i32 two's complement integer?

Is this specified somewhere? The only thing I could find was the **ToJSValue** algorithm which says:

> If w is of the form 𝗂𝟥𝟤.𝖼𝗈𝗇𝗌𝗍 i32, return the Number value for i32.

https://webassembly.github.io/spec/js-api/index.html#tojsvalue

If I understand it correctly, `i32` is not inherently signed or unsigned in general (depends on instruction, `i32.div_u` and `i32.div_s` for example. So maybe the sign interpretation of **ToJSValue** could be stated more clearly?

cc @ColinEberhardt @xtuc

littledan:
Includes a test to verify. Test passes on V8.

Closes #647

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -154,6 +154,7 @@ urlPrefix: https://webassembly.github.io/spec/core/multipage/; spec: WebAssembly
         text: 𝗀𝗅𝗈𝖻𝖺𝗅
     text: global type; url: syntax/types.html#syntax-globaltype
     text: address; url: exec/runtime.html#addresses
+    text: signed_32; url: exec/numerics.html#aux-signed
 </pre>
 
 <pre class='link-defaults'>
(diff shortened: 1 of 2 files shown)


littledan:
It's unclear to me whether we should actually use ToInt32 here (as JS.md did) or more abstract wording. The reasoning for more abstract wording is that ToInt32 is a Number -> Number function, where here we are actually trying to go i32 -> Number.

document/js-api/index.bs
```diff
@@ -786,7 +786,7 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not
 The algorithm <dfn>ToJSValue</dfn>(|w|) coerces a [=WebAssembly value=] to a JavaScript value performs the following steps:
 
 Assert: |w| is not of the form [=𝗂𝟨𝟦.𝖼𝗈𝗇𝗌𝗍=] |i64|.
-1. If |w| is of the form [=𝗂𝟥𝟤.𝖼𝗈𝗇𝗌𝗍=] |i32|, return [=the Number value=] for |i32|.
+1. If |w| is of the form [=𝗂𝟥𝟤.𝖼𝗈𝗇𝗌𝗍=] |i32|, return [=the Number value=] for |i32| when interpreted as a signed integer.
```
rossberg: How about invoking [signed_32](http://webassembly.github.io/spec/core/exec/numerics.html#aux-signed)(i32) explicitly, to make precise what that means?

littledan: That function is perfect for this; thanks for the cross-reference. Switched to use it.

Issue #596, opened 2017-11-06
[test] Split up jsapi.js


rossberg:
Instead of one monolithic test we should at least have separate tests for each "class" and for each main function in the API.

littledan:
I believe @ms2ger has done this in the WPT repository. Probably the needed work item at this point is to upstream the new tests into this repository.

Ms2ger:
The file has been removed (https://github.com/WebAssembly/spec/pull/883), so this issue can be closed.

PR #707, opened 2018-02-17
[spec] Normative: Add i64<->BigInt conversion in JS API

littledan:
See the JS-BigInt-integration proposal here:

https://github.com/WebAssembly/JS-BigInt-integration

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -89,6 +89,8 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
         text: CreateArrayFromList; url: sec-createarrayfromlist
         text: GetMethod; url: sec-getmethod
         text: IterableToList; url: sec-iterabletolist
+        text: ToBigInt64; url: #sec-tobigint64
+        text: BigInt; url: #sec-ecmascript-language-types-bigint-type
     type: abstract-op
         text: CreateMethodProperty; url: sec-createmethodproperty
 urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: dfn
@@ -172,6 +174,7 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     text: current frame; url: exec/conventions.html#exec-notation-textual
     text: module; for: frame; url: exec/runtime.html#syntax-frame
     text: memaddrs; for: moduleinst; url: exec/runtime.html#syntax-moduleinst
+    text: signed_64; url: exec/numerics.html#aux-signed
     text: sequence; url: syntax/conventions.html#grammar-notation
 urlPrefix: https://heycam.github.io/webidl/; spec: WebIDL
     type: dfn
@@ -348,8 +351,11 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
             1. Let |externfunc| be the [=external value=] [=external value|func=] |funcaddr|.
             1. [=list/Append=] |externfunc| to |imports|.
         1. If |externtype| is of the form [=global=] <var ignore>mut</var> |valtype|,
-            1. If [=Type=](|v|) is Number,
-                1. If |valtype| is [=i64=], throw a {{LinkError}} exception.
+            1. If [=Type=](|v|) is Number or BigInt,
+                1. If |valtype| is [=i64=] and [=Type=](|v|) is Number,
+                    1. Throw a {{LinkError}} exception.
+                1. If |valtype| is not [=i64=] and [=Type=](|v|) is BigInt,
+                    1. Throw a {{LinkError}} exception.
                 1. Let |value| be [=ToWebAssemblyValue=](|v|, |valtype|).
                 1. Let |store| be the [=surrounding agent=]'s [=associated store=].
                 1. Let (|store|, |globaladdr|) be [=global_alloc=](|store|, [=const=] |valtype|, |value|).
@@ -871,7 +877,6 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     1. If |v| is undefined,
         1. Let |value| be [=DefaultValue=](|valuetype|).
     1. Otherwise,
-        1. If |valuetype| is [=i64=], throw a {{TypeError}} exception.
         1. Let |value| be [=ToWebAssemblyValue=](|v|, |valuetype|).
     1. If |mutable| is true, let |globaltype| be [=var=] |valuetype|; otherwise, let |globaltype| be [=const=] |valuetype|.
     1. Let |store| be the current agent's [=associated store=].
@@ -884,8 +889,6 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     The algorithm <dfn>GetGlobalValue</dfn>({{Global}} |global|) performs the following steps:
     1. Let |store| be the current agent's [=associated store=].
     1. Let |globaladdr| be |global|.\[[Global]].
-    1. Let |globaltype| be [=global_type=](|store|, |globaladdr|).
-    1. If |globaltype| is of the form <var ignore>mut</var> [=i64=], throw a {{TypeError}}.
     1. Let |value| be [=global_read=](|store|, |globaladdr|).
     1. Return [=ToJSValue=](|value|).
 </div>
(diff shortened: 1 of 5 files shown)


jfbastien:
This is a very significant change if applied towards the current working draft. Is your intent to keep this PR open until the public draft has shipped, and we can easily move non-editorial things into the spec?

littledan:
@jfbastien I was imagining that this would stay as a PR for a while (at least until it meets the implementation requirements of WebAssembly proposals, for example). Is there something more that I should write up to frame this as a proper proposal?

I don't have a strong opinion about how this should relate to the W3C process, but for a comparison, ECMAScript continues development even during the several month IP opt out period, developing in the context of a working draft which follows the regular, official version cut. I don't see why this sort of model couldn't work for Wasm as well.

jfbastien:
We've usually followed this process:
https://github.com/WebAssembly/meetings/blob/master/process/phases.md
(with edits in a PR that @flagxor has to review still)

So you only go to the spec repo after having a fork in a separate repo.

littledan:
OK, this has a fork with the new spec in it (https://github.com/littledan/spec/tree/bigint) and a tracking issue in the design repo (WebAssembly/design#1172). The commit message here describes at a high level what this proposal is doing.

Should I put this commit message into a README.md in that repository, or is it sufficient in this location at the top of the thread? Either way, is the next step is to propose this feature for some sort of stage advancement at a WebAssembly Community Group meeting?

jfbastien:
Tracking issues are the ones that I tag as such. The issue you link to is the "someone has an idea" step. Sorry, I know the process isn't great yet! Gotta improve what's written down.

Process would be that you put it on the agenda for the next CG meeting, and we'll poll to officially create a fork repo for it. Then we can transfer your repo to the organization, like TC39 does. Should pass easy, but I still want us to follow that process so we don't play favorites. It's also a really good way to keep the wider CG informed.

littledan:
Oh, I see, the tracking issue and the fork have to be specifically approved. Makes sense. I didn't realize those words had special meanings here. I guess I'll put something on the CG agenda.

(In TC39, there are two processes for making changes: A formal staged proposal and a PR. We still don't play favorites, but sometimes smaller changes go through less paperwork.)

xtuc:
Thanks @littledan! Sorry for my ignorance, how can you see the rendered version of the document?

test/js-api/jsapi.js
```diff
@@ -881,4 +881,57 @@ test(() => {
...
+        .exportFunc();
+
+    return builder.toBuffer();
+})();
+
+test(() => {
+  let module = new WebAssembly.Module(exportingModuleLongIdentityFn);
+  let instance = new WebAssembly.Instance(module);
+
+  let value = 2n ** 63n;
```
xtuc: That assumes the host has the BigInt proposal support?

xtuc (commented):
Does that means that a BigInt support is now required to be WebAssembly spec compliant?

Can we make it optional?

document/js-api/index.bs
```diff
@@ -785,10 +788,12 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not
 <div algorithm>
 The algorithm <dfn>ToJSValue</dfn>(|w|) coerces a [=WebAssembly value=] to a JavaScript value performs the following steps:
 
-Assert: |w| is not of the form [=𝗂𝟨𝟦.𝖼𝗈𝗇𝗌𝗍=] |i64|.
-1. If |w| is of the form [=𝗂𝟥𝟤.𝖼𝗈𝗇𝗌𝗍=] |i32|, return [=the Number value=] for |i32|.
-1. If |w| is of the form [=𝖿𝟥𝟤.𝖼𝗈𝗇𝗌𝗍=] |f32|, return [=the Number value=] for |f32|.
-1. If |w| is of the form [=𝖿𝟨𝟦.𝖼𝗈𝗇𝗌𝗍=] |f64|, return [=the Number value=] for |f64|.
+1. If |w| is of the form [=𝗂𝟨𝟦.𝖼𝗈𝗇𝗌𝗍=] |i64|,
+    1. Let |v| be [=signed_64=](|i64|).
+    1. Return a [=BigInt=] representing the mathematical value |v|.
```
xtuc: Does "mathematical value" have a special meaning? I don't think we need to mention math, It's just a value?

rossberg:
@xtuc, clearly we won't make this into a part of the Wasm standard before BigInt has become a part of JS standard (it's getting there). At that point, though, there seems to be little point in making it optional.

littledan:
@rossberg My hope is that this can be ready for implementations to ship at the same time as they are shipping BigInt.

I don't have a strong opinion about when it lands in the standard, but it really wouldn't be unusual at all for the Wasm standard to land while BigInt is at Stage 3 (given all the standards which make normative references to "Editor's Draft"s, this would be nothing--Stage 3 is something like "Candidate Recommendation").

test/js-api/jsapi.js
littledan: Yes, it does. When we [split up this file](https://github.com/WebAssembly/spec/issues/596), this test should probably be split off to let the others keep running on systems that don't support BigInt.

document/js-api/index.bs
littledan: "Mathematical value" is a phrase used in the ECMAScript spec. It's discussed [in this section](https://tc39.github.io/ecma262/#sec-mathematical-operations) but not really linkable. I'd be fine with deleting the word "mathematical" here if it's more confusing than helpful.

xtuc: Ok, it's fine for me then. Let's maybe wait for another thought on this.

jfbastien:
> I don't have a strong opinion about when it lands in the standard, but it really wouldn't be unusual at all for the Wasm standard to land while BigInt is at Stage 3 (given all the standards which make normative references to "Editor's Draft"s, this would be nothing--Stage 3 is something like "Candidate Recommendation").

What do you mean by "wouldn't be unusual"? It would be helpful to have points of reference on how we approach this. I'd like to have it as a topic for an upcoming CG meeting.

littledan:
@jfbastien The HTML spec already [references](https://html.spec.whatwg.org/#refsJSBIGINT) BigInt, for example.

xtuc:
I think tthat this PR is obsolete, the specification and tests now lives under the proposal repo. Could we close this?

xtuc:
@littledan could we close this PR in favor of the proposal repo?

littledan:
We'll have to reopen the PR if the proposal ever reaches Stage 5; what is the problem with leaving it open?

document/js-api/index.bs
```diff
@@ -348,8 +351,11 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
             1. Let |externfunc| be the [=external value=] [=external value|func=] |funcaddr|.
             1. [=list/Append=] |externfunc| to |imports|.
         1. If |externtype| is of the form [=global=] <var ignore>mut</var> |valtype|,
-            1. If [=Type=](|v|) is Number,
-                1. If |valtype| is [=i64=], throw a {{LinkError}} exception.
+            1. If [=Type=](|v|) is Number or BigInt,
+                1. If |valtype| is [=i64=] and [=Type=](|v|) is Number,
```
taralx, line 355: This seem unnecessarily restrictive. Why not allow a Number to be passed here? The global gets allocated const regardless of the import's declared mutability.

document/js-api/index.bs
taralx: I think it's fine, but it should probably be "... of |v|". The link is https://tc39.es/ecma262/#mathematical-value, if you want to linkify it.

document/js-api/index.bs
Ms2ger: This has been discussed in https://github.com/WebAssembly/JS-BigInt-integration/issues/12; allowing Numbers to be passed is likely to lead to rounding errors. We can revisit if it turns out there is an actual problem in practice.

document/js-api/index.bs
Ms2ger: I don't think that's right: ES explicitly says that in the phrase "the mathematical value of *x*", *x* is a Number (i.e., a  IEEE double) - which it definitely can't be at this point, because then we've lost precision.

Note that the numbers coming from wasm are already treated as "mathematical values" below, as they're passed to "the Number value for *x*", where *x* is expected to be a MV. We might want to formalize this a little more, but I'm not quite sure how to best do that.

Unless you feel strongly, I'd like to defer the linkifying until after this is merged.

xtuc (approved):

binji (approved):
The proposal is officially phase 5 now, so the PR can be merged: see https://lists.w3.org/Archives/Public/public-webassembly/2020Jun/0000.html

ppKrauss:
Hi,  this new API will offer some way [to obtain the amount of bits of a BigInt](https://stackoverflow.com/q/54758130/287948)? 

Perhaps something as *integerLogarithmBase2(BigInt)*? (not a string conversion to count digits of the bits-string!).

gubonlucid-com (approved):

PR #710, opened 2018-02-21
Fix spec for fne (returns 1 for NaN)

hammacher:
fne is the invert of feq. Hence, it returns 1 if either operand is NaN.
This matches the spec tests and the behaviour of the reference interpreter.


gahaas:
I agree with Clemens that this seems to be a bug in the spec text. In my understanding IEEE 754 states in Table 4 that NaN != x is always true. Also, this is what the spec tests say in https://github.com/WebAssembly/spec/blob/master/test/core/f32_cmp.wast#L445 and following, and it matches the behavior of JavaScript and C++.

gahaas:
LGTM

PR #711, opened 2018-02-21
[spec] Normative: Support [Serializable] for WebAssembly.Module

littledan:
This patchset adds support to WebAssembly.Module to be used in HTML serialization, including from `postMessage` and IndexedDB. The change is based on @flagxor 's [earlier PR](https://github.com/WebAssembly/design/pull/1074), with the following changes:
- Rather than actually including the Module in the serialization, this PR includes only the binary data, as the Module cannot be referenced across agents (which each have their own separate store).
- The same-agent-cluster limitation is removed, as it is nonsensical in the context of *forStorage* serialization, for example IndexedDB, which is permitted by this specification.
- The Memory aspects are omitted in this PR; those will be pursued in a separate PR in the threads repository.
- The bytes of the module are explicit sub-serialized/deserialized

### DIFF
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -147,13 +147,7 @@ The [=serialization steps=], given |value|, |serialized| and |forStorage|, are:
 
     1. Set |serialized|.\[[Bytes]] to |value|.\[[Bytes]].
 
-    1. Set |serialized|.\[[AgentCluster]] to the [=current Realm=]'s corresponding [=agent cluster=].
-
 The [=deserialization steps=], given |serialized| and |value|, are:
-
-    1. If |targetRealm|'s corresponding [=agent cluster=] is not
-        |serialized|.\[[AgentCluster]], then throw a
-        "<a exception>DataCloneError</a>" {{DOMException}}.
 
     1. Let |bytes| be |serialized|.\[[Bytes]].
 
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -137,7 +137,7 @@ Note: This algorithm accepts a {{Response}} object, or a
 
 <h2 id="serialization">Serialization</h2>
 
-On the Web, {{Module}} has the <code>[<a extended-attribute>Serializable</a>]</code> extended attribute.
+Web user agents must augment the {{Module}} interface with the <code>[<a extended-attribute>Serializable</a>]</code> extended attribute.
 
 The [=serialization steps=], given |value|, |serialized|, and |forStorage|, are:
 
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -32,6 +32,12 @@ Prepare For TR: true
     "title": "WebAssembly JS Integration Specification",
     "publisher": "W3C WebAssembly Community Group",
     "status": "Draft"
+  },
+  "SECURECONTEXTS": {
+    "href": "https://w3c.github.io/webappsec-secure-contexts/",
+    "title": "Secure Contexts",
+    "publisher": "WebAppSec WG",
+    "status": "Candidate Recommendation"
   }
 }
 </pre>
@@ -42,6 +48,9 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
     type: interface
         text: ArrayBuffer; url: sec-arraybuffer-objects
+    type: dfn
+        text: agent cluster; url: sec-agent-clusters
+        text: current Realm; url: current-realm
 urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: dfn
     text: function index; url: syntax/modules.html#syntax-funcidx
     text: name section; url: appendix/custom.html?highlight=name%20section#binary-namesec
@@ -61,6 +70,7 @@ urlPrefix: https://webassembly.github.io/spec/js-api/; spec: WASMJS
         text: Exported Function; url: #exported-function
 url:https://html.spec.whatwg.org/#cors-same-origin;text:CORS-same-origin;type:dfn;spec:HTML
 url:https://fetch.spec.whatwg.org/#concept-body-consume-body;text:consume body;type:dfn;spec:FETCH
+url:https://w3c.github.io/webappsec-secure-contexts/#environment-settings-object-contextually-secure; text:contextually secure; type: dfn; spec: SECURECONTEXTS
 </pre>
 
 <pre class='link-defaults'>
@@ -124,6 +134,42 @@ Note: This algorithm accepts a {{Response}} object, or a
          1. [=Reject=] |returnValue| with |reason|.
      1. Return |returnValue|.
 </div>
+
+<h2 id="serialization">Serialization</h2>
+
+On the Web, {{Module}} has the <code>[<a extended-attribute>Serializable</a>]</code> extended attribute.
+
+The [=serialization steps=], given |value|, |serialized| and |forStorage|, are:
+
+    1. If |forStorage| is true and the [=relevant settings object=]
+        of |value| is not [=contextually secure=], throw a
+        "<a exception>DataCloneError</a>" {{DOMException}}
+
+    1. Set |serialized|.\[[Bytes]] to |value|.\[[Bytes]].
+
+    1. Set |serialized|.\[[AgentCluster]] to the [=current Realm=]'s corresponding [=agent cluster=].
+
+The [=deserialization steps=], given |serialized| and |value|, are:
+
+    1. If |targetRealm|'s corresponding [=agent cluster=] is not
+        |serialized|.\[[AgentCluster]], then throw a
+        "<a exception>DataCloneError</a>" {{DOMException}}.
+
+    1. Let |bytes| be |serialized|.\[[Bytes]].
+
+    1. Set |value|.\[[Bytes]] to |bytes|.
+
+    1. [=Compile a WebAssembly module=] from |bytes| and set |value|.\[[Module]] to the result.
+
+The semantics of a structured serialization is as-if the binary source, from which the
+{{Module}} was compiled, is serialized, then deserialized, and recompiled into the target realm.
+Engines should attempt to share/reuse internal compiled code when performing
+a structured serialization, although in corner cases like CPU upgrade or browser
+update, this might not be possible and full recompilation may be necessary.
+
+Given the above engine optimizations, structured serialization provides developers
+explicit control over both compiled-code caching and cross-window/worker code
+sharing.
 
 <h2 id="conventions">Developer-Facing Display Conventions</h2>
 
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -147,11 +147,19 @@ The [=serialization steps=], given |value|, |serialized|, and |forStorage|, are:
 
     1. Set |serialized|.\[[Bytes]] to the [=sub-serialization=] of |value|.\[[Bytes]].
 
+    1. If |forStorage| is *false*, set |serialized|.\[[AgentCluster]] to the
+       [=current Realm=]'s corresponding [=agent cluster=].
+
 The [=deserialization steps=], given |serialized| and |value|, are:
 
     1. Let |bytes| be the [=sub-deserialization=] of |serialized|.\[[Bytes]].
 
     1. Set |value|.\[[Bytes]] to |bytes|.
+
+    1. If |serialized| has an \[[AgentCluster]] field, and if |targetRealm|'s
+        corresponding [=agent cluster=] is not
+        |serialized|.\[[AgentCluster]], then throw a
+        "<a exception>DataCloneError</a>" {{DOMException}}.
 
     1. [=Compile a WebAssembly module=] from |bytes| and set |value|.\[[Module]] to the result.
 
@@ -164,6 +172,10 @@ Note: The semantics of a structured serialization is as-if the binary source, fr
 Given the above engine optimizations, structured serialization provides developers
 explicit control over both compiled-code caching and cross-window/worker code
 sharing.
+
+Note: The web platform does not permit cross-origin sharing of serialized Modules: in the
+case of |forStorage| being true, IndexedDB separates storage per origin; in the case of
+|forStorage| being false, agent clusters in HTML do not span multiple origins.
 
 <h2 id="conventions">Developer-Facing Display Conventions</h2>
 
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -145,11 +145,11 @@ The [=serialization steps=], given |value|, |serialized| and |forStorage|, are:
         of |value| is not [=contextually secure=], throw a
         "<a exception>DataCloneError</a>" {{DOMException}}
 
-    1. Set |serialized|.\[[Bytes]] to |value|.\[[Bytes]].
+    1. Set |serialized|.\[[Bytes]] to the [=sub-serialization=] of |value|.\[[Bytes]].
 
 The [=deserialization steps=], given |serialized| and |value|, are:
 
-    1. Let |bytes| be |serialized|.\[[Bytes]].
+    1. Let |bytes| be the [=sub-deserialization=] of |serialized|.\[[Bytes]].
 
     1. Set |value|.\[[Bytes]] to |bytes|.
 
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -141,13 +141,12 @@ Web user agents must augment the {{Module}} interface with the <code>[<a extende
 
 The [=serialization steps=], given |value|, |serialized|, and |forStorage|, are:
 
-    1. If |forStorage| is true and the [=relevant settings object=]
-        of |value| is not [=contextually secure=], throw a
+    1. If |forStorage| is true, throw a
         "<a exception>DataCloneError</a>" {{DOMException}}
 
     1. Set |serialized|.\[[Bytes]] to the [=sub-serialization=] of |value|.\[[Bytes]].
 
-    1. If |forStorage| is *false*, set |serialized|.\[[AgentCluster]] to the
+    1. Set |serialized|.\[[AgentCluster]] to the
        [=current Realm=]'s corresponding [=agent cluster=].
 
 The [=deserialization steps=], given |serialized| and |value|, are:
@@ -156,8 +155,7 @@ The [=deserialization steps=], given |serialized| and |value|, are:
 
     1. Set |value|.\[[Bytes]] to |bytes|.
 
-    1. If |serialized| has an \[[AgentCluster]] field, and if |targetRealm|'s
-        corresponding [=agent cluster=] is not
+    1. If |targetRealm|'s corresponding [=agent cluster=] is not
         |serialized|.\[[AgentCluster]], then throw a
         "<a exception>DataCloneError</a>" {{DOMException}}.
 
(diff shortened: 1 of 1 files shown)


document/web-api/index.bs
```diff
@@ -2,7 +2,7 @@
 Title: WebAssembly Web API
 Shortname: wasm-web-api
 Group: wasm
-Status: FPWD
+Status: w3c/FPWD
```
littledan: This change was needed for me to be able to build locally

document/web-api/index.bs
```diff
@@ -125,6 +135,36 @@ Note: This algorithm accepts a {{Response}} object, or a
      1. Return |returnValue|.
 </div>
 
+<h2 id="serialization">Serialization</h2>
+
+On the Web, {{Module}} has the <code>[<a extended-attribute>Serializable</a>]</code> extended attribute.
```
annevk: Can't this be in its IDL? This seems somewhat unclear.

document/web-api/index.bs
```diff
@@ -125,6 +135,36 @@ Note: This algorithm accepts a {{Response}} object, or a
...
+
+The [=serialization steps=], given |value|, |serialized| and |forStorage|, are:
```
annevk: Oxford comma?

document/web-api/index.bs
```diff
@@ -125,6 +135,36 @@ Note: This algorithm accepts a {{Response}} object, or a
...
+
+The semantics of a structured serialization is as-if the binary source, from which the
+{{Module}} was compiled, is serialized, then deserialized, and recompiled into the target realm.
+Engines should attempt to share/reuse internal compiled code when performing
+a structured serialization, although in corner cases like CPU upgrade or browser
+update, this might not be possible and full recompilation may be necessary.
+
+Given the above engine optimizations, structured serialization provides developers
+explicit control over both compiled-code caching and cross-window/worker code
+sharing.
```
annevk: This seems like a note?

document/web-api/index.bs
```diff
@@ -125,6 +135,36 @@ Note: This algorithm accepts a {{Response}} object, or a
...
+The [=deserialization steps=], given |serialized| and |value|, are:
+
+    1. Let |bytes| be the [=sub-deserialization=] of |serialized|.\[[Bytes]].
+
+    1. Set |value|.\[[Bytes]] to |bytes|.
+
+    1. [=Compile a WebAssembly module=] from |bytes| and set |value|.\[[Module]] to the result.
+
+The semantics of a structured serialization is as-if the binary source, from which the
+{{Module}} was compiled, is serialized, then deserialized, and recompiled into the target realm.
```
annevk: This seems like a note as well.

document/web-api/index.bs
littledan: Well, the JS API specification is trying to avoid referencing anything which is Web-specific. [Serializable] is defined in HTML. I'm not sure how to put this in its IDL without violating that layering.

annevk: If these objects are not defined in IDL then you need to patch HTML directly as you did with BigNum.

annevk: (I said as much on the original PR.)

littledan: They are defined in IDL.

littledan: Here's the IDL: https://webassembly.github.io/spec/js-api/index.html#module

annevk: It also has Exposed=(Window,Worker,Worklet) so I think you're failing somewhat?

littledan: @annevk Well, there's a plan for that: https://github.com/heycam/webidl/issues/468 . This layering is still a goal.

annevk: Okay. I think you need to make this something like "Web user agents must augment the Module interface with the [Serializable] extended attribute." so it's an actual requirement and not a statement of fact that isn't based on any established facts...

jfbastien:
This is a spec addition, right? I think you need to follow the same [process](https://github.com/WebAssembly/meetings/blob/master/process/phases.md) to get this addition into WebAssembly. Should be very straightfoward, but we can't commit it until the working draft ships (unless we want to re-spin a new public working draft, and all that entails).

domenic:
Hmm, I sure hope this group has the ability to publish new editor's draft without going through the W3C working draft process stuff...

domenic:
This PR is very confusing BTW because it says "Memory" several times (e.g. in the PR title) when I believe the actual spec text is for "Module".

jfbastien:
> Hmm, I sure hope this group has the ability to publish new editor's draft without going through the W3C working draft process stuff...

Maybe I misunderstand, but IIUC this PR requires me to change code in WebKit. I agree with the change, but it's not editorial. It's a new capacity, and as such should go through the CG as documented in our [process](https://github.com/WebAssembly/meetings/blob/master/process/phases.md). Doesn't matter if it's small and obvious, we can't just commit non-editorial changes to the spec. All work would then iterate in a separate repo, which we'd merge into the spec when it's ready to be merged.

Separately, even if we go through the process, I'm worried that merging this change into the tip-of-tree of this repo *does* require going through the W3C working draft process. A way to avoid this is to work purely in a branch for each working draft. This change then wouldn't affect the draft.

flagxor:
This was content that was present in the design repo material before it got converted to our current spec. It was lost in transcription as the old verbiage was very informal (and Dan had the impression it got moved to the HTML spec).
https://github.com/WebAssembly/design/blob/master/JS.md#structured-clone-of-a-webassemblymodule
The PR was my attempt to transform it into something more formal, but I paused that once Dan took over drafting our JS spec.

We could certainly churn through the proposal process, but I would argue that's unnecessarily bureaucratic.
This certainly exceeds the "editorial" bar and is a "substantive" change as it does have bearing on implementations, however, I don't think that requires the full process as this goes back to the seed material for the spec. I.e. this would be similar to if Andreas had managed to forget an opcode in the core spec. It is something we need to track and note when publishing the next working draft, but because of its origin, it seems like overkill to run the whole proposal process.

document/web-api/index.bs
flagxor: If you update your bikeshed version that should fix it (it will then know wasm is a w3c project).
The reason I changed these is that bikeshed rejects some of the normal stamping with the full string. I still owe Tab filing an issue on that.

jfbastien:
Brad's explanation makes sense. Agreed.

littledan:
@domenic Oops re: Memory. Fixed up the cases I saw.

@jfbastien I think there are two discussions here and we're talking past each other. One question is whether this is a new feature. I don't think anyone is saying no to this. This definitely isn't an editorial change, it's a normative change and new feature (which might deserve to cut to the head of the line for @flagxor 's historical reasons). I'd be fine asking the Wasm CG for a repo here, but sounds you're not asking me to. 

What @domenic is suggesting is that the WebAssembly CG may want to maintain two documents in parallel: A FWPD (what you've already sent out, leaving it frozen), and an "Editor's draft" (W3C speak for a document that you can just edit whenever you want). Basically, the standard that you put through the IP opt-out period and the other formal processes that follow it would be a fork of the editor's draft, which we'd occasionally backport editorial patches to or whatever else is urgent.

A nice thing about maintaining an Editor's Draft is that people working on new features don't have to block on W3C process to get their work done. At the same time, people who are focusing on W3C process have a stamp-ready, feature-filled document ready for them once the cycle opens up again--there's no need to "open the floodgates" and merge a bunch of things at a particular time.

document/web-api/index.bs
littledan: Thanks for the wording; I've switched to use it.

jfbastien:
> @jfbastien I think there are two discussions here and we're talking past each other. One question is whether this is a new feature. I don't think anyone is saying no to this. This definitely isn't an editorial change, it's a normative change and new feature (which might deserve to cut to the head of the line for @flagxor 's historical reasons). I'd be fine asking the Wasm CG for a repo here, but sounds you're not asking me to.

Correct, if this is just putting in something that was agree on, and that was dropped, then I don't think you need to go through the process. I encourage you to provide this type of context next time, so non-editorial changes are easy to categorize.

> What @domenic is suggesting is that the WebAssembly CG may want to maintain two documents in parallel: A FWPD (what you've already sent out, leaving it frozen), and an "Editor's draft" (W3C speak for a document that you can just edit whenever you want). Basically, the standard that you put through the IP opt-out period and the other formal processes that follow it would be a fork of the editor's draft, which we'd occasionally backport editorial patches to or whatever else is urgent.
>
> A nice thing about maintaining an Editor's Draft is that people working on new features don't have to block on W3C process to get their work done. At the same time, people who are focusing on W3C process have a stamp-ready, feature-filled document ready for them once the cycle opens up again--there's no need to "open the floodgates" and merge a bunch of things at a particular time.

Agreed, it's simply more maintenance to keep two branches going.

document/web-api/index.bs
littledan: How should I update it? I tried `bikeshed update` and it still gave this error.

domenic: I think you need to actually update your Bikeshed source code, so e.g. go to the director you checked out bikeshed and run `git pull`.

littledan:
It seems like web-platform-tests already has good tests for module serialization, which verify that
- `postMessage` works with [Workers](https://github.com/w3c/web-platform-tests/blob/master/wasm/wasm_serialization_tests.html) and [iframes](https://github.com/w3c/web-platform-tests/blob/master/wasm/wasm_local_iframe_test.html).
- [Modules can be serialized to IndexedDB](https://github.com/w3c/web-platform-tests/blob/master/wasm/wasm_indexeddb_test.https.html)
- [postMessage of a Module from a ServiceWorker doesn't work](https://github.com/w3c/web-platform-tests/blob/master/wasm/wasm_service_worker_test.https.html). This was previously specified by checking the agent cluster, but that check isn't in the current PR.

Two things from here:
- Probably there should be a test that checks that IndexedDB is not allowed to serialize Modules from an insecure context.
- Should we re-add the same agent cluster check? Does anyone remember why it was there in the first place? If we do add it back, I think we still want it disabled in the *forStorage* case (since it doesn't seem to make sense there). Otherwise, I can update the test to check that postMessage does actually work from a ServiceWorker.

flagxor:
So my recollection was that there was disagreement around requiring a secure context for IDB serialization (from JSC + Firefox)? Chrome has taken the position that persisting code from a potentially insecure context is problematic.

I think the agent cluster check just fell out of the assumption that code sharing was most essential between threads. I'd be happy dropping it if others don't care?

littledan:
Given the past disagreement on this security question about storing code from insecure contexts, how should we proceed? Should the specification permit the various positions taken by different browsers, or should we come to a common, single semantic?

jfbastien:
The security question is specifically about caching compiled code in IDB from an insecure context, not serializing, correct?

What do other web caching things to?

domenic:
We must become interoperable on this web-observable behavior.

Other modern caching things (e.g. the service worker cache API) are not accessible from insecure contexts.

littledan:
@jfbastien "Serializing" is the name in the HTML specification that describes how things are stored in IndexedDB as well as postMessage. There is a forStorage mode switch which specifications can use to differentiate those two cases.

domenic:
Right, it's about injecting content into the site. If I understand correctly the scenario, under insecure HTTP, an attacker can control everything, including the IDB key.

jfbastien:
As we discussed in the CG call today, I'd like our decision to mention HTTP / HTTPS to be a principled one, not just "hey we could do it for IDB". It sounds like at least Chrome wouldn't allow IDB serialization under HTTP. I'd rather spec any secure context restrictions separately, following a WebAssembly overall guidance. Maybe this can be an offshoot out the [secure context](https://github.com/WebAssembly/meetings/issues/153) discussion?

littledan:
We heard in the call that Chrome is unlikely to ship the ability to store Modules  from insecure contexts in  IndexedDB. If it's possible, I hope we can come to agreement among all the browsers that what we specify will eventually be shipped soon.

To ensure compatibility, I'd prefer to aim to be only adding features over time, and not shipping additional restrictions on previously shipped features unless it is to correct a previous mistake. (If there isn't compatibility risk from a future restriction, it's because the feature isn't being used; in that case, why bother specifying and shipping it?)

I think additional restrictions are a decent conservative default given disagreement. In this case, I could see logic to the restriction--persistent storage of attacker code is worse than ephemerally running attacker code on a particular page load.

jfbastien:
> We heard in the call that Chrome is unlikely to ship the ability to store Modules from insecure contexts in IndexedDB. If it's possible, I hope we can soon come to agreement among all the browsers that what we specify here will eventually be shipped.
>
> To ensure compatibility, I'd prefer to aim to be only adding features over time, and not shipping additional restrictions on previously shipped features unless it is to correct a previous mistake. (If there isn't compatibility risk from a future restriction, it's because the feature isn't being used; in that case, why bother specifying and shipping it?)

I don't really see Chrome not shipping this for HTTP as a big issue since all sites have to handle IDB storage failure already, and not all browsers currently support IDB module storage. The incompatibility is indistinguishable from the current state of the web, as well as from random failure.

If you want to aim for specifying HTTPS for module serialization, that's also fine with me. As I've outlined I think we should do so in a principled manner, not ad-hoc for this proposal. We can block serialization of modules on figuring out a precise policy for how WebAssembly should act under an insecure context. Should we discuss this with webappsec alongside CSP, or as a separate thing?

littledan:
> I don't really see Chrome not shipping this for HTTP as a big issue since all sites have to handle IDB storage failure already, and not all browsers currently support IDB module storage. The incompatibility is indistinguishable from the current state of the web, as well as from random failure.

If we're OK with unpredictable performance, I don't understand the motivation for this feature. Browsers could just transparently cache modules based on a hash of the bitcode. I thought the idea of serializing modules in IndexedDB was that you'd be able to fairly reliably control when they are cached and read out.

There's a bit of difference between a permanent state of differences between browsers and other cases where this feature won't be available: when comparing to the current state of the web, this is hopefully a one-time upgrade, and then all browsers should support IDB persistence of modules reliably. For eviction from storage, this can be controlled by the persistent storage API. For random failures, these should be pretty rare and analogous to a first page load; it should be pretty rare for a second load immediately after that to face the same issue.

jfbastien:
@littledan I'm making a very narrow argument from the current state of which browser ships what. From that current state of things, there's little difference in a world where HTTP and HTTPS don't act the same, whether normative or not, because already different browsers don't act the same and already there's random eviction.

Longer term (say, later this year) we'll figure out how to handle HTTPS in this spec. I just don't think an ad-hoc solution for this specific feature is right for the long term, because I'm not sure we'll get the big picture right (we might, I'm simply not confident that we will). Otherwise we'd already have that big picture in the spec.

littledan:
We discussed structured clone of modules in the April 4th WebAssembly CG call. To summarize the discussion:
- The same agent cluster restriction for Module serialization through postMessage was justified for the following reasons:
  - This restriction means that Modules are not sent cross-origin, which may be helpful as a security invariant
  - There weren't clear cross-agent-cluster use cases.
  - In some browsers, restricting to a single agent cluster makes for a simpler implementation
- We did not address or come to a conclusion on the secure context restriction for storage of WebAssembly Modules. This will have to be discussed in the future.

I'll follow up with a patch which maintains the secure context restriction and follows the other restrictions listed above.

annevk:
> This restriction means that Modules are not sent cross-origin, which may be helpful as a security invariant

Not cross-site, but they can go cross-origin.

littledan:
@annevk Are you talking about the postMessage case? Can you give an example/reference for more details about the cross-site/cross-origin differences and how it relates to agent clusters? We were having trouble in the last CG call.

annevk:
@littledan if example.com embeds site.example.com through a frame they'll be in the same similar-origin window agent. That's because if they both set `document.domain` to `example.com` they can access each other directly.

Basically, an agent cluster by definition can consists of multiple origins, but those origins always have the same eTLD+1 (registrable domain).

littledan:
Thanks for the explanation. Seems like we should discuss in the CG whether we want a same or similar origin restriction here.

annevk:
@littledan that would require storing origin data on the `Module` object, which I would strongly recommend against (and have already done elsewhere). You want a capability-based model for objects, not tagging.

littledan:
@annevk What do you think about this current patch, which stores the agent cluster if it's not forStorage (similar to SharedArrayBuffer, but for different reasons).

annevk:
@littledan I think that's the correct approach. Though writing that down I do now recall that folks raised a concern in private with me about that approach combined with `MessageChannel` not working well in implementations for `SharedArrayBuffer`. I've asked them last year to report an issue for this, but that hasn't yet happened.

littledan:
Apologies for my delay. Chrome is [considering shipping](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/zxsHPf8Zmv8) same-agent-cluster postMessage but not IndexedDB serialization of WebAssembly modules. I've updated this PR to match that proposal. See https://github.com/WebAssembly/spec/pull/711/commits/314099cd2db1e8970f23fd7647a5d9fb2cdc65b6 for more information.

document/web-api/index.bs
```diff
@@ -125,6 +135,45 @@ Note: This algorithm accepts a {{Response}} object, or a
...
 </div>
 
+<h2 id="serialization">Serialization</h2>
+
+Web user agents must augment the {{Module}} interface with the <code>[<a extended-attribute>Serializable</a>]</code> extended attribute.
+
+The [=serialization steps=], given |value|, |serialized|, and |forStorage|, are:
+
+    1. If |forStorage| is true, throw a
+        "<a exception>DataCloneError</a>" {{DOMException}}
```
domenic: Nit: `"{{DataCloneError}}"` is more conventional.

document/web-api/index.bs
```diff
@@ -125,6 +135,45 @@ Note: This algorithm accepts a {{Response}} object, or a
...
+update, this might not be possible and full recompilation may be necessary.
+
+Note: The semantics of a structured serialization is as-if the binary source, from which the
+{{Module}} was compiled, is serialized, then deserialized, and recompiled into the target realm.
+Given the above engine optimizations, structured serialization provides developers
+explicit control over both compiled-code caching and cross-window/worker code
+sharing.
+
+Note: The web platform does not permit cross-origin sharing of serialized Modules:
+Agent clusters in HTML do not span multiple origins.
```
annevk: This is incorrect. It does not allow sharing when two origins's hosts are not https://url.spec.whatwg.org/#host-same-site or do not have the same scheme. Otherwise sharing is possible.

annevk: (This is also why we call it the similar-origin window agent, not just window agent.)

littledan: I've just removed the note.

littledan:
When do we want to consider landing this PR? Should it be in the v1 spec, or are we OK with waiting for v2? It's a significant new feature, so I'm not sure if it might reset some sort of clock. @lukewagner suggested that waiting for v2 might be fine, as most of the use cases for postMessage are motivated by threads, which are only coming in v2.

@Ms2ger Would you be interested in looking into the tests which @annevk [suggested](https://github.com/WebAssembly/spec/pull/711#pullrequestreview-128755936)?

binji:
> Should it be in the v1 spec, or are we OK with waiting for v2?

We discussed having this for v1 already, and only decided recently not to move forward with serialization for IndexedDB part. I think it makes sense to keep for v1.

> It's a significant new feature, so I'm not sure if it might reset some sort of clock.

It will, yes, but we're past the point where that matters. We need to put out a new version regardless which will necessarily reset the clock.

littledan:
OK, let's discuss this at the next CG meeting tomorrow and merge it if it seems OK

binji:
We discussed this in the June 26th CG meeting and decided to include this PR in v1 of the spec.

littledan:
@Ms2ger Could you make sure we have good tests for this patch in web-platform-tests? In particular, we should have tests that serializing in IndexedDB throws, and that serialization in postMessage and MessageChannel works.

Ms2ger:
Looks like https://github.com/web-platform-tests/wpt/pull/12488 covers this neatly.

Ms2ger:
Looks like https://github.com/web-platform-tests/wpt/pull/12488 covers this neatly.

PR #712, opened 2018-02-21
[spec] Layering: Improve way of throwing on detach

littledan:
Detaching an ArrayBuffer which comes from a WebAssembly.Memory
throws an error. Previously, this was simply asserted in a comment;
this patch hooks into a new ECMAScript layering device to make
things more clear.

Depends on https://github.com/tc39/ecma262/pull/1112

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -555,11 +555,10 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
         1. Return |map|[|memaddr|].
     1. Let |block| be a Data Block which is [=identified with=] the underlying memory of |memaddr|
     1. Let |buffer| be a new {{ArrayBuffer}} whose \[[ArrayBufferData]] is |block| and \[[ArrayBufferByteLength]] is set to the length of |block|.
+    1. Set |buffer|.\[[ArrayBufferDetachKey]] to "WebAssembly.Memory".
     1. Let |memory| be a new {{Memory}} instance with \[[Memory]] set to |memaddr| and \[[BufferObject]] set to |buffer|.
     1. [=map/Set=] |map|[|memaddr|] to |memory|.
     1. Return |memory|.
-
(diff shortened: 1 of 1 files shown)


littledan:
I'll wait to land this change until the JS side lands.

littledan:
The JS side received consensus in TC39, so I think this patch should be good to land (even if there is a slight delay on the JS patch landing).

littledan:
The related JS patch landed https://github.com/tc39/ecma262/pull/1112

ajklein:
I was looking at some odd behavior today that led me back to this change. There's a difference in behavior between Firefox and Chrome for the following code:

```js
postMessage('foo', '*', [new WebAssembly.Memory({initial: 4}).buffer]);
```

Firefox throws a `TypeError`, while Chrome happily "transfers" the buffer. Except we don't transfer (because the buffer is not detachable); instead we silently copy it.

I'm curious how the Firefox behavior is specced: I would have expected there to be an HTML spec change for this as well. But I don't see anything in the [transfer algorithm](https://html.spec.whatwg.org/multipage/structured-data.html#structuredserializewithtransfer) that checks for detachability of an ArrayBuffer.

@littledan @domenic Can you help enlighten me?

cc @dtig

ajklein:
@domenic If I'm reading it right, that section disallows transferring an already-detached buffer. But in the case above, the buffer isn't detached; it's simply "undetachable" (because of its association with the WebAssembly.Memory). It's this "undetachability check" which I'm searching for, and I still don't see one in the transfer algorithm.

littledan:
Cc @ms2ger

Ms2ger:
@ajklein you appear to be correct; could you file an issue on the HTML spec and cc me?

domenic:
FWIW I wasn't really aware of "undetachable" ArrayBuffers before yesterday. I vaguely remember some TC39 work around WebAssembly and ArrayBuffers, I guess this is that, and it had an HTML impact I wasn't aware of :). Thanks @Ms2ger and @littledan for being on top of things.

ajklein:
Filed https://github.com/whatwg/html/issues/4601, thanks everyone

Issue #737, opened 2018-03-02
Parse error w/ flat call_indirect


binji:
The following .wat file doesn't parse:
```
(table 1 anyfunc)
(func
  i32.const 0
  call_indirect)
```

It seems that this rule is the problem, requiring a following `instr`:
```
call_instr_results :
  | LPAR RESULT value_type_list RPAR call_instr_results
    { fun c -> let ts, es = $5 c in $3 @ ts, es }
  | instr
    { fun c -> [], $1 c }
```

Changing it to `/* empty */ { fun c -> [], [] }` introduces shift/reduce conflicts.

rossberg:
Aw, frack. If only we had never introduced the flat syntax. Or yacc was LR(2). See #738 for the fix.

binji:
This was fixed in https://github.com/WebAssembly/spec/pull/738.

PR #738, opened 2018-03-02
[interpreter] Fix flat call_indirect

rossberg:
Addresses #737.

### DIFF
--- a/test/core/call_indirect.wast
+++ b/test/core/call_indirect.wast
@@ -80,38 +80,6 @@
     ))
   )
 
-  (func
-    (block i32.const 0 call_indirect)
-    (loop i32.const 0 call_indirect)
-    (if (i32.const 0) (then i32.const 0 call_indirect))
-    (if (i32.const 0)
-      (then i32.const 0 call_indirect)
-      (else i32.const 0 call_indirect)
-    )
-    (block i32.const 0 i32.const 0 call_indirect (param i32))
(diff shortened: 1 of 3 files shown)


test/core/call_indirect.wast
```diff
@@ -80,6 +80,38 @@
     ))
   )
 
+  (func
```
binji: Are we still trying to keep stacky/flat stuff in stack.wast (e.g. for Binaryen)? If not, I'd rather keep it here.

test/core/call_indirect.wast
```diff
@@ -80,6 +80,38 @@
...
+    (block i32.const 0 call_indirect)
+    (loop i32.const 0 call_indirect)
+    (if (i32.const 0) (then i32.const 0 call_indirect))
+    (if (i32.const 0)
+      (then i32.const 0 call_indirect)
+      (else i32.const 0 call_indirect)
+    )
+    (block i32.const 0 i32.const 0 call_indirect (param i32))
```
binji: Might be good to test some with results too.

test/core/call_indirect.wast
rossberg: Good point, moved to stack.wast

test/core/call_indirect.wast
rossberg: Added more tests, including (type $t) and multiple fields.

PR #740, opened 2018-03-03
[spec] Normative: Detach previous JS ArrayBuffer from grow_memory

### LINKED ISSUE #646: JS API Spec for grow memory and detaching array buffers
The JS API spec definitively states that growing a memory through the `WebAssembly.Memory.grow()` operation detaches the array buffer and allocates a new one. However, the core spec has no notion of the JS API, and thus the specification for `grow_memory` does not mention detachment of any array buffers as a result of executing a `grow_memory` bytecode.

I think the reasonable interpretation is that the `grow_memory` bytecode does indeed detach a JS-level array buffer for a memory, and I started to update our implementation to do this. However this is underspecified in a way since this is a case of additional behavior being added to a core bytecode by an embedding spec.

1.) is the interpretation reasonable (i.e. that the `grow_memory` should have a JS API side-effect)?
2.) how should we handle this at the specification level?

littledan:
Closes #646

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -155,6 +155,10 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     text: global type; url: syntax/types.html#syntax-globaltype
     text: address; url: exec/runtime.html#addresses
     text: signed_32; url: exec/numerics.html#aux-signed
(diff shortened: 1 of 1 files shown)


littledan:
Note that this is missing tests. Eventually we may want to get to a point where we don't land spec changes without tests, but given that we're pretty far behind here, I think I'll just land this change to describe shipping implementations more accurately and follow up with a test later.

PR #745, opened 2018-03-07
[spec] Normative: Permit work in parallel during instantiation

### LINKED ISSUE #741: [js-api] Allow optional asynchronous steps after instantiation?
The previous JS.md includes this phrase in a [the definition of instantiation](https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstantiate):

> After the instantiation task runs and before any subsequent steps are taken, other unspecified asynchronous tasks may be run.

The current JS API spec doesn't include this delay [in its definition of instantiation](https://webassembly.github.io/spec/js-api/index.html#instantiate-a-promise-of-a-module). In #webkit, @domenic and @kmiller68 discussed this change. @kmiller68 raised the concern that this is a change over what was previously agreed, and that JSC does make this kind of delay, though it'd be possible to remove it. (Anyone know where the #webkit logs are?)

My misunderstanding when writing the spec initially was that, because things were "in parallel", any  more delays would be unobservable and don't need specific spec text. However, the current text has moved to more specific "queue a task" language, and instantiation actually executes JS code just before resolving a promise.

I see three ways we could specify things now:
1. Prohibit additional asynchronous work, as in the current spec text
1. "Queue a task" to resolve the Promise
1. Permit either of 1. and 2.

Because the execution order of task queues and Promises is pretty tightly specified, there's an observable difference possible between 1. and 2. The current spec text immediately resolves the Promise, so no HTML tasks will run in between reading the properties of the import object and calling then `then` callbacks that are attached to the Promise for the instantiated module. Further, the microtask queue order is specified, so there is even more detail. By contrast, in the semantics of 2., the microtask queue would be drained and some HTML tasks may run before going back and calling the `then` callbacks for the Promise.

I think option 3 corresponds to the previously documented semantics in JS.md, but at the same time, the trend among modern JS/web specifications seems to be to very tightly specify the ordering one way or the other.

Which of these three options should we choose?

littledan:
On JSC, some compilation work takes place the first time that a
Module is instantiated. If the module is instantiated in an
asynchronous way, the work happens off the main thread. This patch
loosens the specification of WebAssembly module instantiation
to permit this specific type of delay.

Based on skimming the code (but not running tests), I believe that
the new specification matches 3/4 of the implementations I examined.
The one which might not fit is V8, which defers reading properties
of the import object to a microtask queue item at the beginning
of WebAssembly.instantiate(Module), unlike the others, which read it
synchronously.

The previous specification didn't match any implementations, as it
specified to queue an HTML task at the beginning of the
WebAssembly.instantiate(Module) method, which no implementation
actually did to my knowledge. Punting to a queue there doesn't really
serve any purpose and was simply an error in drafting the previous
specification.

Closes #741
See also https://github.com/webpack/webpack/issues/6433

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -592,6 +592,8 @@ interface Module {
     1. If |module| is [=error=], throw a {{CompileError}} exception.
     1. Set **this**.\[[Module]] to |module|.
     1. Set **this**.\[[Bytes]] to |stableBytes|.
+
+Note: Some implementations enforce a size limitation on |bytes|. Use of this API is discouraged, in favor of asynchronous APIs.
 </div>
 
 <h3 id="instances">Instances</h3>
@@ -610,6 +612,8 @@ interface Instance {
     1. [=Read the imports=] of |module| with imports |importObject|, and let |imports| be the result.
     1. [=Instantiate the core of a WebAssembly module=] |module| with |imports|, and let |instance| be the result.
     1. [=initialize an instance object|Initialize=] **this** from |module| and |instance|.
+
+Note: The use of this synchronous API is discouraged, as some implementations sometimes do long-running compilation work when instantiating.
 </div>
 
 <div algorithm>
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -462,13 +462,15 @@ The verification of WebAssembly type requirements is deferred to the
     1. Let |module| be |moduleObject|.\[[Module]].
     1. [=Read the imports=] of |module| with imports |importObject|, and let |imports| be the result.
         If this operation throws an exception, catch it, [=reject=] |promise| with the exception, and return |promise|.
-    1. [=Queue a task=] to perform the following steps:
-        1.  [=Instantiate the core of a WebAssembly module=] |module| with |imports|, and let |instance| be the result.
-            If this throws an exception, catch it, [=reject=] |promise| with the exception, and terminate these substeps.
-        1.  Let |instanceObject| be a [=/new=] {{Instance}}.
-        1.  [=initialize an instance object|Initialize=] |instanceObject| from |module| and |instance|.
-            If this throws an exception, catch it, [=reject=] |promise| with the exception, and terminate these substeps.
-        1. [=Resolve=] |promise| with |instanceObject|.
+    1. Run the following steps [=in parallel=]:
+        1. [=Queue a task=] to perform the following steps:
+            Note: Implementation-specific work may be performed here.
+            1.  [=Instantiate the core of a WebAssembly module=] |module| with |imports|, and let |instance| be the result.
+                If this throws an exception, catch it, [=reject=] |promise| with the exception, and terminate these substeps.
+            1.  Let |instanceObject| be a [=/new=] {{Instance}}.
+            1.  [=initialize an instance object|Initialize=] |instanceObject| from |module| and |instance|.
+                If this throws an exception, catch it, [=reject=] |promise| with the exception, and terminate these substeps.
+            1. [=Resolve=] |promise| with |instanceObject|.
     1. Return |promise|.
 </div>
 
@@ -487,14 +489,15 @@ The verification of WebAssembly type requirements is deferred to the
 
     1. Let |promise| be [=a new promise=].
     1. [=Upon fulfillment=] of |promiseOfModule| with value |module|:
-        1. [=synchronously instantiate a WebAssembly module|Synchronously instantiate the WebAssembly module=] |module| importing |importObject|, and let |instance| be the result.  If this throws an exception, catch it, [=reject=] |promise| with the exception, and abort these substeps.
-        1. Let |result| be the {{WebAssemblyInstantiatedSource}} value «[ "{{WebAssemblyInstantiatedSource/module}}" → |module|, "{{WebAssemblyInstantiatedSource/instance}}" → |instance| ]».
-        1. [=Resolve=] |promise| with |result|.
+        1. [=asynchronously instantiate a WebAssembly module|Instantiate the WebAssembly module=] |module| importing |importObject|, and let |innerPromise| be the result.
+        1. [=Upon fulfillment=] of |innerPromise| with value |instance|.
+            1. Let |result| be the {{WebAssemblyInstantiatedSource}} value «[ "{{WebAssemblyInstantiatedSource/module}}" → |module|, "{{WebAssemblyInstantiatedSource/instance}}" → |instance| ]».
+            1. [=Resolve=] |promise| with |result|.
+        1. [=Upon rejection=] of |innerPromise| with reason |reason|:
+            1. [=Reject=] |promise| with |reason|.
     1. [=Upon rejection=] of |promiseOfModule| with reason |reason|:
         1. [=Reject=] |promise| with |reason|.
     1. Return |promise|.
-
-    Note: It would be valid to perform certain parts of the instantiation [=in parallel=], but several parts need to happen in the event loop, including JavaScript operations to access the |importObject| and execution of the start function.
 </div>
 
 <div algorithm>
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -473,7 +473,7 @@ The verification of WebAssembly type requirements is deferred to the
 </div>
 
 <div algorithm="instantiate">
-  To <dfn>instantiate a WebAssembly module</dfn> from a {{Module}} |moduleObject| and imports |importObject|, perform the following steps:
+  To <dfn>synchronously instantiate a WebAssembly module</dfn> from a {{Module}} |moduleObject| and imports |importObject|, perform the following steps:
     1. Let |module| be |moduleObject|.\[[Module]].
     1. [=Read the imports=] of |module| with imports |importObject|, and let |imports| be the result.
     1. [=Instantiate the core of a WebAssembly module=] |module| with |imports|, and let |instance| be the result.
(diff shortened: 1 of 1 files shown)


littledan:
cc @kmiller68 @tschneidereit @domenic @lukewagner

tschneidereit:
This looks good to me as a solution to the problem at hand. However, I do wonder if it wouldn't make sense to be more strict about this: I could well imagine the kinds of timing differences permitted here to cause compatibility issues at some point.

The alternative would be to specify something matching 2/4 implementations. Assuming v8 removes the additional microtask queue item before reading import object properties, that'd go up to 3/4.

@kmiller68, could you imagine living with not being compliant for a while until you've made the necessary changes? I guess we could also just make the change in this PR, have all engines be compliant for now, and tighten things later. I'm not quite sure what that'd gain, though.

littledan:
@tschneidereit Did you see @linclark's explanation of the motivation for JSC's implementation strategy in https://github.com/webpack/webpack/issues/6433 ?

tschneidereit:
> @tschneidereit Did you see @linclark's explanation of the motivation for JSC's implementation strategy in webpack/webpack#6433 ?

I did see that, yes. Your and @kmiller68's comments[1][2] in #741 led me to believe that this is a slightly different case. If JSC won't change its behavior in the foreseeable future, then we should have that be reflected in the spec for sure.

[1] ["@kmiller68 raised the concern that this is a change over what was previously agreed, and that JSC does make this kind of delay, though it'd be possible to remove it."](https://github.com/WebAssembly/spec/issues/741#issue-302079512)
[2] ["I don't think this behavior is essential and we could probably implement something different with enough effort."](https://github.com/WebAssembly/spec/issues/741#issuecomment-370599697)

littledan:
OK, let's get more input from JSC, but this seems like part of what should go into V1 either way we decide.

littledan:
We discussed this change in the WebAssembly WG meeting on April 19th, and my understanding is that it had consensus, to continue permitting the flexibility. Should we land this patch?

binji:
@littledan I don't see the notes for this discussion (also that date seems wrong, perhaps a typo?)

If this was discussed and agreed, we should merge, though it currently has conflicts.

littledan:
The meeting date was clearly a typo. I've pushed a rebase. I can't find any meeting notes which refer to this PR. Maybe we should discuss it in the future. cc @kmiller68

littledan:
In [a later WebAssembly CG meeting](https://github.com/WebAssembly/meetings/blob/master/2018/CG-09-18.md), we discussed this issue further, with the suggestion to *always* take an event loop turn, even if it's not necessary, for consistency across platforms. In effect, this means each module instantiation has an extra `setImmediate()`. Presumably this would carry forward to ESM integration as well.

I'm a little worried about the performance cost here. Is anyone else? How serious would the compatibility issue be of permitting this timing difference across implementations?

@kmiller68 Could you clarify, is there any chance you could move the slow operations from the instantiate phase to the compile phase, as @tschneidereit was asking about above?

littledan:
I've pushed a patch to make the change as described in https://github.com/WebAssembly/spec/pull/745#issuecomment-434645244 of including an event loop turn unconditionally for asynchronous instantiation. This is important to get right, partly because the semantics here also inform the WebAssembly/ESM integration semantics (where I imagine that this event loop turn will take place during module evaluation, using top-level await).

Reviews welcome. cc @Ms2ger @xtuc

document/js-api/index.bs
```diff
@@ -398,7 +397,42 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
 
         Note: the validity and uniqueness checks performed during [=WebAssembly module validation=] ensure that each property name is valid and no properties are defined twice.
     1. Perform ! [=SetIntegrityLevel=](|exportsObject|, `"frozen"`).
-    1. Let |instanceObject| be a new {{Instance}} object whose internal \[[Instance]] slot is set to |instance| and the \[[Exports]] slot to |exportsObject|.
+    1. Return a new {{Instance}} object whose internal \[[Instance]] slot is set to |instance| and the \[[Exports]] slot to |exportsObject|.
+</div>
+
+<div algorithm>
+  To <dfn>instantate the core of a WebAssembly module</dfn> from a module |module| and imports |imports|, perform the following steps:
```
xtuc:
```suggestion
  To <dfn>instantiate the core of a WebAssembly module</dfn> from a module |module| and imports |imports|, perform the following steps:
```

xtuc: "core of a WebAssembly module" sounds strange to me, what about "internals of a WebAssembly module"?

littledan:
@Ms2ger You've incrementally landed many parts of this PR. Would you be interested in rebasing it?

Ms2ger:
I rebased the PR; it should now be straightforward to review. @tschneidereit, others: thoughts?

lukewagner:
As discussed in the last CG call, I landed this [change](https://hg.mozilla.org/integration/mozilla-inbound/rev/4a546c424e47) in FF.  (It should make it into the next Nightly.)  I'll report back if we hear of any breakage, although I don't expect any.

xtuc (dismissed):

littledan:
@kmiller68, @tschneidereit and I have been discussing the semantics here. With this PR, each time a WebAssembly module is instantiated, there would be an event loop turn, in order to leave time for compilation. It sounds like this is needed for JSC the way it's implemented right now; it *may* be possible to rearchitect JSC for this purpose, but it's not clear yet whether that would lead to performance regressions.

I'd just like to confirm that we're comfortable with these semantics, as they're rather different from JavaScript semantics. The [ESM integration proposal](https://webassembly.github.io/esm-integration/js-api/index.html#module-execution) would carry over this per-module event loop turn.

In https://github.com/whatwg/html/issues/4400 , @nyaxt , @smaug---- and @rniwa expressed some skepticism about yielding to the event loop for each JavaScript module; this patch does something similar for WebAssembly.

It may be that this is OK for WebAssembly and not for JavaScript as WebAssembly is generally a small number of big modules, whereas JavaScript may be deployed in many small modules. We may also be able to specify an event loop turn here, and later decide that it's optional or missing (however, such a two-step change could have compatibility risk).

rossberg:
> WebAssembly is generally a small number of big modules, whereas JavaScript may be deployed in many small modules.

I'm not sure that we should be making assumptions of this sort.

littledan:
I worded that a bit clumsily; I meant more like, *if* we expect that, then that would be one story for why not to worry about this impact/alignment with JS.

smaug----:
Currently the web doesn't really have explicit event loop spinning (modulo some implementation details and the somewhat unclear alert() handling in the spec). I'd rather not see us adding such. It complicates many things quite a big. You may want to also ask blink folks why they wanted to remove showModalDialog rather quickly after finding some issue (IIRC it was causing major issues there because of nested event loop spinning, but perhaps I misremember).

If we really want event loop to spin (why?), then it needs to be clearly defined what happens if the world dies during spinning (closing window, or removing script element or modifying the contents of the script element etc.).

littledan:
To clarify, no one here is talking about spinning the event loop. Instead, instantiating a WebAssembly module would, in this patch, have the possibility to do asynchronous work (e.g., compilation based on the imports), before queueing a task to finish the instantiation.

lukewagner:
While writing the above-mentioned [patch](https://hg.mozilla.org/integration/mozilla-inbound/rev/4a546c424e47), I realized another use case for being able to do non-trivial (off-main-thread) compilation during instantiation after the import values are known: Web IDL Bindings where, for best perf, you want to generate thunks specialized to the caller+callee.

smaug----:
> To clarify, no one here is talking about spinning the event loop.

I see. Well, then just queue a task and continue. Of course the programming model is then a tad more error prone since the world may change before the task runs, but happens with other APIs too.

What needs to be defined is that does the new task block page's load event firing, if the initial step is executed during pageload.

littledan:
@smaug---- Nothing in this PR would affect when the loaded event fires, because loading a WebAssembly module does not relate to that event. WebAssembly is loaded with APIs like [instantiateStreaming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming#Examples), which would see one more task queued internally before running the start function; you can find examples of usage in that MDN page. With `<script type=module>` in the context of [Wasm/ESM integration](https://github.com/WebAssembly/esm-integration/tree/master/proposals/esm-integration), the effects would be similar, but also not change when those events fire.

lukewagner:
As an update, the [patch](https://hg.mozilla.org/integration/mozilla-inbound/rev/4a546c424e47) is about to advance to Beta.  No breakage yet reported.

smaug----:
Are there wpt tests ensuring page's load event isn't affected?

littledan:
I don't know how I would write such a test, since I am not sure what scenario you are concerned about. It doesn't seem plausible to me that the JS instantiateStreaming API would block the load event; there is currently no declarative way to fetch Wasm.

smaug----:
It has nothing to do with declarative or imperative. Just about fetching wasm.
If fetch is started right before load event for the page would otherwise fire, is the fetch finished before or after page's load event actually fires?

rossberg:
@Ms2ger, ping on this one.

rossberg:
Any updates on this PR?

rossberg:
This PR is 5 years old and received no reaction from multiple pings over the past year. Unless there is visible activity, I will take the liberty to close it.

Ms2ger:
As far as I can tell, this is ready to land.

rossberg:
@Ms2ger, sounds good. I'm not sure if @littledan is still available, could you take over landing this?

tomstuart:
This PR was opened against `master`, which was the default branch at the time, but five years later the default branch is now `main` and should’ve been merged into that branch instead. I’ve opened #1606 to target the correct branch.

PR #748, opened 2018-03-13
[test] Remove redundant kExprReturn

gahaas:
The .addBody function of the WasmModuleBuilder automatically adds kExprEnd at the end of a function. Therefore the explicit addition of kExprEnd produces invalid wasm code. If you want to add kExprEnd explicitly, you have to use .addBodyWithEnd.


hammacher:
LGTM.

Issue #727, opened 2018-02-28
[spec review] 3 - Validation


flagxor:
Detail oriented review of draft text:
http://webassembly.github.io/spec/core/bikeshed/index.html#validation%E2%91%A1

Try to be ready to discuss issues by the next in person WG meeting.

lukewagner:
I made a fairly-fine-grained pass through Section 3.  Overall, this section feels like it's in pretty good shape.
* General:
  * It would be really nice to put some form of box or shaded rectangle around/behind the deduction rules and all other mathematical definitions.  Otherwise, especially when combined with the awkward typesetting, they look out of place.
* § 3.1.1
  * "It is convenient to define"; could strengthen wording here; seems like more than a "convenience", but rather just a central definition used throughout the section.
  * Note on vectors: I'm not sure the Note helps more than it distracts, given that `*` is used to define mathematical lists throughout the entire spec.
  * "In addition to field access `C.field` the following notation is adopted": I'd expand `C.field` to ", written `C.field`,", otherwise sentence seems grammatically off.
  * Is the `C,field A*` notation ubiquitous or invented here?  If the latter: I've actually squinted at this before; is there perhaps another notation that "looks" a little more like "prepend to this field"?
* § 3.2.2.1
  * "This restriction may be removed in future versions of WebAssembly.": Given the distance, "This" seems ambiguous; it'd be good to call out the return type.
* § 3.3.4.1
  * I'd expand "width" to "bit width" to match linked definition.
* § 3.3.5.9
  * I had a bit of confusion here because `C.return` is effectively a `valtype??`. Further confusing me was the fact that `resulttype` is a `valtype?`, not a `valtype*` which the note in this section assumes. Could we: (1) perhaps use a more explicit variant type for C.return to indicate "not in a function", (2) change `resulttype` to be a `valtype*` (or, if not, change the note to not use `[]`)?
* § 3.3.7.2:
  * The formal rule for `C⊢instr end const` should I think be `C⊢instr* end const`
* § 3.4.8.5:
  * Seems like a good place for a note that the `const` restriction may be lifted in future versions.
* § 3.4.10:
  * In the prose description, I think `externtype*` and `externtype'*` are stale and should be changed to `it*` and `et*`.

rossberg:
@lukewagner, thanks for the review, see #752 for changes.

> * General:
[...]

Issue #725, opened 2018-02-28
[spec review] 1 - Introduction + Consistency


flagxor:
Detail oriented review of draft text:
http://webassembly.github.io/spec/core/bikeshed/index.html#introduction%E2%91%A0

Also includes front-matter + overall document consistency.

Try to be ready to discuss issues by the next in person WG meeting.

lukewagner:
Random driveby:
* § 1.1: "WebAssembly is an open standard developed by a W3C Community Group that includes representatives of all major browser vendors." - This sentence was meaningful in early readmes/intros when wasm was just getting started but it seems out of place permanently backed into a W3C spec (especially once wasm starts showing up in non-browser embeddings)

liminzhu:
Yeah I agree. Early on it was nice to show that all browsers were behind it but it's time for wasm to become a standard rather than perceived as a browser thing. Plus the standard is also technically developed by the WG? 

It's a small sample, but I looked up [service worker](https://www.w3.org/TR/service-workers-1/) and [webRTC](https://www.w3.org/TR/webrtc/) and none of them include similar text.

liminzhu:
Did an attentive pass on intro and a rough pass at the rest of the core spec for consistency issues. Overall the spec is in very good shape! Some feedback:

* https://github.com/WebAssembly/spec/issues/759
* In 1.1 `Its main goal is to enable high performance applications on the Web, but it does not make any Web-specific assumptions or provide Web-specific features, so it can be employed in other environments as well.` Wasm does provide web specific features in JS and web spec, so I suggest,
 ..., but **the core spec** does not make any Web-specific assumptions ...
* nit - there are lots of arbitrary uses of colon or period before bullet points or formulas. (e.g. [1](http://webassembly.github.io/spec/core/bikeshed/index.html#dependencies①) [2](http://webassembly.github.io/spec/core/bikeshed/index.html#concepts①) [3](http://webassembly.github.io/spec/core/bikeshed/index.html#grammar-notation①) [4](http://webassembly.github.io/spec/core/bikeshed/index.html#auxiliary-notation①)). Maybe we can use colon in places where phrases like 'the following' or 'as follow' are used immediately before points or formulas and use period otherwise, or just apply some general principle. Certainly low pri as it doesn't really affect readability or anything much.

rossberg:
Thanks for the review!

[...]

PR #752, opened 2018-03-16
[spec] Address comments by Luke

rossberg:
Addresses @lukewagner's comments on #725, #726, and #727, plus his [suggestion](https://github.com/WebAssembly/reference-types/pull/3#discussion_r174529288) for back-porting a minor refactoring of the elements grammar (analogously applied to data segments).

### DIFF
--- a/document/core/valid/conventions.rst
+++ b/document/core/valid/conventions.rst
@@ -40,14 +40,14 @@ which collects relevant information about the surrounding :ref:`module <syntax-m
 * *Globals*: the list of globals declared in the current module, represented by their global type.
 * *Locals*: the list of locals declared in the current function (including parameters), represented by their value type.
 * *Labels*: the stack of labels accessible from the current position, represented by their result type.
-* *Return*: the return type of the current function, represented as a result type.
+* *Return*: the return type of the current function, represented as a result type, or a special marker |NORETURN| indicating that no return is allowed, as in free-standing expressions.
 
 In other words, a context contains a sequence of suitable :ref:`types <syntax-type>` for each :ref:`index space <syntax-index>`,
 describing each defined entry in that space.
 Locals, labels and return type are only used for validating :ref:`instructions <syntax-instr>` in :ref:`function bodies <syntax-func>`, and are left empty elsewhere.
 The label stack is the only part of the context that changes as validation of an instruction sequence proceeds.
 
-It is convenient to define contexts as :ref:`records <notation-record>` :math:`C` with abstract syntax:
+More concretely, contexts are defined as :ref:`records <notation-record>` :math:`C` with abstract syntax:
 
 .. math::
    \begin{array}{llll}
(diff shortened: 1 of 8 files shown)


document/core/valid/conventions.rst
```diff
@@ -69,14 +69,12 @@ In addition to field access written :math:`C.\K{field}` the following notation i
 * When spelling out a context, empty fields are omitted.
   Likewise, the |CRETURN| field may be omitted when it is |NORETURN|.
 
-* :math:`C,\K{field}\,A^\ast` denotes the same context as :math:`C` but with the elements :math:`A^\ast` prepended to its :math:`\K{field}` component sequence.
+* The notation :math:`\K{field}\,A^\ast, C` denotes the same context as :math:`C` but with the elements :math:`A^\ast` prepended to its :math:`\K{field}` component sequence.
```
lukewagner: Do you have a rendered link I could see?  Naively, it looks like it would read `labels A, C` which is different (and more confusing, imho) than the `C.A labels` or `A C.labels` I was expecting.  Actually, what about parens: `C.(A labels)`?  You could maybe even hyperlink the ".(" and ") to this particular rule...

rossberg: This is the main repo, but since all the changes are editorial I took the liberty to upload it to the main GH.io.

I agree that this isn't ideal either, and after a couple of days I like it less. :)  But I am somewhat reluctant to introduce extra indexing notation for one single use case. I'm almost tempted to stay with the original version, and just trying to make the note clearer. WDYT?

lukewagner: Yeah, looking at the rendered version is, imho, just as confusing as before.  The problem with improving the note is that the use sites are still hard to read.  I still think both `C.(A* labels)` and `C.labels﻿﻿⟬l⟭` (with hyperlink to the explanation in conventions) would be a lot more readable.  As another idea: could we do something with `with` (which already seems to be in the overall conventions for field update)?

rossberg: Okay, I reverted the inverse context extension notation.

Hm, dot notation expresses *projection*, while extension is a form of *injection*, i.e., the opposite, so "C.(A* labels)" would be backwards. In general, I think it's preferable to stick as close to standard notation as possible, so that readers with prior exposure to type systems can understand it as is and aren't confused for no reason -- ultimately there is nothing magic going on here, it's just standard de Bruijn indexing.

I would be surprised if introducing multiple different lookup notations would help clarity for most readers. Conceptually, I also think that lookup is the wrong place to tweak: there is nothing noteworthy about the lookup of labels, the crux is in how the index space is organised, i.e., how the bindings are introduced.

Maybe it was the note that was just more confusing than necessary? I gave another shot at trying to make it clearer. I also added extra notes at the extension and lookup sites. Let me know what you think.

lukewagner: Fair enough and good notes; lgtm!

rossberg:
Thanks!

PR #755, opened 2018-03-19
Use async compilation in the Web Platform Tests

gahaas:
The tests generated by test/build.py use the synchronous WebAssembly API to compile and instantiate WebAssembly modules and instances, respectively, i.e. "new WebAssembly.Module()" and "new WebAssembly.Instance()". This is not ideal, because the preferred way to compile and instantiate WebAssembly modules is "WebAssembly.compile()" and "WebAssembly.instantiate()". In addition, for "new WebAssembly.Module()" there exists an input size restriction of (I think) 4KB.

This PR contains the following changes:

* a new JS test harness, async_index.js, which uses the asynchronous JS API. To make sure that all tests run in the correct order, the promises returned by the JS API calls are connected by a promise chain, see the "chain" variable in async_index.js.

* I renamed the existing JS test harness from index.js to sync_index.js

* I adjusted test/build.py to use async_index.js by default, and optionally sync_index.js if specified by a parameter.

* I made some small fixes in test/build.py, e.g. the WPT harness was not included properly in the html version of the generated tests.


test/build.py
```diff
@@ -137,27 +137,31 @@ def ignore(_src, names):
...
-def build_html(html_dir, js_dir):
+def build_html(html_dir, js_dir, use_sync):
     print("Building HTML tests...")
 
     js_html_dir = os.path.join(html_dir, 'js')
 
-    build_html_js(js_html_dir, js_dir)
+    build_html_js(js_html_dir, js_dir, include_harness=True)
+    for js_file in glob.glob(os.path.join(js_html_dir, '*.js')):
+        wrap_single_test(js_file)
```
bnjbvr: ditto, this now happens after every call to `build_html_js`, so it could be part of this function.

PR #758, opened 2018-03-22
[test] Import referring undefined type is invalid

pepyakin:
Check that if function import descriptor referres undefined type then the module is invalid.


pepyakin:
Ah, I finally found one, 
https://github.com/WebAssembly/spec/blob/4f078ff69de51d07c1c265273e396f42b2fb2dd3/test/core/func_ptrs.wast#L49

rossberg:
Yeah, that test is a bit misplaced, so I wouldn't mind landing this PR anyway. :)

Issue #675, opened 2018-02-13
[bikeshed] Limit TOC to 3 levels


rossberg:
The deeper levels aren't really sections but rather section*.

flagxor:
Fixed.

Issue #669, opened 2018-02-13
[bikeshed] Wrong fonts and spacing


rossberg:
The Bikeshed version of the spec does not seem to respect \mathit and \mathrm formats in some contexts. Consequently, all letters are type-set as individual variables in italic font and spacing and kerning is off.

(See e.g. rendering of the "s with [i] = e" notation [here](http://flagxor.github.io/spec/core/bikeshed/index.html#auxiliary-notation①) or the rendering of variable names like "i16" in the note [here](http://flagxor.github.io/spec/core/bikeshed/index.html#integers①).)

Note: Cases like "with" and "if" are probably due to the use of \mbox?

flagxor:
The bad kerning seems to be a general issue with katex:
https://khan.github.io/KaTeX/

Filed this:
https://github.com/Khan/KaTeX/issues/1259

flagxor:
Worked around it for now.

rossberg:
It seems like e.g. the "with" notation linked to in the OP is still rendered unreadable.

rossberg:
Apparently I was looking at the wrong copy.

rossberg:
It seems like there are still places where spaces get swallowed, see e.g.
http://webassembly.github.io/spec/core/bikeshed/index.html#tokens①
http://webassembly.github.io/spec/core/bikeshed/index.html#modules⑦  (foralli at bottom of section)

binji:
> It seems like there are still places where spaces get swallowed, see e.g.
http://webassembly.github.io/spec/core/bikeshed/index.html#tokens①
http://webassembly.github.io/spec/core/bikeshed/index.html#modules⑦ (foralli at bottom of section)

These issues seem to be fixed by the update to v0.10.0, and by replacing `\mbox` with `\text`. PTAL

rossberg:
Indeed, excellent, thanks!

Issue #667, opened 2018-02-13
[bikeshed] Math links don't work


rossberg:
Links to math definitions are missing from the Bikeshed version of the core spec.

flagxor:
Fixed.

PR #765, opened 2018-04-07
Fixing up formatting for bikeshed.

flagxor:
Limiting TOC depth.
  https://github.com/WebAssembly/spec/issues/675
Bringing back hyperlinks.
  https://github.com/WebAssembly/spec/issues/667
Fix missing spaces.
  https://github.com/WebAssembly/spec/issues/669
Working around katex bug with multi-line + hrefs.

### DIFF
--- a/document/core/util/mathdefbs.py
+++ b/document/core/util/mathdefbs.py
@@ -1,0 +1,126 @@
+# Version of mathdef.py for bikeshed build.
+# Tweaked to generate single page links.
+# TODO(bradnelson): Figure out a way to merge this back into
+# mathdef.py controlled by buildername.
+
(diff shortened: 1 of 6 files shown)


document/core/util/mathdefbs.py
```diff
@@ -0,0 +1,121 @@
+from sphinx.ext.mathbase import math
```
rossberg: Can you insert some comment here explaining how this differs from mathdefs? Do you think it's possible to avoid the code duplication with some config parameter?

flagxor: Spent an hour trying to find such a way just now to no avail.
Added comment and a TODO.

Issue #673, opened 2018-02-13
[bikeshed] Spurious grey bar


rossberg:
The subitem of [step 9](http://flagxor.github.io/spec/core/bikeshed/index.html#instantiation②) of module instantiation have a spurious grey bar to the left that does not appear elsewhere in the algo. Also appears in a couple of other places e.g. [in this section](http://flagxor.github.io/spec/core/bikeshed/index.html#modules①⑤).

tabatkins:
This isn't a CSS issue; the HTML is broken.  For some unknown reason, the algorithm HTML is using `<blockquote>` on some of the sub-algorithms.  By default blockquotes get the gray-line styling.  That's just the wrong markup and should be removed.

tabatkins:
Ah, I see. The markup problem is clearer in Andreas' [second link](http://flagxor.github.io/spec/core/bikeshed/index.html#modules%E2%91%A0%E2%91%A4) - rather than the nested steps being represented by a nested `ol`, the markup is instead:

```
<ol><li>...</ol>
<blockquote>
  <ol>...</ol>
</blockquote>
<ol><li>...</ol>
```

That is, the blockquote is being used for stylistic purposes to indent the inner algorithm, and the steps of the outer algorithm are broken into two separate lists. This is all kinds of bad HTML.

PR #775, opened 2018-04-10
Fixing borders + gray bar.

flagxor:
Removes extra border on footnotes.
Fixes:
https://github.com/WebAssembly/spec/issues/673


tabatkins:
As cited in #673, this fix should be reverted; the problem is broken HTML rather than anything wrong with the CSS.

PR #777, opened 2018-04-10
[editorial] Use mathdef for signed leb to prevent double href.

flagxor:
Fixing a place where we get nested <a hrefs>
(fails W3C HTML checker).


document/core/binary/types.rst
```diff
@@ -25,7 +25,7 @@ Value Types
 
 .. note::
    In future versions of WebAssembly, value types may include types denoted by :ref:`type indices <syntax-typeidx>`.
-   Thus, the binary format for types corresponds to the encodings of small negative :math:`\xref{binary/values}{binary-sint}{\sN}` values, so that they can coexist with (positive) type indices in the future.
+   Thus, the binary format for types corresponds to the encodings of small negative :math:`\BsN` values, so that they can coexist with (positive) type indices in the future.
```
rossberg: Hm, that makes the wording somewhat nonsensical. How about moving the link to "encoding" and leave \sN as is?

```
the :ref:`encoding <binary-sint>` of small negative |sN| values
```

PR #782, opened 2018-04-16
[spec] Reorganise main links

rossberg:
Make links more explicit and equal.


rossberg:
@flagxor, friendly ping.

PR #807, opened 2018-05-18
Miscellanous floating-point and unicode tests

sunfishcode:
This adds a handful of miscellaneous floating-point and Unicode tests.

### DIFF
--- a/test/core/float_exprs.wast
+++ b/test/core/float_exprs.wast
@@ -1022,7 +1022,7 @@
 (assert_return (invoke "f32.no_fold_lt_select_to_abs" (f32.const 0.0)) (f32.const 0.0))
 (assert_return (invoke "f32.no_fold_lt_select_to_abs" (f32.const -0.0)) (f32.const -0.0))
 (assert_return (invoke "f32.no_fold_le_select_to_abs" (f32.const nan:0x200000)) (f32.const nan:0x200000))
-(assert_return (invoke "f32.no_fold_le_select_to_abs" (f32.const -nan))(f32.const -nan))
+(assert_return (invoke "f32.no_fold_le_select_to_abs" (f32.const -nan)) (f32.const -nan))
 (assert_return (invoke "f32.no_fold_le_select_to_abs" (f32.const 0.0)) (f32.const -0.0))
 (assert_return (invoke "f32.no_fold_le_select_to_abs" (f32.const -0.0)) (f32.const 0.0))
 (assert_return (invoke "f32.no_fold_gt_select_to_abs" (f32.const nan:0x200000)) (f32.const -nan:0x200000))
--- a/test/core/float_literals.wast
+++ b/test/core/float_literals.wast
@@ -37,6 +37,9 @@
   (func (export "f32_dec.max_finite") (result i32) (i32.reinterpret/f32 (f32.const 3.4028234e+38)))
   (func (export "f32_dec.trailing_dot") (result i32) (i32.reinterpret/f32 (f32.const 1.e10)))
 
+  ;; https://twitter.com/Archivd/status/994637336506912768
+  (func (export "f32_dec.root_beer_float") (result i32) (i32.reinterpret/f32 (f32.const 1.000000119)))
+
   ;; f64 special values
   (func (export "f64.nan") (result i64) (i64.reinterpret/f64 (f64.const nan)))
   (func (export "f64.positive_nan") (result i64) (i64.reinterpret/f64 (f64.const +nan)))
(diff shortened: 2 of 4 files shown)


test/core/const.wast
```diff
@@ -68,6 +70,14 @@
   (module quote "(func (f32.const -0x1.ffffffp127) drop)")
   "constant out of range"
 )
+(assert_malformed
+  (module quote "(func (f32.const 0x1.fffffefffffff8000000p127) drop)")
```
binji: This test fails in wabt, but I'm curious about the behavior here. Doesn't this mean that even when parsing a 32-bit float, it is parsed and rounded first as a 64-bit float? I noticed that if you try the same thing in clang, you get the following:

```
// clang
0x1.fffffefffffff8000000p127  => float +Inf
0x1.fffffefffffff8000000p127f => float 3.40282347E+38
```
(see https://godbolt.org/g/ezwtex)

We have similar behavior for non-hex floats:

```
// interpreter
f32.const 3.4028235677973366E+38 => constant out of range

// clang
3.4028235677973366E+38  => float +Inf
3.4028235677973366E+38f => float 3.40282347E+38
```

I guess I'm just wondering if this is the desired behavior or if I've gotten confused. :-)

rossberg: @binji:
> ```
> // clang
> 0x1.fffffefffffff8000000p127  => float +Inf
> 0x1.fffffefffffff8000000p127f => float 3.40282347E+38
> ```

That seems super-wierd to me. Why does it overflow to infinity for the type with the *larger* value range, but not for the type with the smaller one?

> Doesn't this mean that even when parsing a 32-bit float, it is parsed and rounded first as a 64-bit float?

Yeah, you are right, something is fishy here. But I think it's actually the other way round: this test should be correct, but the new ones on lines 55-56 are bogus.

F32 has 23 bits of mantissa. So 0x1.ffff_fe is the largest mantissa you can represent accurately. Anything larger that still rounds down should be fine, i.e., everything up to 0x1.ffff_fe7fffffffff.... OTOH, starting from 0x1.ffff_fe800000000 it should try to round up to 0x1.ffff_ff (round-to-nearest-tie-to-even), which is out of range and thereby becomes infinity.

So I don't know what exactly these tests are supposed to check, but it seems wrong, while apparently some implementations (including the reference interpreter) expose matching incorrect behaviour on some of them. What seems to be going on is that all these implementations go through f64 and then run into strange double rounding errors (which I cannot quite explain).

So I think the correct tests need to be:

wellformed: (i32.const 0x1.ffff_fe7fffffffffp127) (i32.const -0x1.ffff_fe7fffffffffp127)
malformed: (i32.const 0x1.ffff_fe80p127) (i32.const -0x1.ffff_fe80p127)

And then we need to fix implementations.

rossberg: Ah, sorry, tie-to-even of course implies that 0x1.ffff_fe80 is still okay, because it also rounds down. But 0x1.ffff_fe800000000000000001 should round up. Bummer. The question then is, how do we implement that?

binji: > strange double rounding errors (which I cannot quite explain)

The way I understand it, the value is parsed as an f64, so it checks the 53rd bit of the significand to determine rounding behavior. For `0x1.fffffefffffff8p127` the bit is set (the `8`), so it rounds up to `0x1.ffffffp127` this then is converted to an f32, which is larger than the maximum finite value. This also explains why `0x1.fffffefffffff7ffffffp127` is parsed the same as `0x1.fffffep127`.

> But 0x1.ffff_fe800000000000000001 should round up.

Yes, that makes sense to me as long as we assume that the value should be parsed with infinite precision and then rounded. But it seems that at least the C compilers I tried don't have this behavior, and instead the additional hex-digits after the precision are just ignored. That would explain why parsing
`0x1.fffffefffffff8000000p127f` as an f32 works, I think.

sunfishcode:
@rossberg I agree. The tests proposed here test the limits of the the current interpreter implementation does, but ideally, it should effectively interpret the input as a floating-point value with infinite precision, then round to the destination type. I agree with your assessment of:
> wellformed: (i32.const 0x1.ffff_fe7fffffffffp127) (i32.const -0x1.ffff_fe7fffffffffp127)
> malformed: (i32.const 0x1.ffff_fe80p127) (i32.const -0x1.ffff_fe80p127)

(except that the `i32`s should be `f32`). I'll remove the patch which adds these tests from the PR here, so that the other patches here can be merged, and I'll file a new issue.

sunfishcode:
I've now removed the test for floating-point literals being rounded to overflow from this PR. And, I've now filed https://github.com/WebAssembly/spec/issues/809.

binji:
Let's merge this.

PR #812, opened 2018-05-30
[test] Remove non-conformance checkponits

Honry:
Spec allows maximum memory size is 4GiB, but it does not restrict
  all implementations should support allocating 4GiB memory.


test/core/address.wast
```diff
@@ -101,3 +101,25 @@
   )
   "i32 constant"
 )
+
+;; Test that load/store addreses are full int32 (or int64), and not OCaml int
```
rossberg: Nit: can you remove the "not Ocaml int" part? I think it neither makes sense nor is it relevant.

Honry: Sure :)

Honry:
@rossberg, can we land this?

binji:
The address is the first operand to load/store, so this is just testing that you can store a full 32-bit value, not the address.

Honry:
@binji, many thanks for you comment!
You are correct, I did misunderstand the checkpoint. Have fixed in new commit, PTAL. Thanks!

Honry:
Totally agree, this is not appropriate as a conformance test. 

> - test that when allocating 4GiB, accessing index -1 fails

This one is improper as well.

So how about just drop these non-conformance checkpoints?

rossberg:
I agree we should just drop these points.

Honry:
@rossberg, @binji, many thanks for your comments, I've refurbished this PR to remove non-conformance checkpoints. PTAL.

PR #815, opened 2018-06-07
Mark `value` to be an optional argument

saschanaz:
As non-optional arguments cannot have a default value. (https://heycam.github.io/webidl/#prod-ArgumentRest)


binji:
@littledan an alternate PR with the same fix was landed here: https://github.com/WebAssembly/spec/pull/818

Issue #816, opened 2018-06-07
Default value in constructor of Global


dontcallmedom:
The current definition of the constructor for `Global` uses a default value for the 2nd argument:
  `Constructor(GlobalDescriptor descriptor, unrestricted double value = 0)`

AFAIU, default values can only be used with optional arguments - so I think `value` should be marked as `optional`.

rossberg:
Actually, I think there is another problem here, namely that the type of the value depends on the descriptor. In the future, it might denote a non-numeric type, so double is not adequate. It should probably be `any` and default to `undefined`.

binji:
@rossberg That's a good point, but we discussed in the meeting having globals of the numeric types default to 0. So I suppose we need to handle it specially in the constructor's algorithm.

binji:
Fixed in #818.

PR #818, opened 2018-06-07
[spec] Use `any` in `Global` constructor

binji:
See discussion in issue #816. In the future, we'll want to store
non-numeric values in a `WebAssembly.Global` object, so using
`unrestricted double` is over-constrained.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -720,10 +720,10 @@ dictionary GlobalDescriptor {
   boolean mutable = false;
 };
 
-[LegacyNamespace=WebAssembly, Constructor(GlobalDescriptor descriptor, unrestricted double value = 0), Exposed=(Window,Worker,Worklet)]
+[LegacyNamespace=WebAssembly, Constructor(GlobalDescriptor descriptor, optional any value = undefined), Exposed=(Window,Worker,Worklet)]
 interface Global {
-  unrestricted double valueOf();
-  attribute unrestricted double value;
+  any valueOf();
+  attribute any value;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -753,11 +753,19 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     1. Otherwise, throw a {{TypeError}} exception.
 </div>
 
+<div algorithm>
+    The algorithm <dfn>DefaultGlobalValue</dfn>(|valuetype|) performs the following steps:
```
rossberg: Nit: I would probably just call this DefaultValue, since it doesn't have any dependency on globals and might potentially be useful in other contexts.

saschanaz:
Web IDL currently [does not support `undefined`](https://github.com/heycam/webidl/issues/60).

document/js-api/index.bs
```diff
@@ -720,10 +720,10 @@ dictionary GlobalDescriptor {
   boolean mutable = false;
 };
 
-[LegacyNamespace=WebAssembly, Constructor(GlobalDescriptor descriptor, unrestricted double value = 0), Exposed=(Window,Worker,Worklet)]
+[LegacyNamespace=WebAssembly, Constructor(GlobalDescriptor descriptor, optional any value = undefined), Exposed=(Window,Worker,Worklet)]
```
saschanaz: Web IDL currently [does not support `undefined`](https://github.com/heycam/webidl/issues/60). How about just removing `= undefined`?

binji: Done.

PR #820, opened 2018-06-09
[spec] Normative: Remove ahead of time bounds checks from instantiation

conrad-watt:
See https://github.com/WebAssembly/threads/issues/94 for background. The current specification does not extend neatly to the multi-threaded case.

This PR corresponds to potential solution 2, which received a broadly positive initial response, and would allow data initialisers to be used with shared memories.

I choose to make the per-initialiser bounds checks result in a Trap/RuntimeError rather than a Fail/LinkError, since this makes the behaviour indistinguishable from using bulk memory operations in the start function, and I believe users might expect LinkError to guarantee no side effects have taken place.

### DIFF
--- a/document/core/exec/modules.rst
+++ b/document/core/exec/modules.rst
@@ -548,7 +548,7 @@ Given a :ref:`store <syntax-store>` :math:`S`, a :ref:`module <syntax-module>` :
 
 Instantiation checks that the module is :ref:`valid <valid>` and the provided imports :ref:`match <match-externtype>` the declared types,
 and may *fail* with an error otherwise.
-Instantiation can also result in a :ref:`trap <trap>` from executing the start function.
+Instantiation can also result in a :ref:`trap <trap>` from initializing :ref:`element <syntax-elem>` or :ref:`data <syntax-data>` segments, or from executing the start function.
 It is up to the :ref:`embedder <embedder>` to define how such conditions are reported.
 
 1. If :math:`\module` is not :ref:`valid <valid-module>`, then:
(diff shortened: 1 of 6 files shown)


document/core/exec/modules.rst
```diff
@@ -613,10 +613,6 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
 
     h. Let :math:`\X{eend}_i` be :math:`\X{eo}_i` plus the length of :math:`\elem_i.\EINIT`.
```
rossberg: Nit: Maybe move this down to where it is actually used first now.

document/core/exec/modules.rst
```diff
@@ -635,17 +631,17 @@ It is up to the :ref:`embedder <embedder>` to define how such conditions are rep
 
     h. Let :math:`\X{dend}_i` be :math:`\X{do}_i` plus the length of :math:`\data_i.\DINIT`.
```
rossberg: Same here.

document/core/exec/modules.rst
conrad-watt: done

document/core/exec/modules.rst
conrad-watt: done

conrad-watt:
12th is probably too soon, so maybe the meeting on the 26th?

rossberg:
Sounds good. I won't be able to attend the 12th anyway.

binji:
Added [future meeting item here](https://github.com/WebAssembly/meetings/pull/255)

conrad-watt:
I'll keep this open until at least the next CG, but it seems the consensus from this week's meeting was that we should go with this solution, but bundle it with the bulk memory proposal.

AndrewScheidecker:
Is it practical to further relax this to do the bounds check for each byte/element copied?

As noted in the original issue, an extension that adds shrinking memories/tables or page protection makes it impractical to ensure the copy will be all-or-nothing.

conrad-watt:
After @jfbastien pointed this out, we concluded in https://github.com/WebAssembly/threads/issues/96 that we would need to do basically as you propose. The current idea is that segment initialization and bulk operations will (observably) perform writes from low to high addresses in order at byte/element granularity until oob. I think this still needs to be fleshed out in the proposal repo, however.

binji:
Closing, as this behavior is moved to the bulk memory proposal.

PR #824, opened 2018-06-16
WIP: Test that one can't declare more than 2^32 locals

### LINKED ISSUE #823: Is it possible to have more than 2^32 locals per function?
The only thing I found in the spec is that the limitation of locals count is [determined by the embedder](https://webassembly.github.io/spec/core/appendix/implementation.html).

However, I believe, that this is not accurate. For one, you can't access locals with indices greater than 2^32. This is because [`{get,set,tee}_local` has a 32-bit immediate index](https://webassembly.github.io/spec/core/binary/modules.html#binary-localidx).

I inspected only one implementation for this matter, wabt, and it turned out it is susceptible for [integer overflow](https://github.com/WebAssembly/wabt/blob/d23c99d036ab31f7de21bc4a90b651c4cfd4a356/src/binary-reader-interp.cc#L1148). (cc/ @binji )

So, is it possible for **any** implementation to have more than 2^^32?
If no, should we add the test for that case?

pepyakin:
Closes #823 

`./wasm` reports me this

```
./wasm: uncaught exception Stack overflow
Raised by primitive operation at file "parsing.ml", line 164, characters 4-28
```

And i'm not sure how to proceed there...

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -325,3 +325,20 @@
   )
   "zero flag expected"
 )
+
+;; One can not declare more than 2^32 locals.
+(assert_invalid 
+  (module binary
+    "\00asm" "\01\00\00\00"
+    "\01\04\01\60\00\00"       ;; Type section
+    "\03\02\01\00"             ;; Function section
+    "\0a\0c\01"                ;; Code section
+
+    ;; function 0
+    "\0a\02"
+    "\ff\ff\ff\ff\0f\7f"       ;; 0xFFFFFFFF i32
+    "\01\7e"                   ;; 0x00000001 i64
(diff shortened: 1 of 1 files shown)


rossberg:
Hm, that is a bit weird. I cannot reproduce that stack overflow and there is no module named `parsing.ml` in the interpreter, so I don't quite get that. On my laptop the example just never completes because it tries to allocate a list of 2^32 elements, which would probably require 96 GB or more of memory -- I aborted it after a few minutes.

But either way, the relevant check for the total count was missing. See #825 for an implementation that fast-paths the failure case. It includes a version of your test.

pepyakin:
Then let's close this issue in favor of #825 !

PR #828, opened 2018-06-22
Fix some editorial issues with the JS API specification.

Ms2ger:


binji:
> With respect to the W3C process, is it too late to merge this PR?

I discussed w/ Brad, and we're currently past the 90-day mark anyway, so we're always at least 90-days out at this point. Which is to say, it's not too late. :)

littledan:
>  he also said that that in general, non-functional changes like this do not affect the clock.

OK, merging this patch then.

littledan:
@Ms2ger Could you rebase? It seems that there is a conflict.

Ms2ger:
Rebased.

Issue #827, opened 2018-06-22
[js-api] Allow creating i64 globals


rossberg:
The JS API does not yet support i64 for functions or globals because JS cannot represent int64.

However, there is no reason to disallow *creating* WebAssembly.Global instances of type i64. It is sufficient to disallow *accessing* them from JS through the getter and setter. After all, such objects can already be created indirectly by exporting an i64 global from a module.

I hence suggest the following relaxations:

* The algorithm ToValueType does not throw on "i64" but return i64.
* The algorithm DefaultValue does not throw on i64 but return (i64.const 0).

That way, a Global of type i64 can be created in JS but only with a default initialisation value.

Motivation: It is problematic for a generic linker, mocker or other module-level abstractions if they cannot externally create every possible export object (even if unusable in JS).

littledan:
Given that BigInt is already relatively far along, I wonder if it would make sense to make this change together with [the WebAssembly i64/BigInt integration proposal](https://github.com/WebAssembly/spec/pull/707). This could leave the world with fewer states in the compatibility matrix to worry about. Anyway, @rossberg 's proposal doesn't seem so bad to me.

rossberg:
BigInt support is definitely not going to be in 1.0. This suggestion is to tweak what we ship in 1.0, so not gonna change the size of the matrix.

littledan:
If WebAssembly has already shipped to the web without this feature, it's part of the matrix even if the 1.0 document already includes the feature.

rossberg:
The Global constructor has not shipped in most browsers yet.

xtuc:
This issue has been closed by https://github.com/WebAssembly/spec/pull/829.

During https://github.com/WebAssembly/meetings/pull/268, we discussed about coercing values to/from i64 for function exports and globals but It's not covered by https://github.com/WebAssembly/spec/pull/829.

PR #829, opened 2018-06-24
Allow creation of default i64 global values

binji:
See issue #827.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -748,6 +748,7 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
 <div algorithm>
     The algorithm <dfn>ToValueType</dfn>(|s|) performs the following steps:
     1. If |s| equals "i32", return [=𝗂𝟥𝟤=].
+    1. If |s| equals "i64", return [=𝗂𝟨𝟦=].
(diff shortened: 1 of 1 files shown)


binji:
Good point, though this is a relatively minor change, we should discuss it in the CG.

littledan:
I'm fine with it not going through a stage process, but process-wise, you might want to lay out a lightweight set of conditions for a tiny normative change like this to be added to the standard. For example, TC39 requires conformance tests for normative changes, even if they don't go through the stage process.

binji:
That's a good idea, though currently it seems that we've just taken a poll for changes like this. Perhaps there needs to be more for direct changes to the spec, however.

lars-t-hansen:
IMO this needs a CG poll.  While I agree this ought to be uncontroversial and I contemplated including this behavior when implementing globals in SpiderMonkey, it's a visible non-bug-fix spec change and should not quietly slip in.

rossberg:
I agree this should have a CG poll.

binji:
Added to tomorrow's CG, though the agenda is pretty full so it may be pushed to the next meeting.

binji:
We discussed this in the June 26th CG meeting and decided to include this change in the v1 spec.

littledan:
Is this ready to merge then?

PR #832, opened 2018-07-06
[test] Enhance test coverage of address.wast

Honry:
Current address.wast only cover partial load operators with
 different offset and align combinations, this PR fulfills
 the rest operators with consistent combinations


Honry:
@rossberg, @binji, could you PTAL? Thanks!

PR #833, opened 2018-07-10
[test] Remove tested checkpoints in Todo.md

Honry:
Following 3 checkpoints have been tested in [memory_grow.wast(L36-L60)](https://github.com/WebAssembly/spec/blob/master/test/core/memory_grow.wast#L36), [memory.wast(L52-L75)](https://github.com/WebAssembly/spec/blob/master/test/core/memory.wast#L52), [align.wast(L436)](https://github.com/WebAssembly/spec/blob/master/test/core/align.wast#L436) respectively
 - test that too-big `memory.grow` fails appropriately
 - test that too-big linear memory initial allocation fails
 - test that an i64 store with 4-byte alignment that's 4 bytes out of bounds traps without storing anything.

### DIFF
--- a/test/Todo.md
+++ b/test/Todo.md
@@ -4,10 +4,7 @@ have a link to an open issue/PR, or be obvious. Comments/corrections/additions
 welcome.
 
 Linear memory semantics:
- - test that too-big `memory.grow` fails appropriately
- - test that too-big linear memory initial allocation fails
  - test that one can clobber the entire contents of the linear memory without corrupting: call stack, local variables, program execution.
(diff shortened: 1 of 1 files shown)


binji:
> test that an i64 store with 4-byte alignment that's 4 bytes out of bounds traps without storing anything.

I don't think this test case is covered by the code you referenced. That test is a validation error, but the notes are expecting a runtime trap.

Honry:
@binji, thanks! You are right, we need test for this checkpoint, I've added one at https://github.com/WebAssembly/spec/pull/834, PTAL.

PR #834, opened 2018-07-12
[test] Add test to cover one checkpoint in Todo.md

Honry:
- test that an i64 store with 4-byte alignment that's 4 bytes
   out of bounds traps without storing anything

### DIFF
--- a/test/core/align.wast
+++ b/test/core/align.wast
@@ -848,3 +848,13 @@
 (assert_return (invoke "i64_align_switch" (i32.const 6) (i32.const 2)) (i64.const 10))
 (assert_return (invoke "i64_align_switch" (i32.const 6) (i32.const 4)) (i64.const 10))
 (assert_return (invoke "i64_align_switch" (i32.const 6) (i32.const 8)) (i64.const 10))
+
+;; Test that an i64 store with 4-byte alignment that's 4 bytes out of bounds traps without storing anything
+
+(module
+  (memory 1)
+  (func (export "store") (param i32 i64)
+    (i64.store align=4 (get_local 0) (get_local 1))
+  )
+)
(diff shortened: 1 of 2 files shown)


test/core/align.wast
```diff
@@ -848,3 +848,13 @@
...
+
+;; Test that an i64 store with 4-byte alignment that's 4 bytes out of bounds traps without storing anything
+
+(module
+  (memory 1)
+  (func (export "store") (param i32 i64)
+    (i64.store align=4 (get_local 0) (get_local 1))
+  )
+)
+(assert_trap (invoke "store" (i32.const 65531) (i64.const 1)) "out of bounds memory access")
```
rossberg: You want to store at 65532 (or 0xfffc). Also, the main point of the item is checking that the memory has not been modified. So you want to store e.g. -1 and then after the trap have a check that the memory is still zero.

Honry: @rossberg, thanks for comment! 
It's my mistake, have addressed in new commit, PTAL.

test/core/align.wast
```diff
@@ -856,5 +856,11 @@
   (func (export "store") (param i32 i64)
     (i64.store align=4 (get_local 0) (get_local 1))
   )
+  (func (export "load") (param i32) (result i32)
+    (i32.load (get_local 0))
+  )
 )
-(assert_trap (invoke "store" (i32.const 65531) (i64.const 1)) "out of bounds memory access")
+
+(assert_trap (invoke "store" (i32.const 65532) (i64.const 1)) "out of bounds memory access")
```
rossberg: Perhaps better store -1, so that all bits would be set.

Honry: Indeed, I just updated it. :)

PR #837, opened 2018-07-23
[test] Add some tests for `br_if`

Honry:
- Add missing assertions for funcs: "as-block-first-value" and "as-block-mid-value"
 - Use `br_if` with multiple values


test/core/br_if.wast
```diff
@@ -147,6 +147,12 @@
       )
     )
   )
+
+  (func (export "br_if-multi-values") (param i32) (result i32)
```
rossberg: This test fits better into unwind.wast, which has similar tests for the other branch instructions already. Also, don't name it multi-value since that is confusing with actual multi value semantics.

test/core/unwind.wast
```diff
@@ -70,6 +70,11 @@
       (i32.const 3) (block (i64.const 1) (return (i32.const 9)))
     )
   )
+  (func (export "br_if-unwind-by-last-value") (param i32) (result i32)
```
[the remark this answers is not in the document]
rossberg: Do you still plan to address this comment as well?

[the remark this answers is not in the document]
Honry: @rossberg, thanks for your review! I am OK to address it in this PR.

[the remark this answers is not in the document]
Honry: @rossberg, I fail to compile it when moving the unused operand before `br_if`, see code snippet below:
```
(func (export "block-unwind-by-br_if-value") (result i32)
  (block (result i32)
    (i32.const 3) (i64.const 1) (br_if 0 (i32.const 9) (i32.const 1))
  )
)
```

Is it a bug or by design?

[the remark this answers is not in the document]
rossberg: This doesn't have anything to do with the syntax. But because br_if is not an unconditional branch, you have to insert a couple of `drops` after it to flush the superflous stack operands on the path where the branch is not taken. (In your previous version, the `return` had the same effect.)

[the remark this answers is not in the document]
Honry: Many thanks for your explanation! Really useful!
I just add a new commit, but some tests are still failed to be compiled. Their root cause look like the same, the operator type before `br_if` and the result type are mismatched, I suppose that's why it fails but have no idea on how to fix it. :(

I commented out these tests, Could you PTAL?

test/core/unwind.wast
```diff
@@ -70,18 +92,18 @@
...
 
   (func (export "unary-after-unreachable") (result i32)
     (f32.const 0) (unreachable) (i64.eqz)
   )
   (func (export "unary-after-br") (result i32)
     (block (result i32) (f32.const 0) (br 0 (i32.const 9)) (i64.eqz))
   )
+  ;;(func (export "unary-after-br_if") (result i32)
+  ;;  (block (result i32)
+  ;;    (f32.const 0) (drop (br_if 0 (i32.const 9) (i32.const 1))) (i64.eqz)
```
rossberg: This drops the remaining argument of the br_if, but still leaves the f32 on the stack. The i64.eqz expects an i64, though. Maybe just change the const into an i64?

test/core/unwind.wast
```diff
@@ -99,6 +121,12 @@
       (f32.const 0) (f64.const 1) (br 0 (i32.const 9)) (i64.eq)
     )
   )
+  ;;(func (export "binary-after-br_if") (result i32)
+  ;;  (block (result i32)
+  ;;    (f32.const 0) (f64.const 1) (drop (drop (br_if 0 (i32.const 9) (i32.const 1))))
```
rossberg: Similarly here.

test/core/unwind.wast
```diff
@@ -117,6 +145,13 @@
       (f32.const 0) (f64.const 1) (i64.const 0) (br 0 (i32.const 9)) (select)
     )
   )
+  ;;(func (export "select-after-br_if") (result i32)
+  ;;  (block (result i32)
+  ;;    (f32.const 0) (f64.const 1) (i64.const 0)
+  ;;    (drop (br_if 0 (i32.const 9) (i32.const 1)))
```
rossberg: And here. The select does not find the right operands on the stack.

test/core/unwind.wast
```diff
@@ -134,6 +169,11 @@
   (func (export "block-value-after-br") (result i32)
     (block (result i32) (f32.const 0) (br 0 (i32.const 9)))
   )
+  ;;(func (export "block-value-after-br_if") (result i32)
+  ;;  (block (result i32)
+  ;;    (f32.const 0) (drop (br_if 0 (i32.const 9) (i32.const 1)))
```
rossberg: Same here, the block expects an i32 but there's an f32.

test/core/unwind.wast
```diff
@@ -149,6 +189,14 @@
   (func (export "loop-value-after-br") (result i32)
     (block (result i32) (loop (result i32) (f32.const 0) (br 1 (i32.const 9))))
   )
+  ;;(func (export "loop-value-after-br_if") (result i32)
+  ;;  (block (result i32)
+  ;;    (loop (result i32)
+  ;;      (f32.const 0) (drop (br_if 1 (i32.const 9) (i32.const 1)))
```
rossberg: Same here for the loop.

Honry:
@rossberg, you untangle my confusion on the difference between `br` and `br_if`, thanks! 
All comments addressed, PTAL.

rossberg:
Hm, I still see a test failure on unwind.wast on Travis.

Honry:
@rossberg, fixed. :)

PR #839, opened 2018-07-26
Indent substeps correctly in the Global constructor.

Ms2ger:
They need four spaces rather than three to escape from the parent list.


binji:
Thanks!

PR #853, opened 2018-08-09
[js-api] Editorial improvements to "create a host function".

Ms2ger:


document/js-api/index.bs
```diff
@@ -886,18 +886,21 @@ Note: [=call an Exported Function|Calling an Exported Function=] executes in the
...
-        1. If the signature contains an [=𝗂𝟨𝟦=] (as argument or result), the host function throws a {{TypeError}} when called.
-        1. Let |arguments| be a [=list=] of the arguments of the invocation of this function.
-        1. Let |jsArguments| be an empty [=list=].
-        1. For each |arg| in |arguments|,
-            1. [=Append=] [=ToJSValue=](|arg|) to |jsArguments|.
-        1. Let |ret| be ? [=Call=](|func|, undefined, |jsArguments|). If an exception is thrown, trigger a WebAssembly trap, and propagate the exception to the enclosing JavaScript.
-        1. Let [<var ignore>parameters</var>] → [|results|] be |functype|.
-        1. If |results| is empty, return undefined.
-        1. Otherwise, return [=ToWebAssemblyValue=](|ret|, |results|[0]).
+        1. Try running the following steps:
```
littledan: I didn't know you were allowed do this in spec-language. Is there another example you can reference with this pattern?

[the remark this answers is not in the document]
Ms2ger: Please explain how the spec makes sense without widening the scope?

Ms2ger: I copied the phrasing from https://heycam.github.io/webidl/#dfn-attribute-getter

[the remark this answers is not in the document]
rossberg: Maybe I'm missing something, but I cannot spot anything that didn't make sense (other than the exception translation being worded in a somewhat casual manner). Can you elaborate? It's always good to explain PRs with a motivational description.

rossberg: It isn't supported by the [Algorithmic Conventions](https://tc39.github.io/ecma262/#sec-algorithm-conventions) of the ES spec, which requires exceptional results to be handled explicitly through completion values and the use of [auxiliary meta operators and conventions](https://tc39.github.io/ecma262/#sec-runtime-semantics) like `?`. It seems like the WebIDL prose does not follow the ES conventions, while the Wasm API tries to, as witnessed by its use of `?`. Is there a strong reason to deviate from these conventions in this case?

littledan: This specification is written with Web-style conventions with a couple ES-isms mixed in where needed, like WebIDL.

littledan:
@rossberg I'm thinking of landing this patch. Do you still have strong reservations about it?

rossberg:
@littledan, yes. The part adding the missing functype argument makes total sense, of course. For the rest, my question about what it's actually supposed to fix still stands. And I prefer not to adopt undefined algorithmic conventions without need.

littledan:
@Ms2ger Would you be open to splitting out the part adding the `|functype|` argument as a separate PR?

Ms2ger:
This no longer uses the try-catch approach.

PR #859, opened 2018-08-17
Add build commands to diff JS+Web documents too

binji:
The diff script was first added in #847, but only had a command to diff
the core spec.


document/Makefile
```diff
@@ -54,6 +57,10 @@ $(DIRS:%=clean-%): clean-%:
 	(cd $(@:clean-%=%); make BUILDDIR=$(BUILDDIR) clean)
 	rm -rf $(BUILDDIR)/$(@:clean-%=%)
 
+.PHONY: $(DIRS:%=diff-%)
+$(DIRS:%=diff-%): diff-%:
```
rossberg: Nit: isn't the `diff-%:` redundant here?

binji: Yes, I wasn't sure what this syntax is, but it looks like they're static pattern rules: https://www.gnu.org/software/make/manual/html_node/Static-Usage.html

It looks like you added these in previous CLs for the other similar rules, why are the needed?

rossberg: Yes, sorry, I was confused. The part before the first colon is a kind of filter. This formulation restricts `diff-%` to a given list of instantiations for `%`, here the known directories. E.g., this rule expands to
```
diff-core diff-js-api diff-web-api: diff-%:
  do stuff generically for %
```
which essentially defines 3 targets with one rule. The alternative would be 3 separate copies of the rule (or one over-generic rule that doesn't actually work in most cases).

PR #867, opened 2018-08-27
[js-api] Fix some inconsistencies around handling bytes and buffers.

Ms2ger:
This changes the specification to consistently grab the bytes immediately upon
receiving a BufferSource from user code, and store bytes directly in the
[[Bytes]] field.

This also clarifies the parsing of custom sections from the module bytes.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -89,7 +89,7 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     url: valid/modules.html#valid-module
         text: valid
         text: WebAssembly module validation
-    text: binary format of a module; url: binary/modules.html
+    text: module grammar; url: binary/modules.html#binary-module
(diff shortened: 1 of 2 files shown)


document/js-api/index.bs
```diff
@@ -295,7 +296,7 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
 </div>
 
 <div algorithm>
-  To <dfn>asynchronously compile a WebAssembly module</dfn> from a {{BufferSource}} |bytes|, using optional [=task source=] |taskSource|, perform the following steps:
+  To <dfn>asynchronously compile a WebAssembly module</dfn> from source bytes |bytes|, using optional [=task source=] |taskSource|, perform the following steps:
```
binji: Perhaps fix the reference to this definition in `web-api/index.bs` too?

Ms2ger: Done.

PR #874, opened 2018-09-13
Use "Element Segment" terminology in js-api tests

binji:
This matches the nomenclature of the spec.

Also:

* Remove the `function_table` array from `wasm-module-builder.js` which
  is not used.
* The `appendToTable` function should reference
  `function_table_length` instead of `function_table.length`.

### DIFF
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -66,7 +66,7 @@ const complexExportingModuleBinary = (() => {
     builder.addMemory(1, 1, /* exported */ false);
     builder.exportMemoryAs('b');
 
-    builder.setFunctionTableLength(1);
+    builder.setTableLength(1);
     builder.addExportOfKind('c', kExternalTable, 0);
 
     // Default init for global values is 0. Keep that.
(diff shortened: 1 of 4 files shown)


littledan:
@Ms2ger could you review this PR?

binji:
Rebased.

binji:
I made these changes outside of v8. I can land in v8 first, though.

gahaas:
@Ms2ger is correct, `wasm-module-builder.js` is just a copy of the V8 version where I removed experimental features.

binji:
I landed the v8 change [here](https://chromium-review.googlesource.com/c/v8/v8/+/1273612), and have updated this copy to be the same (minus future features).

test/js-api/jsapi.js
```diff
@@ -1111,7 +1111,6 @@ const complexTableReExportingModuleBinary = (() => {
             46
         ]).index;
 
-    builder.setFunctionTableLength(3);
```
Ms2ger: Why this change?

binji: It's unnecessary because `appendToTable` will set the correct value for the table length. Previously `appendToTable` would use `function_table.length` to determine the offset of the element segment, but `function_table` was never changed, so it would always use 0. Now `appendToTable` uses `table_length_min`, which would be changed by `setTableLength`. This would make a 6-element table with the first 3 elements `null`.

PR #882, opened 2018-09-26
Fix memory section length

nlordell:
Just a small fix to the `binary.wast` test script. In a couple of assertions, the binary module has the wrong size in the memory sections

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -171,7 +171,7 @@
 (assert_malformed
   (module binary
     "\00asm" "\01\00\00\00"
-    "\05\04\01"                          ;; Memory section with 1 entry
+    "\05\07\01"                          ;; Memory section with 1 entry
     "\00\82\80\80\80\70"                 ;; no max, minimum 2 with unused bits set
   )
   "integer too large"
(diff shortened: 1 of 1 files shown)


cretz:
@nlordell and @binji - My tests no longer fail w/ "integer too large" due to these changes. Can one of y'all confirm which part the "integer too large" is supposed to fail for?

binji:
The minimum and maximum sizes are [u32 values](http://webassembly.github.io/spec/core/bikeshed/index.html#limits%E2%91%A6), so the top 4 bits of the 5th byte of the LEB128 value must be zero. See the [note here](http://webassembly.github.io/spec/core/bikeshed/index.html#binary-int)

cretz:
@binji - OK, thanks, I will have to fix my decoder. Previously I relied on the logic at https://android.googlesource.com/platform/libcore/+/522b917/dex/src/main/java/com/android/dex/Leb128.java#106 which doesn't seem good enough.

binji:
Right, those implementations are OK if you know that you'll never increase the length past 32-bits. But wasm requires the high bits to be zero (or to match the sign for SLEB) so the values can be extended to greater than 32-bits without having the numbers parse differently.

PR #883, opened 2018-10-02
Import jsapi tests from WPT.

Ms2ger:


test/js-api/module/customSections.any.js
```diff
@@ -0,0 +1,167 @@
...
+  const builder = new WasmModuleBuilder();
+  builder.addExplicitSection(binary);
+  const buffer = builder.toBuffer();
+  const module = new WebAssembly.Module(buffer);
+
+  assert_sections(WebAssembly.Module.customSections(module, "name"), []);
+  assert_sections(WebAssembly.Module.customSections(module, "na\uFFFDme"), [
+    bytes,
+  ]);
+  assert_sections(WebAssembly.Module.customSections(module, "na\uDC01me"), [
```
binji: This is a bit surprising to me, since the customSections algorithm says it uses [UTF-8 decode without BOM or fail](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom-or-fail), which doesn't seem to convert the 0xdc01 to a 0xfffd (though I see that elsewhere in the spec). Does the replacement happen somewhere else?

test/js-api/wasm-constants.js
```diff
@@ -0,0 +1,374 @@
...
+let kFunctionSectionCode = 3;  // Function declarations
+let kTableSectionCode = 4;     // Indirect function table and other tables
+let kMemorySectionCode = 5;    // Memory attributes
+let kGlobalSectionCode = 6;    // Global declarations
+let kExportSectionCode = 7;    // Exports
+let kStartSectionCode = 8;     // Start function declaration
+let kElementSectionCode = 9;  // Elements section
+let kCodeSectionCode = 10;      // Function code
+let kDataSectionCode = 11;     // Data segments
+let kNameSectionCode = 12;     // Name section (encoded as string)
```
binji: Remove (unused)

test/js-api/wasm-constants.js
```diff
@@ -0,0 +1,374 @@
...
+  "divide by zero",
+  "divide result unrepresentable",
+  "remainder by zero",
+  "integer result unrepresentable",
+  "invalid function",
+  "function signature mismatch",
+  "invalid index into function table"
+];
+
+function assertTraps(trap, code) {
```
binji: remove (unused?)

test/js-api/wasm-constants.js
```diff
@@ -0,0 +1,374 @@
...
+let kTrapMemOutOfBounds       = 1;
+let kTrapDivByZero            = 2;
+let kTrapDivUnrepresentable   = 3;
+let kTrapRemByZero            = 4;
+let kTrapFloatUnrepresentable = 5;
+let kTrapFuncInvalid          = 6;
+let kTrapFuncSigMismatch      = 7;
+let kTrapInvalidIndex         = 8;
+
+let kTrapMsgs = [
```
binji: Not specified, right?

littledan:
Before landing this patch, please verify that it works with the procedures used in at least one browser Wasm implementation to regularly update from this repository (last time I checked, all four take the tests from here, in a manual pull process).

Ms2ger:
(There will be no need to import from here into SpiderMonkey, as these tests already run there.)

binji:
cc @titzer for trying these tests on v8

test/js-api/module/customSections.any.js
Ms2ger: `customSections` takes a `USVString`, which introduces the replacement character; the "UTF-8 decode without BOM or fail" algorithm is only applied to the bytes in the module, which contain an explicit U+FFFD from the `emit_string` call.

test/js-api/module/customSections.any.js
```diff
@@ -0,0 +1,167 @@
...
+  assert_not_equals(WebAssembly.Module.customSections(module, ""),
+                    WebAssembly.Module.customSections(module, ""));
+}, "Empty module: array caching");
+
+test(() => {
+  const bytes1 = [87, 101, 98, 65, 115, 115, 101, 109, 98, 108, 121];
+  const bytes2 = [74, 83, 65, 80, 73];
+
+  const binary = new Binary;
+  binary.emit_section(kUnknownSectionCode, section => {
```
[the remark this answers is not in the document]
Ms2ger: I'd like to bring the module builder code in sync with the other copy in this repo, but I'd rather not block this PR on that.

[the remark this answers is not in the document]
binji: OK, sgtm

test/js-api/module/customSections.any.js
binji: Thanks, that makes sense.

binji:
I started running these in v8. It seems this uses a number of new functions from the https://web-platform-tests.org/writing-tests/testharness-api.html, so we should probably add these to `README.md`:

`assert_throws`
`format_value`
`assert_class_string`
`setup`
`promise_rejects`

See the WIP changes here: https://chromium-review.googlesource.com/c/v8/v8/+/1277724

Ms2ger:
Thanks! Please let me know if you have any more comments; I'll update the readme some time this week.

binji:
OK, I have these tests running (but not yet passing) in v8. Since they also run in spidermonkey, I think this is good to land. We can address other concerns in future CLs.

FYI @kmiller68 @Cellule

PR #891, opened 2018-10-15
Make jsapi.js test strict-compliant

hammacher:
Tests in V8 should also execute in strict mode. This fixes the jsapi
test for that use case.

### DIFF
--- a/test/js-api/jsapi.js
+++ b/test/js-api/jsapi.js
@@ -16,6 +16,9 @@
 
 const kC0DEFEFE = new Uint8Array([0xC0, 0xDE, 0xFE, 0xFE]);
 
+// Polyfill `globalThis` proposal.
+this.globalThis = this;
+
 (function testJSAPI() {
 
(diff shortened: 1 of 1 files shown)


test/js-api/jsapi.js
```diff
@@ -16,6 +16,9 @@
 
 const kC0DEFEFE = new Uint8Array([0xC0, 0xDE, 0xFE, 0xFE]);
 
+// Polyfill `globalThis` proposal.
+const globalThis = this;
```
mathiasbynens: Nit: `this.globalThis = this` would be a more accurate polyfill. Top-level `const` doesn't create a global.

hammacher: Thanks, done.

binji:
FYI: these tests will be removed when this lands: https://github.com/WebAssembly/spec/pull/883

hammacher:
Ok, it that case we don't need to land this.

PR #900, opened 2018-10-24
[test] Fix building error introduced in #883

Honry:


Honry:
@rossberg, @Ms2ger, PTAL.

PR #901, opened 2018-10-24
[test] Test 'tee_local' and 'set_global' in more operators

Honry:


Honry:
@rossberg, @binji, PTAL. Thanks!

PR #902, opened 2018-10-25
[test] Fixed TODO in exports.wast

Honry:
- Add tests to cover accessing table and memory


rossberg:
I believe that the cases you are adding in this PR are already covered extensively by other tests. The respective todos were referring to something else: accessing internally defined tables/memories *externally*, analogous to the similar tests for for function and global exports that use of "invoke" and "get" actions earlier in this file. However, that would require extending the assertion language with analogous actions on tables and memories, which is why it was left as todos.

Issue #897, opened 2018-10-22
Comments on wasm-module-builder.js


Ms2ger:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1500442, @bnjbvr suggests:

::: testing/web-platform/tests/wasm/jsapi/wasm-module-builder.js
@@ +134,5 @@
> +  }
> +
> +  addLocals(locals, names) {
> +    const old_num_locals = this.getNumLocals();
> +    if (!this.locals) this.locals = []

Wouldn't it be better to have this.locals = [] in the ctor?

@@ +137,5 @@
> +    const old_num_locals = this.getNumLocals();
> +    if (!this.locals) this.locals = []
> +    this.locals.push(locals);
> +    if (names) {
> +      if (!this.local_names) this.local_names = [];

ditto

@@ +243,5 @@
>                         type: type_index});
>      return this.num_imported_funcs++;
>    }
>  
> +  addImportedGlobal(module = "", name, type, mutable = false) {

Uh, mixing optional params with non-optional always makes me a bit nervous...

Maybe @hammacher would like to respond or address the comments in the V8 repo?

PR #903, opened 2018-10-26
[spec] Sync wasm-module-builder.js to v8

hammacher:
This brings the wasm-module-builder.js back in sync with v8.
Note that there is another copy of this in test/js-api, which I plan to
address in a separate CL.

Drive-by: Remove some v8 specific non-spec'ed parts.


hammacher:
Btw, this fixes the comments in issue #897.

xtuc:
Is still this needed? we have #944

hammacher:
Note that this is a different file (test/harness/wasm-module-builder.js, not test/js-api/wasm-module-builder.js).
We should probably just remove that first file.

xtuc:
Ah right, more duplication of this file.

PR #905, opened 2018-10-31
Editorial: Explicitly convert between list and sequence

littledan:
JS/Web specs use mutable lists as a common data structure, whereas
the WebAssembly core specification uses a logical (immutable) sequence
data structure. This patch explicitly converts from a list to a
sequence to reduce ambiguity.


littledan:
cc @rossberg

PR #918, opened 2018-11-23
[test] Move tests in store_retval.wast to relavent test files

Honry:
- Tests in store_retval.wast could be merged into to memory.wast
    and set_local.wast
  - Add similar tests to test load and get_local without retval.


test/core/get_local.wast
```diff
@@ -146,3 +146,21 @@
   "unknown local"
 )
 
+;; get_local should have retval
+(assert_invalid
```
rossberg: Nit: Can you move these tests up to the other "invalid typing" tests?

test/core/set_local.wast
```diff
@@ -255,3 +255,21 @@
   "type mismatch"
 )
 
+;; set_local should have no retval
+(assert_invalid
```
rossberg: Same here.

test/core/get_local.wast
Honry: Sure, done.

Honry:
> Btw, vice versa I wouldn't mind splitting memory.wast into {memory,load,store}.wast, to match the convention of having one file per construct. ;)

@rossberg, I will do this in a follow-up PR. :)

test/core/get_local.wast
```diff
@@ -117,6 +117,26 @@
 )
 
 
+;; get_local should have retval
```
rossberg: I actually meant making the tests part of the "Invalid typing of access to parameters" section above, since they're really just additional variations of similar tests. You could even follow the naming scheme and name the functions i32-vs-empty etc.

Honry: OK, done.

PR #922, opened 2018-11-28
[test] Split memory.wast

Honry:
into {memory,load,store}.wast, to match the convention of having one file per construct.


Honry:
@rossberg, PTAL. Thanks!

PR #934, opened 2018-12-12
[test] Rewrite the embedding limits test.

Ms2ger:
It did not work as written, and review comments were not addressed before
landing. A fair amount of code explicitly marked as "DO NOT COMMIT" was also
committed.

This commit makes the test use the `.any.js` feature that allows integration
with wpt and automated test suited. It also removes the unnecessary
abstraction of the run_* functions, ensures that the promise_test() tests
actually work, removes unactionable TODO comments, removes calls to print()
that break the test in a browser context, and uses the testharness.js library
much more idiomatically. The test now also uses `const` instead of `let` where
appropriate.


binji:
Thanks for cleaning this up!

dtig:
It looks like this rewrite is causing subtests from limits.any.js to fail in Chrome, I've [linked the logs](https://logs.chromium.org/logs/v8/buildbucket/cr-buildbucket.appspot.com/8926706426746075760/+/steps/Check__flakes_/0/logs/limits/0) here. I haven't managed to get this working correctly in the SpiderMonkey shell. @Ms2ger Checking to see if this was tested in Firefox/Chrome?

Ms2ger:
@dtig would you mind filing an issue with more details? I will try to take a look later this week.

PR #939, opened 2019-01-02
[test] Move operand missing tests from typecheck.wast to appropriate files

Honry:
This is the first step to complete TODO tasks in [typecheck.wast](https://github.com/WebAssembly/spec/blob/master/test/core/typecheck.wast),

- ;; TODO: move all tests in this file to appropriate operator-specific files.


Honry:
@rossberg, sure! Sorry for the inconvenient.

Honry:
@rossberg, now this PR is specific for the 1st TODO, and I will add next PR for the 2st TODO once this one is landed. PTAL.

rossberg:
@Honry, thanks!

While browsing through some of these tests, it seems that many of them are actually duplicates of tests that already exist in slight variation in some of those files. It would probably be worthwhile to put a TODO about next to their new location about identifying and removing duplicates.

PR #943, opened 2019-01-16
[test] Remove duplicated tests

Honry:
- br.wast: "$type-operand-missing" and "$type-operand-missing-in-block"
   are similar to "$type-arg-empty-vs-num"
 - return.wast "$type-operand-missing" is similar to "$type-value-empty-vs-num"
 - if.wast: no duplicates

### DIFF
--- a/test/core/br.wast
+++ b/test/core/br.wast
@@ -442,31 +442,6 @@
   "type mismatch"
 )
 
-(assert_invalid
-  (module
-    (func $type-operand-missing-in-if
-      (block
-        (i32.const 0) (i32.const 0)
-        (if (result i32) (then (br 0)))
-      )
-      (i32.eqz) (drop)
-    )
-  )
-  "type mismatch"
-)
-(assert_invalid
-  (module
-    (func $type-operand-missing-in-else
-      (block
-        (i32.const 0) (i32.const 0)
-        (if (result i32) (then (i32.const 0)) (else (br 0)))
-      )
-      (i32.eqz) (drop)
-    )
-  )
-  "type mismatch"
-)
-
 
 (assert_invalid
   (module (func $unbound-label (br 1)))
--- a/test/core/br.wast
+++ b/test/core/br.wast
@@ -444,25 +444,6 @@
 
 (assert_invalid
   (module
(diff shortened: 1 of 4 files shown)


Honry:
@rossberg, PTAL. Thanks!

test/core/br.wast
```diff
@@ -442,25 +442,6 @@
...
-      (i32.const 0)
-      (block (result i32) (br 0))
-      (i32.eqz) (drop)
-    )
-  )
-  "type mismatch"
-)
 (assert_invalid
   (module
     (func $type-operand-missing-in-if
```
rossberg: I believe the this and the next are duplicates  of `$type-{then,else}-break-empty-vs-num` in if.wast, so can be removed as well.

test/core/if.wast
```diff
@@ -768,8 +768,6 @@
   "type mismatch"
 )
 
-;; TODO: Compare above "*operand-missing*" tests to others, identify and remove duplicates.
```
rossberg: Can we rename the above to `$type-condition-empty-*`, to closer match the naming scheme of the other tests?

Also, `in-if` could be `in-then` for symmetry.

test/core/return.wast
```diff
@@ -320,14 +320,6 @@
...
-  (module
-    (func $type-operand-missing (result i32)
-      (return)
-    )
-  )
-  "type mismatch"
-)
 (assert_invalid
   (module
     (func $type-operand-missing-in-block (result i32)
```
rossberg: Can we rename these to $type-value-empty-vs-num-in-*` and move them next to the basic version of the test?

test/core/br.wast
Honry: Indeed, I will remove them as well.

test/core/if.wast
Honry: Yeh, good suggestion, I will rename them soon.

test/core/return.wast
Honry: Fine. :)

Honry:
@rossberg, many thanks for your review, all comments have been addressed. Pls. take a look again. :)

PR #944, opened 2019-01-16
wasm-module-builder: use a loop instead of spreading

xtuc:
Spreading many parameters tends to blow up the stack, whereas a loop is safer.

I ran in this issue with the Wasm limits test.

cc @Ms2ger

### DIFF
--- a/test/js-api/wasm-module-builder.js
+++ b/test/js-api/wasm-module-builder.js
@@ -74,7 +74,9 @@ class Binary extends Array {
     // Emit section length.
     this.emit_u32v(section.length);
     // Copy the temporary buffer.
-    this.push(...section);
+    for (let i = 0, len = section.length; i < len; i++) {
+      this.push(section[i]);
+    }
   }
 }
 


hammacher:
The wasm-module-builder.js is actually a copy from the v8 repository. There, we already fixes this issue.
Did you check for more differences between the two versions? Maybe it makes sense to fully sync to the v8 version?

xtuc:
Yes, that's correct. I'm currently trying to sync them, only the needed changes tho because some are breaking changes.

I think it would be better to have both in sync and I can probably do a followup PR.

binji:
I don't think we should sync the parts that are from the proposal repos here, though.

test/js-api/wasm-module-builder.js
```diff
@@ -148,7 +150,7 @@ class WasmModuleBuilder {
     return this;
   }
 
-  addMemory(min, max, exp) {
+  addMemory(min, max, exp, shared = false) {
```
xtuc: I intentionally kept the `shared` param because it's passed in `limits.any.js` test suite, but it's not emited.

xtuc:
This is now ready for review @hammacher @binji

test/js-api/wasm-module-builder.js
xtuc: Yes, that's a good point. Some changes are needed in the `limits.any.js` test suite so I can remove it here and in another PR its usage.

xtuc:
Good catch @hammacher, I updated the change.

PR #947, opened 2019-01-17
[js-api] use DOMString

### LINKED ISSUE #915: [jsapi] inconsistent utf-8 decoding
Two UTF-8 decoders are specified:
- Wasm's binary format name decoding: https://webassembly.github.io/spec/core/binary/values.html#binary-name.
- [`WebAssembly.Module.customSections`](https://webassembly.github.io/spec/js-api/index.html#dom-module-customsections)'s name decoding (defined in the WHATWG): https://encoding.spec.whatwg.org/#utf-8.

From the WHATWG algorythm, the rules
> 5. Set UTF-8 lower boundary to 0x80 and UTF-8 upper boundary to 0xBF.

> 6. Set UTF-8 code point to (UTF-8 code point << 6) | (byte & 0x3F)

are not used in Wasm.

My understanding is that `U+DC01` and `U+FFFD` should be equal in the JS API, as tested here https://github.com/WebAssembly/spec/blob/5aaea96eceb1e1a3956d7cbb499920e5b8c1109f/test/js-api/module/customSections.any.js#L156-L160 While in Wasm they would be considered as two different sections, which could cause sublte mismatchs.

Note that this is the only occurence of UTF-8 decoding in the JS spec.

xtuc:
Switching from USVString to DOMString will allow the string to contain non normalized surrogates.

Currently, JavaScript engines already use a DOMString as argument.

Fixes #915 

cc @Ms2ger @littledan

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -446,14 +446,14 @@ enum ImportExportKind {
 };
 
 dictionary ModuleExportDescriptor {
-  required DOMString name;
+  required USVString name;
   required ImportExportKind kind;
   // Note: Other fields such as signature may be added in the future.
 };
 
 dictionary ModuleImportDescriptor {
-  required DOMString module;
-  required DOMString name;
+  required USVString module;
+  required USVString name;
   required ImportExportKind kind;
 };
 
@@ -732,7 +732,7 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
 
 <pre class="idl">
 dictionary GlobalDescriptor {
-  required DOMString value;
+  required USVString value;
   boolean mutable = false;
 };
 
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -446,14 +446,14 @@ enum ImportExportKind {
 };
 
 dictionary ModuleExportDescriptor {
-  required USVString name;
+  required DOMString name;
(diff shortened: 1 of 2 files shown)


document/js-api/index.bs
```diff
@@ -446,22 +446,22 @@ enum ImportExportKind {
...
   required ImportExportKind kind;
   // Note: Other fields such as signature may be added in the future.
 };
 
 dictionary ModuleImportDescriptor {
-  required USVString module;
-  required USVString name;
+  required DOMString module;
+  required DOMString name;
   required ImportExportKind kind;
```
Ms2ger: These are only used for values that the JS engine constructs, so might as well keep USVString for clarity.

document/js-api/index.bs
```diff
@@ -732,7 +732,7 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
 
 <pre class="idl">
 dictionary GlobalDescriptor {
-  required USVString value;
+  required DOMString value;
```
Ms2ger: This doesn't make a difference, because it's only compared against strings that don't contain U+FFFD.

It would probably be better as an enum, though.

xtuc:
@Ms2ger I updated the change, I initially though It could avoid any future confusion, but It doesn't seem necessary.

PR #948, opened 2019-01-18
[test] Add more tests for operand missing -Part1

Honry:
This is the first part to fix following task in typecheck.wast:

Split this task into sevaral PRs is for the purpose of easier reviewing.

### DIFF
--- a/test/core/local_set.wast
+++ b/test/core/local_set.wast
@@ -183,6 +183,118 @@
 )
 
 (assert_invalid
+  (module
+    (func $type-param-arg-empty-vs-num (param i32)
+      (local.set 0)
+    )
+  )
+  "type mismatch"
+)
+(assert_invalid
+  (module
+    (func $type-param-arg-empty-vs-num-in-block (param i32)
+      (i32.const 0)
+      (block (local.set 0))
+    )
+  )
+  "type mismatch"
+)
+(assert_invalid
+  (module
+    (func $type-param-arg-empty-vs-num-in-loop (param i32)
+      (i32.const 0)
+      (loop (local.set 0))
+    )
+  )
+  "type mismatch"
+)
+(assert_invalid
+  (module
+    (func $type-param-arg-empty-vs-num-in-then (param i32)
+      (i32.const 0)
+      (if (i32.const 1) (then (local.set 0)))
+    )
+  )
+  "type mismatch"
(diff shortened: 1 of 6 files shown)


Honry:
@rossberg, PTAL. Thanks!

test/core/local_set.wast
```diff
@@ -182,81 +182,17 @@
   "type mismatch"
 )
 
-(assert_invalid
-  (module (func $i32-vs-empty (param i32) (result i32) (local.set 0 (i32.const 1))))
```
rossberg: I just noticed that versions of these tests already existed and were removed in #933. Do you remember why? Can we restore the more explicit naming conventions they had before?

Honry: Um, are you talking about those tests of `i*-vs-empty`? I don't see they were removed in #933, actually they were added in #918 and their names were suggested by you at https://github.com/WebAssembly/spec/pull/918#discussion_r236176421.

Honry: Those `i*-vs-empty` are testing `local.set` should have no retval, and totally different from testing operand missing.  Seems it's a bit confusing on the name with "empty", how about something like "i32-vs-i32"?

Honry:
> Hadn't realised that there were many uses of the "missing" convention already. Sorry for the noise!
So I assume the rest of this PR is just moving tests around?

Hi, @rossberg, thank you for the comments!
I meant adding more tests with same checkpoints for other control constructs and instructions in rest PRs.

test/core/local_set.wast
rossberg: See the first batch of removals from local_set.wast in #933, multiple functions named $type-local-num-vs-num (for whatever reason). Aren't the ones here variations of the same?

Honry: In a sense, `$type-local-num-vs-num` are the same with `i*-vs-empty`, that's why I removed them.
These tests fail for reasons completely unrelated to the type of the param or local, it's because of `local.set` should have no retval.
So, how about rename `i*-vs-empty` as "type-num-vs-num"? Or maybe we can just separate them with a standalone annotation something like ";; local.set should have no retval". WDYT?

Honry: Thanks, I just rename them in new commit. PTAL.

PR #950, opened 2019-01-24
[test] Add more tests for operand missing -Part2

Honry:
This is the second part to fix following task in typecheck.wast:

### DIFF
--- a/test/core/br.wast
+++ b/test/core/br.wast
@@ -442,6 +442,155 @@
   "type mismatch"
 )
 
+(assert_invalid
+  (module
+    (func $type-arg-empty-in-br
+      (i32.const 0)
+      (block (result i32) (br 0 (br 0))) (i32.eqz) (drop)
+    )
+  )
+  "type mismatch"
+)
+(assert_invalid
+  (module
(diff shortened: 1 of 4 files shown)


Honry:
@rossberg, PTAL.

Honry:
@rossberg, thanks for the comment! I agree on you that so much combinations look like meaningless.  
So how about narrow down the scope in 3~4 combinations or smaller?

rossberg:
I'm of course fine with landing more tests, and I appreciate all the effort you are putting into this. But I am worried that the benefit of tests like these may be too marginal to justify spending more of your precious time with them.

On the other hand, there are features for which we have too few tests at the moment, e.g. tables, segments, edge cases of the binary format, etc. Some of them don't even have their own file. Can I perhaps get you interested in shifting focus to those? :)

Honry:
@rossberg, I'm also appreciate your patient and careful reviews. Actually I have already prepared the rest operand missing tests in local, so there won't be too much time to spend on it. :)

After complete all TODOs in typecheck.wast, I will shift my focus to those features with few tests you mentioned.

PR #955, opened 2019-01-25
Remove wasm-{constants,module-builder}.js from harness/.

Ms2ger:
These copies are not used by the jsapi tests, and they're not needed for the
core tests (which is what build.py handles today).


dontcallmedom:

PR #962, opened 2019-02-05
[test] update wasm-module-builder.js

xtuc:
This is for consistency with other WebAssembly tests, in v8 or in wpt.
Edit: also removes legacy methods.

cc @binji @hammacher

### DIFF
--- a/test/js-api/wasm-module-builder.js
+++ b/test/js-api/wasm-module-builder.js
@@ -98,6 +98,12 @@ class WasmFunctionBuilder {
   }
 
   addBody(body) {
+    const bodyCopy = body.slice();
+    bodyCopy.push(kExprEnd);
+    return this.addBodyWithEnd(bodyCopy);
+  }
+
+  addBodyWithEnd(body) {
     this.body = body;
     return this;
   }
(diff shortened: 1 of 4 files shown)


test/js-api/wasm-module-builder.js
```diff
@@ -262,6 +268,11 @@ class WasmModuleBuilder {
     return this.setTableBounds(length);
   }
 
+  // TODO(ssauleau): legacy, remove this
```
hammacher: Can we drop this?

xtuc: Sure, the comment or the method?

hammacher: the method, if possible.

xtuc: Ok, the method is actually not used in this repo but is in wpt (I don't know if v8 does). I think that it would be simpler to later synchronize if we keep this method. Files would stay similar. What do you think?

hammacher: Not sure what exactly you are proposing. V8 does use the method, but that's fine since it tests v8-specific messages. spec or wpt should not use it, and if it's still defined in wpt we should aim to remove it there instead of adding it here (in spec).

xtuc: Right, I agree. I'll remove it here and followup on the wpt side to remove it.

xtuc:
@binji Could you please merge this?

PR #963, opened 2019-02-05
[test] merge wasm-constants.js and wasm-module-builder.js 

### LINKED ISSUE #958: [test] merge wasm-constants.js and wasm-module-builder.js
We merged wasm-constants.js and wasm-module-builder.js files together in v8 (https://chromium-review.googlesource.com/c/v8/v8/+/1446452).

Since those files are copied in this repo, for consistency, I would like to merge them here as well. However, some other projects might depends on one of the files independently.

Are people happy with that change?

cc @binji @hammacher @Ms2ger

xtuc:
Closes #958


test/js-api/wasm-module-builder.js
```diff
@@ -8,6 +8,385 @@ let byte_view = new Int8Array(__buffer);
...
+let kTrapMemOutOfBounds       = 1;
+let kTrapDivByZero            = 2;
+let kTrapDivUnrepresentable   = 3;
+let kTrapRemByZero            = 4;
+let kTrapFloatUnrepresentable = 5;
+let kTrapFuncInvalid          = 6;
+let kTrapFuncSigMismatch      = 7;
+let kTrapInvalidIndex         = 8;
+
+let kTrapMsgs = [
```
hammacher: These messages are not spec'ed. Do we use them anywhere in this repo?

test/js-api/wasm-module-builder.js
```diff
@@ -8,6 +8,385 @@ let byte_view = new Int8Array(__buffer);
...
+function assertTraps(trap, code) {
+  try {
+    if (typeof code === 'function') {
+      code();
+    } else {
+      eval(code);
+    }
+  } catch (e) {
+    assertEquals('object', typeof e);
+    assertEquals(kTrapMsgs[trap], e.message);
```
hammacher: Ah, here it's used. But maybe the whole function is dead. Otherwise I would expect this to fail on VMs other than v8.

test/js-api/wasm-module-builder.js
xtuc: After a quick search, no they aren't. I might have introduce them in my last sync. I'll remove them.

test/js-api/wasm-module-builder.js
xtuc: It seems to be dead code, yes.

xtuc:
@binji as discussed during the WebAssembly GC, could you please merge this one too?

PR #964, opened 2019-02-06
Fix some test function names that got out of sync with the recent instruction renaming

AndrewScheidecker:
Looks like these function names never matched the instruction names exactly, but with the instruction renaming they ended up being confusingly close-but-not-the-same.


AndrewScheidecker:
I linked my W3C account to my GitHub account, so the ipr check should pass now, but I can't manually rerun it.

binji:
> I linked my W3C account to my GitHub account, so the ipr check should pass now, but I can't manually rerun it.

Thanks, it looks like I can sign it to the bot to rerun, but it requests a lot of GH features which I can't really agree to. It's fine though, I know you're a CG member.

PR #971, opened 2019-02-21
[test] Add more tests for operand missing -Part3

Honry:
This is the last part to fix following task in typecheck.wast


Honry:
@rossberg, this is the rest part for operand missing, I've narrowed down the number of combinations, PTAL. Thanks!

PR #973, opened 2019-02-22
[test] Move rest tests in typecheck.wast to appropriate files

Honry:
Partial tests are duplicated with existing tests, so remove them directly.


Honry:
@rossberg, PTAL. Thanks!

Issue #972, opened 2019-02-21
[jsapi] limit tests optimization


xtuc:
The `get_buffer` is slow, the generation of the Wasm module can be > 10min. Of course the execution itself is up to the engine, but here are my suggestions to optimize the test suite a bit:
- Split it in multiple files to allow multithreading
- Cache the buffers (min, limit and limit + 1) instead of generating them for each test.

What do you think?

binji:
Caching the get_buffer results sounds good to me. I'd prefer to keep it one file if we can, but if it's still really slow I'd be OK separating it out. @Ms2ger thoughts?

Ms2ger:
Caching sounds great. It might be interesting to see if any particular subtests are slow, or if they're all equally slow

xtuc:
Ok, I will send a PR with that caching and post my benchmarks here.

xtuc:
By the way, is there any reason the `limits.any.js` testsuite isn't in WPT?

xtuc:
<details><summary>Here are the results (in ms)</summary>
<p>

```
types: get_buffer 1, 4.990000
types: get_buffer 1000000, 33252.488000
types: get_buffer 1000001, 42535.581000
functions: get_buffer 1, 1.442000
functions: get_buffer 1000000, 360336.039000
functions: get_buffer 1000001, 504540.306000
imports: get_buffer 1, 1.248000
imports: get_buffer 100000, 8153.456000
imports: get_buffer 100001, 1802.223000
exports: get_buffer 1, 1.441000
exports: get_buffer 100000, 16557.174000
exports: get_buffer 100001, 24683.390000
globals: get_buffer 1, 1.151000
globals: get_buffer 1000000, 83019.206000
globals: get_buffer 1000001, 160315.990000
data segments: get_buffer 1, 11.270000
data segments: get_buffer 100000, 8152.780000
data segments: get_buffer 100001, 2339.004000
initial declared memory pages: get_buffer 1, 0.432000
initial declared memory pages: get_buffer 65536, 0.089000
initial declared memory pages: get_buffer 65537, 0.051000
maximum declared memory pages: get_buffer 1, 0.128000
maximum declared memory pages: get_buffer 65536, 0.062000
maximum declared memory pages: get_buffer 65537, 0.053000
initial imported memory pages: get_buffer 1, 0.619000
initial imported memory pages: get_buffer 65536, 0.162000
initial imported memory pages: get_buffer 65537, 0.083000
[...]

PR #975, opened 2019-02-25
[jsapi] update limits.any.js

xtuc:
Caches the generated Wasm buffer instead of re-generating
it for each test.

Refs https://github.com/WebAssembly/spec/issues/972


xtuc:
Cc @Ms2ger

Issue #852, opened 2018-08-09
GlobalDescriptor::value is an odd name


Ms2ger:
It can accept the values "i32", "i64", "f32" and "f64". This sounds more like something that should be called "type" or something along those lines.

(Also, it could be a [WebIDL enumeration](https://heycam.github.io/webidl/#idl-enums).)

rossberg:
See [this issue](https://github.com/WebAssembly/mutable-global/issues/16) for context, and the [notes of the May 15 meeting](https://github.com/WebAssembly/meetings/blob/master/2018/CG-05-15.md).

ericprud:
Does the CG decision referenced in [binji's comment](https://github.com/WebAssembly/mutable-global/issues/16#issuecomment-389234666) close this issue?

binji:
Yes, closing.

xtuc:
We discussed about the name of the field.

I would be in favor of switch it to a enumeration ('i32', 'i64', 'f32', or 'f64') of value types (as @rossberg mentioned). I would be happy to bring that up in the next meeting.

PR #976, opened 2019-03-04
Normative: GlobalDescriptor shoud use DOMString

xtuc:
Following https://github.com/WebAssembly/spec/issues/915.

There's no reasons (I'm aware of) that the `value` property in the GlobalDescriptor use a `USVString` String. As for Module.customSection let's switch it to `DOMString`.

cc @Ms2ger

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -732,7 +732,7 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
 
 <pre class="idl">
 dictionary GlobalDescriptor {
-  required USVString value;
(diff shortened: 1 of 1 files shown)


Ms2ger:
It doesn't matter, so I prefer keeping the reasonable default, which would be USVString. If you want to change it, why not change to an enum?

xtuc:
Yes I agree (ref https://github.com/WebAssembly/spec/issues/852#issuecomment-469304122).

littledan:
OK, let's close this then in favor of switching to an enumeration.

PR #977, opened 2019-03-06
Normative: declare ValtypeEnum and update GlobalDescriptor

xtuc:


document/js-api/index.bs
```diff
@@ -730,8 +730,14 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
 <h3 id="globals">Globals</h3>
 
 <pre class="idl">
+enum ValtypeEnum {
```
xtuc: @littledan I'm wondering if we should either hoist it in the spec to allow it to be re-used, but there's no use-case yet. Or I would like to scope it here. Does an inline enum in the GlobalDescriptor struct work or should I change its name?

littledan: WebIDL doesn't really have scoping; you definitely can't put the enum inside the dictionary. If you want to avoid overlaps, you could prefix the name with WebAssembly.

I'd suggest avoiding using an abbreviated name here. Also, I don't think you need to include Enum in the name of an enum.

Putting that all together, you could use WebAssemblyValueType as the name.

xtuc: Ok, sounds good to me; thanks.

rossberg:
The direction of introducing a dedicated type seems right. Cf. the more general [JS type reflection proposal](https://github.com/WebAssembly/js-types/blob/master/proposals/js-types/Overview.md) which we will want to work towards.

Beyond that we need to keep in mind ongoing proposals extending the type structure. In particular, with reference types and follow-ups, ValueType will no longer be a simple enum, but a more general union(*). My understanding of WebIDL conventions is limited; does defining it as an enum induce any sort of guarantees that these later additions would break? (Such as a closed set of string-like values?)

As for naming, it seems a bit inconsistent to prefix ValueType with WebAssembly but not the other interface types?

(*) For perspective, with e.g. the reference and function types proposals we will have the equivalent of
```
ValueType = NumType | RefType
NumType = "i32" | "i64" | "f32" | "f64"
RefType = "anyref" | "funcref" | {referee: DefType}
DefType = FuncType  // more options with the GC proposal
```

Ms2ger:
I don't quite understand your syntax. Does it mean that the value would either be a string (from a static set of values), or an object with a single "referee" field? Adding support for the object will need a change to the type in the spec regardless of whether we use an enum or a more general string type. (This should not be a problem, though.) As long as the set of possible string values will remain static, I think using an enum makes sense.

As far as observability goes, I don't think this change is actually observable. The only thing that happens in the current spec between the processing of the "value" field on the IDL side and the ToValueType() call in the prose is reading the "v" argument, but because that's typed as `any`, that's not observable.

rossberg:
@Ms2ger:
> Does it mean that the value would either be a string (from a static set of values), or an object with a single "referee" field?

Yes, and we may have to add more (optional) fields or additional objects in the future.

> As long as the set of possible string values will remain static, I think using an enum makes sense.

We will soon have API that *returns* values of this type (see the JS type reflection proposal). Reading the WebIDL def it says vaguely that an enum defines a closed set of string values. If I took that literally then adding new values later would technically become a breaking change to the API, because clients would receive values they weren't contracted for. But perhaps we can handwave around this by adding a note that future version of the API may add additional values?

Ms2ger:
> We will soon have API that returns values of this type (see the JS type reflection proposal). Reading the WebIDL def it says vaguely that an enum defines a closed set of string values. If I took that literally then adding new values later would technically become a breaking change to the API, because clients would receive values they weren't contracted for. But perhaps we can handwave around this by adding a note that future version of the API may add additional values?

Sure; the API contract really only is the actual JS-visible behaviour, not so much the IDL syntax used to express it. I'd consider it as much a breaking change if the new values were defined in prose only.

xtuc:
@rossberg I understand that some proposal might conflict with that definition of the Valtypes; I would like to scope it locally and duplicated it in other defintion if needed.

I believe we need to dupplicate the WebAssembly ValueType to a WebIDL anyway because of the WebIDL generators or for correctness.

What do you think if I renamed the enum to `GlobalDescriptorValtype` or similar?

littledan:
I don't think this type has anything to do with global descriptors in particular. It's about types of WebAssembly values. So I don't understand the motivation for putting Global in its name.

rossberg:
Indeed, introducing multiple incompatible copies of a definition of value type would be counter productive. Part of the goal of the JS type reflection proposal was to reconcile the current ad-hoc notions of descriptors with the actual Wasm type system, in a way that allows type information to round-trip seamlessly.

xtuc:
Ok, sorry if I misunderstood;

Following @rossberg's comment, we would have a "global" ValueType like this currently:

```
ValueType = NumType
NumType = "i32" | "i64" | "f32" | "f64"
```

Would it make sense to translate it into WebIDL notation and write it in the spec?
It will allow for extensions in the proposal specificiations.

littledan:
Making it an enum for now does this; if you later split out the enum and redefine WebAssemblyValueType to a union type, then that should do the trick.

Are folks here comfortable with merging this patch, then?

rossberg:
Can we add a big fat note clarifying that this enum may be extended in future versions of Wasm and hence additional values may occur?

xtuc:
@rossberg I added

> Note: the WebAssemblyValueType enum is subject to future changes; several incomming WebAssembly GC proposals will add new values or extend the types. 

Could you please confirm that signals your concern.

document/js-api/index.bs
```diff
@@ -729,9 +729,20 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
...
+<pre class="idl">
+enum WebAssemblyValueType {
+  "i32",
+  "i64",
+  "f32",
+  "f64"
+};
+</pre>
+
+Note: the WebAssemblyValueType enum is subject to future changes; several incomming WebAssembly GC proposals will add new values or extend the types.
```
rossberg: Nit: it's not just the GC proposal. In general, I wouldn't mention proposals in the spec. How about:

"Note: This type may be extended with additional cases in future versions of WebAssembly. Client code should not assume that no other constants can occur."

document/js-api/index.bs
```diff
@@ -729,9 +729,20 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
 
 <h3 id="globals">Globals</h3>
 
+<pre class="idl">
+enum WebAssemblyValueType {
```
rossberg: I'm still confused why we prefix (only) this IDL type with "WebAssembly", that seems inconsistent. Perhaps remove?

xtuc: Good point, I forgot to ask if we should rather remove it.

xtuc:
I'm not familiar with WebIDL to understand if we should change the current definition of `valuetype` too, see https://github.com/WebAssembly/spec/blob/73deada595949e8e9c34f5a7b425db72acd29178/document/js-api/index.bs#L138-L142

I think it would make sense to refer to the new `ValType` enum?

rossberg:
@xtuc:
> I think it would make sense to refer to the new ValType enum?

AFAIU, this is still needed to connect the WebIDL enum to the corresponding definition from the Wasm spec.

document/js-api/index.bs
```diff
@@ -736,6 +736,11 @@ enum WebAssemblyValueType {
   "f32",
   "f64"
 };
+</pre>
+
+Note: the WebAssemblyValueType enum is subject to future changes; several incomming WebAssembly GC proposals will add new values or extend the types.
```
littledan: I'd remove the word GC here; otherwise LGTM.

xtuc: Sorry Dan you are looking at an outdated version.

Currently we have:
> Note: this type may be extended with additional cases in future versions of WebAssembly. Client code should not assume that no other constants can occur.

document/js-api/index.bs
```diff
@@ -729,9 +729,20 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
...
+<pre class="idl">
+enum ValueType {
+  "i32",
+  "i64",
+  "f32",
+  "f64"
+};
+</pre>
+
+Note: this type may be extended with additional cases in future versions of WebAssembly. Client code should not assume that no other constants can occur.
```
Ms2ger: "should" is an RFC2119 term, so must not be used in non-normative text, including notes.

xtuc: Good catch, can I rephrase it to "We advise client code not to assume...."?

littledan: Let's just delete this sentence.

xtuc: Before I do the change, cc @rossberg do we agree to delete the last sentence?

rossberg: If you feel that the warning is sufficiently clear without it and can guarantee that the enum will not become a "breaking change" trap for us later then I'm fine.

xtuc:
@littledan @Ms2ger I addressed your comments, PTAL.

PR #981, opened 2019-03-11
Fix store handling in 'instantiate a WebAssembly module'.

Ms2ger:
The store variable was not defined before it was used, and the result was not
saved.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -364,11 +364,14 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
             2. Let |tableaddr| be |v|.\[[Table]]
             1. Let |externtable| be the [=external value=] [=external value|𝗍𝖺𝖻𝗅𝖾=] |tableaddr|.
             1. [=Append=] |externtable| to |imports|.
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -364,7 +364,9 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
             2. Let |tableaddr| be |v|.\[[Table]]
             1. Let |externtable| be the [=external value=] [=external value|𝗍𝖺𝖻𝗅𝖾=] |tableaddr|.
             1. [=Append=] |externtable| to |imports|.
+    1. Let |store| be the [=surrounding agent=]'s [=associated store=].
     1. Let (|store|, |instance|) be [=instantiate_module=](|store|, |module|, |imports|).
+    1. Set the [=surrounding agent=]'s [=associated store=] to |store|.
```
binji: This should happen after checking whether it is an error. Also, the [embedding API](http://webassembly.github.io/spec/core/appendix/embedding.html#embed-instantiate-module) says that `instantiate_module` returns a pair of `(store, instance)`, or `error`, so we should probably rewrite this to destructure it after checking for `error` instead.

Ms2ger: Done.

PR #982, opened 2019-03-12
[jsapi] Fix tests after wasm-module-builder.js update

hammacher:
Updating the wasm-module-builder.js broke the jsapi tests. I was under
the impression that travis would run them, but it seems like this is
not the case. Rolling the updated version to v8 failed then:
https://crrev.com/c/1516684

This fixes two things:
1) {toBuffer} returns a Uint8Array as before, not an ArrayBuffer,
2) {addExplicitSection} expects a byte sequence, hence pass the
   truncated buffer instead of the {Binary} object.

### DIFF
--- a/test/js-api/module/customSections.any.js
+++ b/test/js-api/module/customSections.any.js
@@ -96,7 +96,7 @@ test(() => {
   });
 
   const builder = new WasmModuleBuilder();
-  builder.addExplicitSection(binary);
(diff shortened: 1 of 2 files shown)


hammacher:
Landing right away to unblock the v8 DEPS roll (https://crbug.com/940854).

binji:
lgtm

PR #995, opened 2019-03-30
Update instructions.rst

zoren:
In note on folded instructions use `local.get` name instead of `get_local` which isn't used elsewhere.


binji:

binji:
Thanks!

PR #998, opened 2019-04-09
[test] Test alignment and offset with too long integer

Honry:


Honry:
@rossberg, I'm going to add some edge cases of the binary format, this is the first one, PTAL. Thanks!

PR #999, opened 2019-04-11
[test] Synchronize js-api test changes from wpt.

Ms2ger:


binji:
Some of these changes (`memory/*`) are from the threads proposal. We should probably avoid syncing those for now.

rossberg:
This PR is stale, can we close it?

Ms2ger:
Yeah, not particularly useful as-is. We should probably discuss the best approach at some point.

PR #1002, opened 2019-04-16
[spec] Use more precise Unicode terminology

### LINKED ISSUE #1000: Use Unicode scalar values instead of Unicode code points
I've noticed that the spec does not allow surrogate code points in names which means the technical definition of `codepoint` is incorrect in terms of the Unicode standard.

[Unicode code point](https://www.unicode.org/glossary/index.html#code_point) is any value in the range [0, 10FFFF].
[Unicode scalar value](https://www.unicode.org/glossary/index.html#unicode_scalar_value) is any Unicode code point except high-surrogate and low-surrogate code points.

So please, rename `codepoint` to something like `scalar_value` or `unicode_scalar_value`.

rossberg:
Say "scalar value" instead of "code point" (renaming grammar symbol to `char`). Fixes #1000.

### DIFF
--- a/document/core/appendix/embedding.rst
+++ b/document/core/appendix/embedding.rst
@@ -82,10 +82,10 @@ Modules
 .. index:: text format
 .. _embed-parse-module:
 
-:math:`\F{parse\_module}(\codepoint^\ast) : \module ~|~ \error`
+:math:`\F{parse\_module}(\char^\ast) : \module ~|~ \error`
 ...............................................................
 
-1. If there exists a derivation for the :ref:`source <text-source>` :math:`\codepoint^\ast` as a :math:`\Tmodule` according to the :ref:`text grammar for modules <text-module>`, yielding a :ref:`module <syntax-module>` :math:`m`, then return :math:`m`.
+1. If there exists a derivation for the :ref:`source <text-source>` :math:`\char^\ast` as a :math:`\Tmodule` according to the :ref:`text grammar for modules <text-module>`, yielding a :ref:`module <syntax-module>` :math:`m`, then return :math:`m`.
 
 2. Else, return :math:`\ERROR`.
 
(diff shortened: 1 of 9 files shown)


KronicDeth:
Is `char` used for scalar value outside of Rust?  When I first started using Rust it confused me as I'm more used to `char` being a byte and not whatever size type fits a unicode scalar value.

rossberg:
@KronicDeth, AFAICT, it's rare for modern languages to still equate char with byte. To give just one other example, Haskell already made char a Unicode character back in 1996 (still 16 bit at the time).

Lyberta:
A nitpick but important one. The name [character](https://www.unicode.org/glossary/index.html#character) is really ambiguous in Unicode and equating "character" and "Unicode scalar value" becomes moot in the presence or combining marks, emoji and other things that are composed of many scalar values yet look as a single character to the user.

The consensus of C++ Unicode study group is that "user perceived character" is an [extended grapheme cluster](https://www.unicode.org/glossary/index.html#extended_grapheme_cluster) and will be the default unit of text for the new Unicode API while code unit and scalar value access is intended for experts.

ericprud:
I think the strategy of this PR is to s/codepoint/char/g to provide an intuitive term and then tie `char` back to Unicode `scalar value` with text like

> A name string must form a valid UTF-8 encoding as defined by |Unicode|_ (Section 2.5) and is interpreted as a string of Unicode scalar values.

In principle, UTF-8 already tells us this. I think this is saying that the interpretation stops there.

I'll write to the i18n list and ask for an elegant way to say "we don't screw with it", which means there's no [normalization](https://unicode.org/reports/tr15/#Norm_Forms), no case-folding of domain names, etc.

binji:
I think it's reasonable to use `char` here too, though I can sense that there may be some confusion. On the other hand, `scalar value` will be easily confused too. If we decide we don't like `char`, maybe the initialism `usv` (for unicode scalar value) instead?

rossberg:
Okay, I'm landing for now. If we think there is pressing need for a rename, we can do so in a follow-up.

PR #1007, opened 2019-04-29
[test] Add more tests for unsigned LEB128 with non-minimal length

Honry:
tweak some tests order a bit to make it looks beauty passingly.

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -33,7 +33,7 @@
 ;; Leading UTF-8 BOM.
 (assert_malformed (module binary "\ef\bb\bf\00asm\01\00\00\00") "magic header not detected")
 
-;; Unknown binary version.
+;; Malformed binary version.
 (assert_malformed (module binary "\00asm") "unexpected end")
 (assert_malformed (module binary "\00asm\01") "unexpected end")
 (assert_malformed (module binary "\00asm\01\00\00") "unexpected end")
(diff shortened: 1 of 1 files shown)


test/core/binary.wast
```diff
@@ -6,6 +6,9 @@
 (assert_malformed (module binary "") "unexpected end")
 (assert_malformed (module binary "\01") "unexpected end")
 (assert_malformed (module binary "\00as") "unexpected end")
+(assert_malformed (module binary "\00asm") "unexpected end")
+(assert_malformed (module binary "\00asm\01") "unexpected end")
+(assert_malformed (module binary "\00asm\01\00\00") "unexpected end")
```
rossberg: I'd prefer to keep this where they were before. That was grouping all tests concerned with the version.

Honry: Sure, done. :)

test/core/binary.wast
```diff
@@ -37,6 +34,9 @@
 (assert_malformed (module binary "\ef\bb\bf\00asm\01\00\00\00") "magic header not detected")
 
 ;; Unknown binary version.
```
rossberg: Nit: can change this heading to "malformed binary version"?

Honry: Yeah, it‘s more explicit, thanks!

PR #1012, opened 2019-05-06
[spec] Constrain name section ordering

### LINKED ISSUE #1010: [spec] name section: duplicate subsections allowed
The [current spec for the `name` section](https://webassembly.github.io/spec/core/appendix/custom.html#name-section) seems to allow multiple subsections within the same `name` section.

This seems to be an oversight, given that the spec is clear that there is at most one `name` section, and that within each name map, keys have to be unique, so it seems odd that one would be allowed to have, say, multiple module name subsections, or multiple function name subsections with non-disjoint keys.

I suggest to add the sentence “Each kind of subsection should appear at most once” or something similar to that end to the spec. Or, if duplicate subsections are intended, clarify their semantics (first wins, or last wins, and how to merge name maps).

rossberg:
Fixes #1010.

### DIFF
--- a/document/core/appendix/custom.rst
+++ b/document/core/appendix/custom.rst
@@ -43,7 +43,10 @@ Each subsection consists of a
    \production{name section} & \Bnamesec &::=&
      \Bsection_0(\Bnamedata) \\
    \production{name data} & \Bnamedata &::=&
-     n{:}\Bname~~\Bnamesubsection^\ast & (\iff n = \text{name}) \\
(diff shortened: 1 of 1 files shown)


rossberg:
Ah yeah, sorry, should have mentioned that this matches the design doc.

Btw, I keep seeing Bikeshed failures on Travis that seem unrelated. Any idea?

binji:
Yeah, I've noticed that too. Not sure what's up with it, but I'll take a look.

binji:
Fix here: https://github.com/WebAssembly/spec/pull/1014

PR #1016, opened 2019-05-13
[test] More tests for overlong LEB128

Honry:


rossberg:
If you plan to include more of these tests, what would you think about splitting them off into a separate `binary-leb128` file?

Honry:
I agree to have a standalone file to cover all LEB128 tests. :)  And will do this in a follow-up with more LEB128 tests.

PR #1018, opened 2019-05-14
[spec] Replace URLs with bikeshed biblio refs

binji:
This will automatically add these entries to the normative references
section which is auto-generated at the bottom of the document.


binji:
This seems to work, but it produces the following entry:

**[UNICODE]**
The Unicode Standard. URL: https://www.unicode.org/versions/latest/

Do we need to reference something else?

binji:
Talking to @ericprud offline, it sounds like this is OK for now. We may need to fix in the future (upstream in bikeshed?)

PR #1020, opened 2019-05-17
[spec][js-api] Fix some links

binji:
Replace `init_store` with `store_init`.

Also, reword the "Security and Privacy Considerations" section, and fix
the references are using ReST syntax (which doesn't work in bikeshed).


binji:
Gonna land now, respond with any wording changes here and I'll fix in another PR.

PR #1021, opened 2019-05-18
[interpreter] Fix edge cases for f32_convert_i64

### LINKED ISSUE #421: [interpreter] f32.convert_u/i64 and f32_convert_s/i64 have incorrect rounding
The implementation of f32.convert_u/i64 has incorrect rounding in some cases. For example, these tests should pass but currently fail:

```
(assert_return (invoke "f32.convert_u_i64" (i64.const 0x7fffff4000000001)) (f32.const 0x1.fffffep+62))
(assert_return (invoke "f32.convert_u_i64" (i64.const 0x8000008000000001)) (f32.const 0x1.000002p+63))
```

A fix analogous to https://github.com/WebAssembly/spec/pull/420 is not sufficient, because OCaml lacks 32-bit float types, so it would round to 64-bit float first, then round to 32-bit float in a separate step, leading to incorrect rounding. For example, the above tests still fail.

rossberg:
Fixes #421.

### DIFF
--- a/interpreter/exec/f32.ml
+++ b/interpreter/exec/f32.ml
@@ -6,8 +6,8 @@
 include Float.Make
   (struct
     include Int32
-    let pos_nan = 0x7fc00000l
-    let neg_nan = 0xffc00000l
-    let bare_nan = 0x7f800000l
+    let pos_nan = 0x7fc0_0000l
+    let neg_nan = 0xffc0_0000l
+    let bare_nan = 0x7f80_0000l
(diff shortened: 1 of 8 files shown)


rossberg:
I'm confident that it works for all cases, since it avoids any rounding while going to f64, while the accumulated bits are always rounded away when going to f32 afterwards. The specific shift value doesn't even matter, anything between 11 and 39 would work, I think, as long as it lands somewhere between f64 and f32 precision.

Issue #728, opened 2018-02-28
[spec review] 4 - Execution


flagxor:
Detail oriented review of draft text:
http://webassembly.github.io/spec/core/bikeshed/index.html#execution%E2%91%A1

Try to be ready to discuss issues by the next in person WG meeting.

eholk:
If no one's done this by the time I'm done with the Appendix, I'd be willing to take it.

binji:
Feedback: https://gist.github.com/binji/57b6b4ff318c6dc77f6d2e1fa7b2bada

inline here:

4.2.12. Refers to the stack containing three kinds of entries "values", "labels", and "activations". Section 4.2.12.1 is called "Values", section 4.2.12.2 is called "Labels" but section 4.2.12.3 is called "Frames" instead.

4.2.12.3. The return arity is named but not labeled here. "Activation frames carry the return arity of the respective function". It should be the same as label above where `n` is mentioned explicitly.

4.2.13. "ultimately reducing the entire program to a single trap instruction", this isn't true when calling out to the embedder, right?

4.2.13. Note #3. The link for the label reduction rule is broken, it should go to 4.4.6.2. Also, the rule is different in the note.

4.2.13.2. What does "home" mean in "home module instance"?

4.2.13.1. Better explanation of "the hole". It's not clear (to me) what it represents and how it works.

4.2.13.3. What is the purpose of the second reduction rule? (Where a `frame` instruction on the stack is modified)

4.3. Mention the meta-variable "z"?

4.3. trunc is defined slightly different here than below in ftrunc_N: there `q` is explicitly annotated with a `+`.

4.3.1.3. Is this bigendian instead? It looks like it extracts the most significant bits and puts it first in the byte sequence.

4.3.2.14 - 4.3.2.17. All define the shift amount as `j_2`, but use `k` in the formula.

4.3.3.1. sp: "correspondence"

4.3.3.3 and others. it's a bit odd that the prose says `nans_N{z_1, z_2}` where the math says `nans_N{}` for the same thing.

4.3.3.4. Is there meant to be a difference between fadd and fsub here? fsub has `fsub_N(\pm 0, \pm q_2) = \mp q_2` and fadd has the same thing but with z instead of q.

4.3.4.4. `trunc` is not defined for `-0`, does it need a special case?

4.3.4.6. Any "arithmetic" NaN of size N

[...]

PR #1022, opened 2019-05-20
[spec] Address feedback on section 4

rossberg:
Addresses lost feedback from #728. Also, adds a bit of frontmatter to multi-page version to match relevant information from the W3C version.

### DIFF
--- a/document/core/util/macros.def
+++ b/document/core/util/macros.def
@@ -3,6 +3,12 @@
 .. External Standards
 .. ------------------
 
+.. |WasmDraft| replace:: http://webassembly.github.io/spec/core/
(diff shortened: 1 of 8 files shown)


document/core/util/macros.def
```diff
@@ -3,6 +3,12 @@
 .. External Standards
 .. ------------------
 
+.. |WasmDraft| replace:: http://webassembly.github.io/spec/core/
+.. _WasmDraft: http://webassembly.github.io/spec/core/
+
+.. |WasmIssues| replace:: http://github.com/webassembly/spec/
```
binji: https://github.com/WebAssembly/spec/issues?

rossberg: Done.

PR #1024, opened 2019-05-22
Rewrite definition of error types.

### LINKED ISSUE #810: js api: improve definition of error objects
There's `[LegacyNamespace=WebAssembly] interface CompileError { };` and then a bunch of prose requirements on the objects that are constructed because of that IDL. AFAICT the only thing that the IDL is supposed to do is define that the namespace object has properties called `FooError`. Let's just do that manually and avoid the conflict?

Ms2ger:
This at least gets somewhat close to defining what's actually intended.

Fixes #810.


Ms2ger:
Who can merge this?

PR #1025, opened 2019-05-27
[test/interpreter] Rounding edge cases for float text literals

### LINKED ISSUE #809: [interpreter] rounding for floating-point constants
The interpreter currently rejects `(f32.const 0x1.fffffefffffff8000000p127)`, because it rounds the value up, out of the finite range of `f32`. Ideally, floating-point literals should be interpreter as if they were floating-point values with infinite precision, and then rounded in a single step to the expression type.

See also the background discussion [here](https://github.com/WebAssembly/spec/pull/807#discussion_r190149497).

rossberg:
Fixes #809 and other double rounding edge cases.

Adds various tests for rounding on f32 and f64 literals.

Adds a gross hack to the interpreter to detect and correct double rounding without having to reimplement float parsing.


test/core/const.wast
```diff
@@ -135,3 +141,300 @@
...
+
+
+;; Rounding behaviour
+
+(module
+  (func (export "f32") (param f32) (result f32) (local.get 0))
+  (func (export "f64") (param f64) (result f64) (local.get 0))
+)
+
+;; f32, small exponent
```
binji: We should probably have some decimal test cases here too.

rossberg: Done.

test/core/const.wast
```diff
@@ -135,3 +141,300 @@
   (module quote "(func (f64.const -269653970229347356221791135597556535197105851288767494898376215204735891170042808140884337949150317257310688430271573696351481990334196274152701320055306275479074865864826923114368235135583993416113802762682700913456874855354834422248712838998185022412196739306217084753107265771378949821875606039276187287552) drop)")
   "constant out of range"
 )
+
+;; NaNs
```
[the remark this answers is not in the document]
rossberg: Right, removed.

PR #1027, opened 2019-05-28
Editorial: Remove links from Number, Object when checking types

### LINKED ISSUE #1026: [js-api] ref to css:Number rather than tc39:Number
[Step 4.5 of the WebAssembly module instantiation algorithm](https://www.w3.org/TR/wasm-js-api/#ref-for-sec-ecmascript-data-types-and-values%E2%91%A0) references the CSS definition of Number. Should that be a reference [20.1.1The Number Constructor](https://tc39.github.io/ecma262/#sec-number-constructor)?

littledan:
This follows the convention followed by the JavaScript spec and HTML.

Fixes #1026


ericprud:
do we care to make those references to ecmascript or is it fine just to leave them as barewords?

littledan:
There's no good single thing to link to, since ES and HTML don't link to something for these terms themselves.

ericprud:
fair enough -- LGTM

PR #1034, opened 2019-06-08
Fix Incorrect Binary Lengths, Invalid Function Returns, and Invalid Types

blackhole12:
Several tests that are intended to throw an error are accidentally testing multiple issues. This is problematic, because which error comes up first can change based on the implementation. In particular, the reference WebAssembly interpreter ignores binary lengths on known sections, which allows many incorrect length values to slip through. This pull request fixes the following things:

- Incorrect binary section lengths
- Incorrect value stack state when returning from a function call
- Two cases of "alignment" that should instead be checking against "alignment must not be larger than natural" to be consistent with the other tests in `align.wast`
- Three cases of incorrect constant types.

The unit tests should be testing exactly one thing, and should not have a second potential error masked by the error being tested. The errors uncovered here are all a result of running the tests through [inNative](https://github.com/innative-sdk/innative), which can be used to check tests against a secondary implementation before committing them.


binji:
We probably don't want to update katex as part of this PR. I suppose it's too late to fix that now, however.

rossberg:
Oops, sorry, I didn't even notice that. It's easy to miss in the GH diff view.

PR #1043, opened 2019-07-12
[interpreter] Fix float literals

gahaas:
I cannot compile the spec interpreter with ToT. This PR fixes the problem for me. I tried to compile with ocaml 4.02.3, 4.06.0, and 4.08.0, by just running `make`.

### DIFF
--- a/interpreter/exec/f32_convert.ml
+++ b/interpreter/exec/f32_convert.ml
@@ -34,14 +34,14 @@ let convert_i64_s x =
   F32.of_float Int64.(
     if abs x < 0x10_0000_0000_0000L then to_float x else
     let r = if logand x 0xfffL = 0L then 0L else 1L in
-    to_float (logor (shift_right x 12) r) *. 0x1p12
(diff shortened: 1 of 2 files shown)


KronicDeth:
It is unfortunate that the original does not work.  It is much clearer you are using a max value.  Could we put the values into named constants to make it clearer that they are max for a given bit width?

rossberg:
Yeah, hex float constants were introduced with Ocaml 4.03, which is from 3+ years ago. Do you still need support for 4.02? Current is 4.08.

gahaas:
I think I forgot to do `make clean`, that's why I was not able to compile it, even with 4.08. The instructions in interpreter/README.md still mention 4.02. Could you adjust it to the version that is actually needed?

rossberg:
Ah, wanted to say that after reconsideration, cutting backwards compatibility does not seem worth it (yet), and we should land this after all. Too late though, so I created #1044, PTAL.

PR #1047, opened 2019-07-20
[spec] Fix header level

rossberg:


binji:
lgtm, I'll merge to wg_v1 branch too.

PR #1049, opened 2019-07-29
[test] Fix error in build.py when test compilation fails

eqrion:
convert_wast_to_js uses multiprocessing to parallelize compilation of
wast test cases. When an error happens, the logging code expects the
result to be a process exit code and a CompletedProcess instance.

This isn't possible, so this commit updates the result to be a
CompletedProcess and modifies the code to manually check the exit code.

### DIFF
--- a/test/build.py
+++ b/test/build.py
@@ -20,6 +20,12 @@ WPT_URL_PREFIX = '/resources'
 
 # Helpers.
 def run(*cmd):
+    return subprocess.run(cmd,
+                          text=True,
+                          stdout=subprocess.PIPE,
+                          stderr=subprocess.STDOUT,
+                          universal_newlines=True)
+def call(*cmd):
     return subprocess.call(cmd,
                            stdout=subprocess.PIPE,
                            stderr=subprocess.STDOUT,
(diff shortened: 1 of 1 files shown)


test/build.py
```diff
@@ -73,8 +77,8 @@ def convert_wast_to_js(out_js_dir):
 
     pool = mp.Pool(processes=8)
     for result in pool.imap_unordered(convert_one_wast_file, inputs):
-        if result != 0:
-            print('Error when compiling {} to JS: {}', wast_file, result.stdout)
+        if result.returncode != 0:
+            print('Error when compiling to JS:\n{}\n{}'.format(result.args, result.stdout))
```
bnjbvr: Would it make sense to print `result.stderr` if it's non empty too?

eqrion: Yeah, that would be good. I'll update.

eqrion: I've updated `run` to merge stderr and stdout, which should fix this.

eqrion:

eqrion:
It looks like travis ran out of log space, so it's maybe an intermittent issue? Is there a way to retrigger it?

rossberg:
Yes, there is a "Restart build" button on the Travis page you get to by clicking on "Details".

eqrion:
Okay, couldn't find a 'Restart Build' option. I'm guessing this is due to not having enough privilege. I did a force-push and that did the trick. Should be ready to merge.

PR #1053, opened 2019-08-13
[spec] Terminology nits

rossberg:
Properly use "instruction" not "operator" in a couple of places.

### DIFF
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -129,9 +129,9 @@ For each type, several subcategories can be distinguished:
 
 * *Constants*: return a static constant.
 
-* *Unary Operators*: consume one operand and produce one result of the respective type.
+* *Unary Operations*: consume one operand and produce one result of the respective type.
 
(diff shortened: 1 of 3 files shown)


document/core/syntax/instructions.rst
```diff
@@ -129,9 +129,9 @@ For each type, several subcategories can be distinguished:
 
 * *Constants*: return a static constant.
 
-* *Unary Operators*: consume one operand and produce one result of the respective type.
+* *Unary Operations*: consume one operand and produce one result of the respective type.
```
lukewagner: Why "operation" instead of instruction" as below?  Is "operation" a synonym for "instruction"?  Grepping, "operation" seems to be used somewhat interchangeably with "instruction", although it's more rare.

rossberg: The idea was that items in this list it ought to be in the same category as constants, conversions, and those other things listed, i.e., not the instructions but the numeric concepts that the corresponding instructions make available.

PR #1069, opened 2019-09-14
[spec] Fix float text format

### LINKED ISSUE #1066: Float literal syntax spec and tests don't match
The spec defines the syntax for [`fN`](https://webassembly.github.io/spec/core/text/values.html#text-float) to require a decimal point or exponent, and defines [`f(32|64).const`](https://webassembly.github.io/spec/core/text/instructions.html#numeric-instructions) to only accept a `fN` as their immediate operand.

However, there are many tests that use `f(32|64).const` without a decimal point or exponent; e.g.:
https://github.com/WebAssembly/spec/blob/d22a7659e5f2f09be9532e2c72c04407134cc08e/test/core/float_literals.wast#L82

rossberg:
Fixes #1066. Also fixes a recent bug in interpreter's double rounding mitigation and adds parsing tests.


AndrewScheidecker:
LGTM!

PR #1070, opened 2019-09-15
[spec] C.return should be replaced not prepend

Huxpro:
Since `C.return` is an optional occurrence, we should use _record field replacement_ defined at [structure - convention](http://webassembly.github.io/spec/core/syntax/conventions.html) instead of _prepend_  defined at [validation - convention](http://webassembly.github.io/spec/core/valid/conventions.html).

An alternative would be declaring the overloads for the `C, field A*` notation but I guess it would be more confusing than using already established `with`.

### DIFF
--- a/document/core/valid/modules.rst
+++ b/document/core/valid/modules.rst
@@ -42,7 +42,7 @@ Functions :math:`\func` are classified by :ref:`function types <syntax-functype>
    \frac{
      C.\CTYPES[x] = [t_1^\ast] \to [t_2^?]
      \qquad
-     C,\CLOCALS\,t_1^\ast~t^\ast,\CLABELS~[t_2^?],\CRETURN~[t_2^?] \vdashexpr \expr : [t_2^?]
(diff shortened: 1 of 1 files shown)


rossberg:
This may be a bit subtle but it actually is correct, because functions cannot be nested. So it is an invariant that C.return is epsilon before.

Huxpro:
@rossberg Thanks, but would you prefer to change `C.labels` and `C.locals` to `with` as well?

1. This is a little bit verbose but it requires no change as long as `with` is assumed left-associative.
```ml
C with locals = ... with labels = ... with return =
```

2. This would be ambiguous with prepending

```ml
C with locals = ..., labels = ..., return =
```

3. Unless we introduce new convention e.g.

```ml
C with locals = ...; labels = ...; return =
```

BTW, I could not get `make html` under `core` working:

```
λ make html
sphinx-build -b html -d _build/doctrees -D latex_paper_size=a4  . _build/html
Running Sphinx v2.2.0

Extension error:
Could not import extension mathdef (exception: cannot import name 'latex_visit_math')
make: *** [html] Error 2
```

It complained about the `mathdef` which I see it's defined under `code/util/mathdef.py`. I have no experience with sphinx and rst before so could you help me with that?

Issue #1072, opened 2019-09-16
[spec] validation rule for function is inaccurate


Huxpro:
The formalization of [Validation - Module - Function](http://webassembly.github.io/spec/core/valid/modules.html#functions) is incorrect, or at least inconsistent with the prose spec and reference interpreter.

In the premise of the validation rule, the `expr` is checked against the context of the original `C` with new locals, labels, and return elements **prepended** to each field, according to the [Validation - Convention](http://webassembly.github.io/spec/core/valid/conventions.html).

While the prose, stated that `locals`, `labels` and `return` are **set** to be the new element.
Also the reference interpreter, state an OCaml record functional update that replaces the field content:

```ocaml
let check_func (c : context) (f : func) =
  let {ftype; locals; body} = f.it in
  let FuncType (ins, out) = type_ c ftype in
  let c' = {c with locals = ins @ locals; results = out; labels = [out]} in
  check_block c' body out f.at
```

This situation, is different from other cases, e.g. `if`, where the original field content is con-ed:

```ocaml
  | If (ts, es1, es2) ->
    check_arity (List.length ts) e.at;
    check_block {c with labels = ts :: c.labels} es1 ts e.at;
    check_block {c with labels = ts :: c.labels} es2 ts e.at;
    [I32Type] --> ts
```

Although at least for the current version of web assembly, function declarations may only occur from the top level of modules, plus we do not have nested functions nor nested modules, we may safely assert these 3 fields (local, labels, return) would always be empty during the precondition of validating functions. Nonetheless, I'd argue having a distinguishment and being consistent and more accurate could be nicer (for all purposes including reading, in-paper formalization, mechanized formalization, proofs...)

I'd propose changing it to either following the record update from [Structure - Convention](http://webassembly.github.io/spec/core/syntax/conventions.html)

```
C, with locals = ... with labels = ... with return =
```

or extening the `with` notation with an OCaml-like batch-able update:

```
C, with locals = ...; labels = ...; return =
```

At very least, the `return` field is an optional `resulttype` that should not be **prepended** in any way. An alternative would be declaring the overloads of the `C, field A*` notation for option, but I guess that would be even more confusing.  A tentative PR of only this change is here https://github.com/WebAssembly/spec/pull/1070.

rossberg:
As noted on #1070, this is actually correct, albeit a bit subtle, for the reasons you mention. I agree that the change may be clearer.

PR #1077, opened 2019-09-26
[spec] C.fields should be replaced in valid/fun

Huxpro:
This commit is an update on #1070 (I carelessly force push the branch then the PR is closed).
This commit also changes `locals` and `labels` via `with` without adding new syntax as I questioned in PR #1070 and issue #1072.

The `document/README.md` talked about deployment as well. I assumed that's something only members do after PR approval?

### DIFF
--- a/document/core/valid/modules.rst
+++ b/document/core/valid/modules.rst
@@ -42,7 +42,7 @@ Functions :math:`\func` are classified by :ref:`function types <syntax-functype>
    \frac{
      C.\CTYPES[x] = [t_1^\ast] \to [t_2^?]
      \qquad
-     C,\CLOCALS\,t_1^\ast~t^\ast,\CLABELS~[t_2^?],\CRETURN~[t_2^?] \vdashexpr \expr : [t_2^?]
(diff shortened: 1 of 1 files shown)


rossberg:
After some consideration, I have a slight preference for leaving the formulation as is. It is fairly standard and avoids dependency on additional notation ("with" is not used anywhere else in the typing rules, only to manage state updates in the execution semantics). It is not uncommon to assume emptiness in certain cases of extending a context.

Huxpro:
@rossberg Sure. It bothers me mainly because of mechanization. It's nice to keep spec clean.

`locals` and `labels` are easier since they are `list` anyway. `return` is trickier since I defined it faithfully as `option resulttype`, and the spec work by overloading and assuming on epsilon: prepending a `resulttype` means `Some resulttype`, prepending epsilon means `None`, and prepending an `resulttype?` means take whatever as is (e.g. [Thread validity](https://webassembly.github.io/multi-value/core/appendix/properties.html#threads))

I will try to manage this 😂

PR #1092, opened 2019-10-08
Add tests for multiple start sections

AndrewScheidecker:


tlively:
Is this due to the proposal to allow multiple start sections in the conditional compilation proposal?

PR #1093, opened 2019-10-17
[spec] fix valid tableinst and meminst premises

Huxpro:
Background: see https://github.com/WebAssembly/multi-value/issues/29

This is the (1): the minimal and  straightforward fix


document/core/appendix/properties.rst
```diff
@@ -226,7 +226,7 @@ Module instances are classified by *module contexts*, which are regular :ref:`co
    \frac{
      ((S \vdash \EVFUNC~\X{fa} : \ETFUNC~\functype)^?)^n
      \qquad
-     \vdashlimits \{\LMIN~n, \LMAX~m^?\} \ok
+     \vdashlimits \{\LMIN~n, \LMAX~m^?\} 2^{32}
```
rossberg: I think these are missing a colon.

Huxpro:
@rossberg Thanks, proses are also updated. I probably shouldn't work before sleeping 😂

PR #1094, opened 2019-10-17
[spec] fix valid-limit (2)

Huxpro:
Background: see https://github.com/WebAssembly/multi-value/issues/29

This is (2). The benefits are that this approach reuses `2^{32}` and `2^{16}` defined in Validation - Type. This is also closer to the future [reference-type one](https://webassembly.github.io/reference-types/core/appendix/properties.html#table-instances) in spirit where both `tableinst` and `memoryinst` instantiate a `tabletype` and `memtype` valid rule to use in the premises.

TODO: Prose need to be updated if this approach is preferred.


Huxpro:
@rossberg Ah I see it. They are not structurally inductive premises is a great point. Thx!

PR #1101, opened 2019-11-16
Reference interpreter, step Return, clear stack

nomeata:
according to
https://webassembly.github.io/spec/core/exec/instructions.html#exec-return
the `Return` instruction pops the top of the stack, until the next
frame. The equivalent in the reference interpreter should be to set
the value stack to `[]`. This is what is happening for `Break` as well.
So for consistency, also do it here.

This has no visible effect: If the first instruction is `Returning`, the
stack is ignored everywhere. So this is purely cosmetic.

(I am only 80% sure; if not I hope to at least learn something here.)


nomeata:
(I just connected Github to my W3 account.)

PR #1104, opened 2019-11-19
[interpreter] Unify assert_result* assertions

rossberg:
Implement the base version of the assertion refactoring discussed [here](https://github.com/WebAssembly/simd/issues/137), unifying `assert_return_*_nan` with the regular `assert_return` by generalising the result literals of the latter to a simple form of pattern for result values:
```
result ::= (t.const pattern*)
pattern ::= literal | nan:canonical | nan:arithmetic
```
Here, `pat` extends the syntax of literals in const instructions with additional patterns that extend the existing nan-with-payload syntax. This could be extended to other non-deterministic value classes.

Motivation:
1. This makes more sense for multiple return values, which can now use different patterns for individual components.
2. Extends better to SIMD, which would extend the result pattern to allow different patterns for individual lanes.
3. Extends better to other proposals, such as reference types, which could further extend the syntax of `result` with additional patterns, such as `(ref.any)`.


tlively:
I can't say my OCaml is up to snuff, but this solution LGTM.

rossberg:
@abrown, for some reason I cannot add you as a reviewer, but please take a look.

interpreter/text/parser.mly
```diff
@@ -825,6 +834,14 @@ const_list :
   | /* empty */ { [] }
   | const const_list { $1 :: $2 }
 
+result :
+  | const { LitResult $1 @@ at () }
+  | LPAR CONST NAN RPAR { NanResult (nanop $2 ($3 @@ ati 3)) @@ at () }
+
+result_list :
+  | /* empty */ { [] }
+  | result result_list { $1 :: $2 }
```
abrown: Should the result list be inside parens? Right now, when there are multiple return values specified, it seems like we would write `assert_return (invoke ...) result1 result2 result3)` but should it instead be  `assert_return (invoke ...) (result1 result2 result3))`? (Asking, not claiming it should be...)

rossberg: That's already the case in the current format. It is symmetric to the argument list, which is not being parenthesised either:
```
(assert_return (invoke "f" arg1 arg2 arg3) res1 res2 res3)
```
In fact, parens there would require inserting another keyword as head.

rossberg:
@abrown, thanks. Could you approve the PR?

There is no build failure, it's just the automatic IPR checker. For some reason, it doesn't recognise my GH account anymore. I should probably try to figure out why some day...

abrown:
I approved but my approval must not count for much ☹️. @tlively are you able to approve and merge?

tlively:
LGTM, thanks for taking a closer look @abrown! I'll let @rossberg merge when he's ready.

abrown:
@rossberg, ping

rossberg:
Sorry for the delay, I was OOO.

PR #1105, opened 2019-11-21
Test argument passing

rianhunter:
This change adds tests to make sure passing many arguments across "call" instructions is correct.


rianhunter:
Apparently I'm not affiliated, that's fine. I release this change into the public domain.

test/core/call.wast
```diff
@@ -225,6 +225,85 @@
   (func (export "as-convert-operand") (result i64)
     (block (result i64) (i64.extend_i32_s (call $dummy (i32.const 1))))
   )
+
+  ;; Test correct argument passing
+
+  (func $return-arg-0-helper (param i32) (result i32)
+        (local.get 0)
```
rossberg: Nit: consistent indentation (here and below)

rianhunter:
Thanks for reviewing, I'll make those changes!

rianhunter:
Okay I made the changes, just a note, I did not verify if the reference interpreter passed this test.

rossberg:
Ah, sorry, that was a misunderstanding. I didn't mean to suggest to enumerate *all* possible parameter lengths up to 100, just add one test with 100. Enumerating all cases feels like overkill and makes the test file difficult to manoeuvre.

Also, it might be good to make this one "big" call test a bit more diverse and mix different parameter types randomly. That is the kind of thing that compilers tend to have problems with when compiling for different calling conventions.

rianhunter:
Oops yeah that was a bit of a reading comprehension fail on my part. I simplified the PR to a single new test. I removed the other tests because this single test was sufficient for my original purposes while also being interesting enough to have wider applicability.

PR #1109, opened 2019-12-10
[interpreter] Use hexadecimal float literals

ggreif:
This should be squashed on merge.

Since OCaml 4.07 is now a prerequisite, the more obvious float literal syntax can be used. Resolves a few FIXMEs.

@rossberg regarding contributor agreement, what is the process here? Alternatively you could git-sign-off this one.


rossberg:
Ah, 4.07 is a prerequisite for the wasm Opam package, but not for building the interpreter per se.

To contribute, all you have to do is join the [Wasm CG](https://www.w3.org/community/webassembly/). The trivial fixes like this would be fine anyway.

ggreif:
This is now merged via #1191. Thanks.

PR #1133, opened 2020-02-17
[spec] Fix negative value nit

### LINKED ISSUE #1132: i32.const -1 violates the formal grammar
Call me pedantic, but [the definition of const](https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-numeric)

> instr ::= ... | i*nn*.const i*nn* | ...

implies 

> instr ::= ... | i32.const i32 | ...

and i32 as defined [here](https://webassembly.github.io/spec/core/syntax/values.html#integers)

> i*N* ::= u*N*
u*N* ::= 0 | 1 | ... | 2^*N* -1

implies

> i32 ::= 0 | 1 | ... | 2^32 -1

which *technically* means that the use of `i32.const -1` [in the definition of memory.grow](https://webassembly.github.io/spec/core/exec/instructions.html#exec-memory-grow) violates the formal grammar.

rossberg:
Fixes #1132.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -591,13 +591,15 @@ Memory Instructions
 
 8. Pop the value :math:`\I32.\CONST~n` from the stack.
 
-9. Either, try :ref:`growing <grow-mem>` :math:`\X{mem}` by :math:`n` :ref:`pages <page-size>`:
+9. Let :math:`\X{err}` be the |i32| value :math:`2^{32}-1`, for which :math:`\signed_{32}^{-1}(\X{err})` is :math:`-1`.
+
+10. Either, try :ref:`growing <grow-mem>` :math:`\X{mem}` by :math:`n` :ref:`pages <page-size>`:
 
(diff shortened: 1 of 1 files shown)


document/core/exec/instructions.rst
```diff
@@ -591,13 +591,15 @@ Memory Instructions
 
 8. Pop the value :math:`\I32.\CONST~n` from the stack.
 
-9. Either, try :ref:`growing <grow-mem>` :math:`\X{mem}` by :math:`n` :ref:`pages <page-size>`:
+9. Let :math:`\X{err}` be the |i32| value :math:`2^{32}-1`, for which :math:`\signed_{32}^{-1}(\X{err})` is :math:`-1`.
```
hexagonrecursion: I think you got it backwards. It should be
```
:math:`\signed_{32}(\X{err})` is :math:`-1`
```

rossberg: Oops, right, thanks, fixed.

PR #1135, opened 2020-02-26
Assert duplicate identifiers are invalid

### LINKED ISSUE #1134: data.wast and elem.wast test assert duplicate identifiers in text format are accepted
According to https://webassembly.github.io/spec/core/text/conventions.html#contexts it says:

> An identifier context is well-formed if no index space contains duplicate identifiers.

But there are two tests which actually assert that duplicate identifiers are accepted:

* https://github.com/WebAssembly/spec/blob/5cc5c69844d879016fc81ba0654c6f02244c315b/test/core/data.wast#L15-L18
* https://github.com/WebAssembly/spec/blob/master/test/core/elem.wast#L15-L18

Albeit these are for future features to wasm so they're not technically covered by the current spec as-is. Is this a bug in the tests that the identifier names should be updated there to not collide?

Additionally this was originally sourced from https://github.com/bytecodealliance/wat/issues/60 where the text parser there doesn't account for duplicate identifiers. This leads me to conclude that there aren't actually any tests in the test suite right now which assert that duplicate identifiers fail to parse, is that perhaps an intentional omission or something that should be added to the test suite? (I'd be happy to send a PR!)

alexcrichton:
This commit adds tests asserting that duplicate identifiers fail to
parse in the text format. This was already a property of the core spec,
and should help aspiring text parsers to be sure to reject programs
which have duplicate identifiers!

Closes #1134 (since it clarifies that)


test/core/elem.wast
```diff
@@ -379,3 +379,18 @@
 (assert_return (invoke $module1 "call-7") (i32.const 67))
 (assert_return (invoke $module1 "call-8") (i32.const 69))
 (assert_return (invoke $module1 "call-9") (i32.const 70))
+
+;; Duplicate table identifiers
```
rossberg: These have nothing to do with elem, so should go to a new table.wast test (which for some reason we are lacking).

alexcrichton: Ah good point, done now! Wasn't sure if that was ok to add here :)

Also the tests should be updated to actually pass now

test/core/tables.wast
```diff
@@ -0,0 +1,14 @@
+;; Duplicate table identifiers
```
rossberg: Nit: rename file to table.wast, to be named after the construct (like similar cases).

alexcrichton: Ah ok, should `globals.wast` get renamed to `global.wast`?

rossberg: Yeah, it should. It's one of the clean-ups that ought to be done for the current test suite. Feel free to include it if you feel like it. :)

alexcrichton:
Ok everything should be pushed up!

PR #1137, opened 2020-03-05
Allow test compilation on Python 3.6

wingo:
The `text` parameter (added in Python 3.7) is simply an alias for `universal_newlines` (added in Python 3.5), which we are already passing.  Remove the duplicate keyword argument to allow tests to generate on Ubuntu 18.04, which has Python 3.6.


wingo:
See https://docs.python.org/3/library/subprocess.html#subprocess.run

wingo:
I just updated to link my w3c account with github, hopefully the bot runs again.  (Affiliation is Igalia.)

PR #1144, opened 2020-03-18
Merge sign-extension-ops proposal into spec

binji:
See the sign-extension-ops proposal here:

https://github.com/WebAssembly/sign-extension-ops

This PR is built on top of #1143 (merge nontrapping-float-to-int).

PR #1143, opened 2020-03-18
Merge nontrapping-float-to-int proposal into spec

binji:
See the non-trapping-float-to-int-conversions proposal here:

https://github.com/WebAssembly/nontrapping-float-to-int-conversions

PR #1145, opened 2020-03-19
Merge multi-value proposal into spec

binji:
See the multi-value proposal here:

https://github.com/WebAssembly/multi-value

This PR is built on top of the following PRs:

* #1143 (merge nontrapping-float-to-int)
* #1144 (merge sign-extension-ops)


rossberg:
Hm, GH isn't good at dealing with dependent PRs. Is there a reason that we cannot merge proposals one at a time?

binji:
> Is there a reason that we cannot merge proposals one at a time?

We can, but when we discussed this in the WG meeting, we decided it would be valuable to explicitly have them ordered with the actual PRs ready to go.

I just realized I can make this a little easier to read by having each PR against the branch of the previous PR, so it's easier to see the diffs. I'll try that instead.

binji:
OK, updated. PTAL

document/core/appendix/properties.rst
```diff
@@ -157,8 +157,6 @@ Module instances are classified by *module contexts*, which are regular :ref:`co
 :ref:`Host Function Instances <syntax-funcinst>` :math:`\{\FITYPE~\functype, \FIHOSTCODE~\X{hf}\}`
 ..................................................................................................
 
-* The :ref:`function type <syntax-functype>` :math:`\functype` must be :ref:`valid <valid-functype>`.
```
rossberg: Hm, not sure why this is removed on the multi-value branch, perhaps some earlier merge mistake. I created a [PR on the repo](https://github.com/WebAssembly/multi-value/pull/43) to fix it.

proposals/multi-value/Overview.md
```diff
@@ -0,0 +1,196 @@
+# Multi-value Extension
```
Ms2ger, line 1: @binji I only realized now that you added the `proposals/` dir too. Was that intentional?

binji: It was; I thought we decided to do this so we could easily see that you could easily see which proposals had been merged into the spec. That said, it was a long time ago and I could be misremembering.

rossberg: At least it matches my memory as well. :)

chfast:
As this extensions is merged, and it modifies existing test files, is there a way to send new tests for 1.0?

Ms2ger:
Are you asking whether someone is maintaining a test suite specifically for Wasm 1.0, excluding tests / test changes for proposals that have been merged since? If so, I don't think so. You can always submit tests to the master branch here, of course.

chfast:
Yes, I can submit tests to maser, but I will not be able to use them directly from master. Thanks for the answer anyway. I don't want to start longer discussion here.

PR #1151, opened 2020-03-24
Set travis env to Python 3.7

ngzhian:
Bikeshed requires 3.7, see
https://github.com/tabatkins/bikeshed/issues/321#issuecomment-601295711

I don't think we can run both 2.7 and 3.7 in travis, we will need to to use 2 different versions of bikeshed, but no way to specific which version of bikeshed in which version of python (could be wrong).

### DIFF
--- a/document/core/util/mathjax2katex.py
+++ b/document/core/util/mathjax2katex.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 # -*- coding: latin-1 -*-
 
 import queue
--- a/document/core/util/mathjax2katex.py
+++ b/document/core/util/mathjax2katex.py
@@ -239,7 +239,7 @@ def Main():
       q.task_done()
       sys.stderr.write('.')
 
-  q = Queue.Queue()
+  q = queue.Queue()
   for i in range(40):
     t = threading.Thread(target=Worker)
     t.daemon = True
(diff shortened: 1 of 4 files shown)


ngzhian:
Cool! builds after some small fixes in `document/core/util/mathjax2katex.py`, see https://travis-ci.org/github/WebAssembly/spec/builds/666530476?utm_medium=notification&utm_source=github_status.

binji:
Wow, surprisingly small amount of changes. I was expecting more effort here :)

binji:
@sbc100 I know you were doing some more python2 -> python3 conversions, do you see anything else we might want to change?

document/core/util/mathjax2katex.py
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env python
```
binji: should this be python3 now?

ngzhian: Think this will be python3 since the travis setup only has python3. But will be useful for local builds. I will make the change.

sbc100:
Seems reasonable.

binji:
Hm, just noticed that there are a lot of errors in the output that look like "!!! Error processing fragment". See line 5821 here: https://travis-ci.org/github/WebAssembly/spec/builds/666836387?utm_source=github_status&utm_medium=notification. Can you take a look?

ngzhian:
Woah woops, should I do a revert of this first?

binji:
I don't think it's urgent to fix immediately -- it looks like https://webassembly.github.io/spec still works, and we don't usually have a lot of changes to it.

ngzhian:
Sg, I'm looking at it now to figure out what the error messages mean, if I can't figure it out (maybe by lunch?) I'll just send a PR to change this back, then fix it proper later.

PR #1163, opened 2020-03-27
Upgrade travis environment to Bionic

ngzhian:
Also use the apt addon [0] to install our dependencies.

[0] https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-with-the-apt-addon

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ addons:
 
 install:
   - pip install Sphinx==2.3.0
-  - sudo apt-get install texlive-full yarn
   - git clone https://github.com/tabatkins/bikeshed.git
   - pip install --editable $PWD/bikeshed
   - bikeshed update
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,20 @@ language: c++
 language: python
 python:
   - "3.7"
+dist: bionic
 
-before_install:
-  - curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,14 +2,20 @@ language: c++
...
+      - sourceline: 'deb https://dl.yarnpkg.com/debian/ stable main'
+        key_url: 'https://dl.yarnpkg.com/debian/pubkey.gpg'
+    packages:
+      - ocaml
+      - ocamlbuild
+      - texlive-full
+      - yarn
 
 install:
-  - ./interpreter/meta/travis/install-ocaml.sh
```
binji: Looks like this could be removed now.

.travis.yml
```diff
@@ -2,14 +2,20 @@ language: c++
...
   - pip install Sphinx==2.3.0
   - sudo apt-get install texlive-full yarn
```
binji: is this still needed?

binji:
Oh sorry just realized this is WIP :)

ngzhian:
Yea sorry, I don't know how to test travis besides sending a PR :) I'll request for review when this is ready, thanks!

.travis.yml
binji: The `install-ocaml.sh` isn't needed anymore?

ngzhian: Yea it can be removed. I use case is when a developer does not have ocaml 4.05.0 from their package repo, they can use the script to download and build ocaml.

binji: OK, makes sense.

Issue #1157, opened 2020-03-26
Upgrade Sphinx


ngzhian:
Currently we fix Sphinx to 1.7.9 (see https://github.com/WebAssembly/spec/blob/master/.travis.yml#L15), which though was just released in Sept 2018, has already seen many new release (~25, it is at v2.4.4 now). And these release have many, many changes https://www.sphinx-doc.org/en/master/changes.html#release-1-8-0-released-sep-13-2018.

I've been mucking around to see if this will be a straightforward task, generally speaking it is not, since we have a bunch of custom extensions (see https://github.com/WebAssembly/spec/blob/master/document/core/util/mathdef.py) that does interesting stuff. I also don't have a good way of comparing the existing output v.s. the output generated by upgrading to 1.8.0 + fixes to our scripts. Best I can do is try to diff and see what changed, and for binary files like pdfs, click around and see if things work, focusing on the math and math_block nodes.

As I'm typing this, I found #1075, so it looks like upgrading to 1.8.0 might not be as easy to upgrade to v2.3.0 (the version where the accidentally removed function was added back), so I'll try that too.

Side note: instructions for building document/ is incorrect, see https://github.com/WebAssembly/spec/blob/master/document/README.md, pip install sphinx will get you a newer version of Sphinx, which will error out when you build stuff.

PR #1164, opened 2020-03-27
Remove sed commands to change mathjax url

ngzhian:
With the Sphinx upgrade in #1157 we no longer need these sed commands.
At the same time, clean up the way we configure mathjax, using a config
value, instead of sed.


rossberg:
Oops, was this not ready for merging yet?

ngzhian:
That's fine :) We will still need #1167 to fix the broken renderings on HTML.

PR #1167, opened 2020-03-28
[spec] Fix TeX maxbuffer option for MathJax

### LINKED ISSUE #1166: 'Instructions' page for binary format is broken
https://webassembly.github.io/spec/core/binary/instructions.html#instructions

'Memory Instructions' section and part of 'Numeric Instructions' section are broken:

スクリーンショット 2020-03-28 14 32 50

スクリーンショット 2020-03-28 14 33 10

My environment is:

- macOS 10.14
- Chrome 81.0.4044.69

ngzhian:
When we upgraded Sphinx, the MathJax version changed, which has a
different way of configuring TeX options. This resulted in TeX's
maxbuffer not being set right, and so processing stopped.

Fixes #1166

### DIFF
--- a/document/core/conf.py
+++ b/document/core/conf.py
@@ -480,7 +480,7 @@ rst_prolog = """
 """
 
 # https://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-mathjax_config
-# https://docs.mathjax.org/en/latest/options/input/tex.html#the-configuration-block
+# http://docs.mathjax.org/en/v2.7-latest/options/input-processors/TeX.html
 mathjax_config = {
     'TeX': { 'MAXBUFFER': 30*1024 },
 }
--- a/document/core/conf.py
+++ b/document/core/conf.py
@@ -482,5 +482,5 @@ rst_prolog = """
 # https://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-mathjax_config
 # https://docs.mathjax.org/en/latest/options/input/tex.html#the-configuration-block
 mathjax_config = {
-    'tex': { 'maxBuffer': 30*1024 },
+    'TeX': { 'MAXBUFFER': 30*1024 },
 }


ngzhian:
Note that this is different from #1164 (which is not yet merged). This change will fix the current issue. I will later test and verify #1164 will not cause the same problem.

ngzhian:
Yea I should have mentioned, somewhere along, in the upgrade, the way MathJax is configured changed: see http://docs.mathjax.org/en/latest/web/configuration.html#configuring-mathjax as the recommended way to configure.
I think it has to do with how the MathJax script is loaded async, the old way, calling `MathJax.Hub.Config`, wouldn't work with an async script, since that variable wouldn't be around yet. By assigning to `MathJax`, the script, when loaded, can check this global variable, and initialize itself with our provided configuration.
I didn't dig too much into which exact versions caused this difference.

ngzhian:
I take my last comment back, I'm wrong.
The version that changed the way configuration worked was 3.0, but we are not on 3.0 yet.
Rather, the reason why those equations weren't rendering was because I was spelling the config wrong - it is case sensitive.
I have a different fix, let me push it to this PR.

ngzhian:
Okay, sorry for the noise, this PR is now ready. The change is to fix the casing for TeX and MAXBUFFER. This change can go on top of #1164 (since that is the change that introduced using `conf.py` to configure MathJax, and is also what introduced the bug with casing).

binji:
> The change is to fix the casing for TeX and MAXBUFFER.

Is there a doc page that has this? The link in the comments there currently says it is `maxBuffer`, I assume that's the newer version?

ngzhian:
Yes there is, since we are using 2.7.5, the configuration value is http://docs.mathjax.org/en/v2.7-latest/options/input-processors/TeX.html

binji:
Ah, nice. Can you update the comment here to point to that one instead?

ngzhian:
Done!

PR #1171, opened 2020-04-02
Upgrade to latest Sphinx release (2.4.4)

### LINKED ISSUE #1157: Upgrade Sphinx

ngzhian:
Fixes #1157

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ addons:
       - yarn
 
 install:
-  - pip install Sphinx==2.3.0
+  - pip install Sphinx==2.4.4
   - git clone https://github.com/tabatkins/bikeshed.git
   - pip install --editable $PWD/bikeshed
   - bikeshed update


ngzhian:
I did not look much at the changelog to see what changed, but there have been a number of bug fixes. Plus this will "hopefully" make it easier to keep up to date with Sphinx in the future.

PR #1121, opened 2020-02-03
[js-api] Bump limit for maximum memory size to 4GiB

jakobkummerow:
Per discussion in #1116, the limit for initial memory pages remains at
32767 (just under 2GiB), while the limit for maximum memory pages is
increased to 65536 (4GiB).
Spec tests are brought in line with these definitions (they actually did
not reflect the spec text before).

Fixes: #1116.

PR #1174, opened 2020-04-02
[js-api] Bump limit for initial memory size to 4GiB

kripken:
As agreed upon in the CG meeting earlier this week, we can allow 4GB in both the initial and maximum memory sizes, and not just the maximum.

Followup to #1121 which raised the limit to the initial size.


Ms2ger:
@kripken please [link your github account to your w3c account](https://www.w3.org/users/myprofile/connectedaccounts).

kripken:
@Ms2ger Oh, sorry about that, looks like it took a while for it to propagate. It looks ok now after I added more fake commits.

Ms2ger:
Thanks!

PR #1176, opened 2020-04-08
[spec] Fix notation of optional else block

dszakallas:
I think I found a minor nit: the optional cardinality is supposed to be applied to the whole `else` block, parens are unbalanced this way.


binji:
While we're waiting for @rossberg (I believe he's on holiday), do you mind updating the PR to integrate with changes to master?

rossberg:
As a note, this was an intentional pun, reusing the object parens as meta parens, because the double parens probably read oddly when rendered. But I agree it's probably clearer this way.

aleee1333 (approved):

PR #1179, opened 2020-04-20
[spec] Add missing i64.extend32_s syntax

### LINKED ISSUE #1180: Typo
There is a missing star after `result` in the expression in the end of section 6.6.3 after "..of the form:".

rossberg:
Instruction was missing from abstract syntax. Fixes #1180 as well. Also, tweak some identifier names for consistency.

Also, bump spec release to 1.1.


document/core/exec/numerics.rst
```diff
@@ -614,14 +614,14 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
 
 .. _op-iextendn_s:
```
binji, line 615: change this too?

rossberg: This doesn't actually expand to anything, so I'm fine with keeping it. The other made it difficult to understand the def from the Latex source, because the M on the rhs seemed to becoming out of nowhere.

PR #1186, opened 2020-04-28
Add a README to the proposals directory.

Ms2ger:


Ms2ger:
I wasn't sure these were supposed to be there, but apparently yes per https://github.com/WebAssembly/spec/pull/1145#discussion_r415822066.

Ms2ger:
I could if you think that's useful, but it seems like it'd get out of date easily and there's already a list at the top when you look at https://github.com/WebAssembly/spec/tree/master/proposals; this readme would be rendered below that.

PR #1194, opened 2020-05-20
Add .ocamlformat file to help format code

ngzhian:
This file could be useful for autoformatting code. I would like to start
using it in some of the SIMD work, and it might be useful to have in in
the spec repo.

Note that this is not going to cause any automatic formatting, you would
have to run ocamlformat yourself. I also don't intend to reformat all
the code, since that will cause troublesome merge conflicts. I will
however like to format files that only exist in the SIMD proposal, which
will not cause a problem when we eventually merge it in.


ngzhian:
The version used in .ocamlformat was chosen based on https://github.com/ocaml-ppx/ocamlformat#what-configuration-should-i-use

rossberg:
Hm, I'm fine with using this for SIMD stuff, but does the file have to be in the repo for that? Including it may send a confusing message and make other contributors think that it's supposed to be applied elsewhere, which is problematic for the reasons you mentioned.

ngzhian:
Yea, I didn't think of the message it will send. Agreed on you point, I will keep it locally for formatting :)
Thanks for pointing it out!

PR #1195, opened 2020-05-20
Inconsistency of implementation-defined limits

gahaas:
This PR resolves an inconsistency of implementation-defined limits in the js-api specification, and execution limits in the core specification. The core specification requires under [execution limits](https://webassembly.github.io/spec/core/appendix/implementation.html#execution) that implementations throw a runtime error if tables or memories grow beyond an implementation-defined limit. The js-api specification, however, required a compile error if the module defines limits beyond the implementation-defined limits.

This PR resolves this inconsistency by adjusting the js-api specification. For the maximum memory size and maximum table size, a RuntimeError is now required instead of a CompileError.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1122,6 +1122,7 @@ Note: ECMAScript doesn't specify any sort of behavior on out-of-memory condition
 The WebAssembly core specification allows an implementation to define limits on the syntactic structure of the module.
 While each embedding of WebAssembly may choose to define its own limits, for predictability the standard WebAssembly JavaScript Interface described in this document defines the following exact limits.
 An implementation must reject a module that exceeds one of the following limits with a {{CompileError}}:
+In practice, an implementation may run out of resources for valid modules below these limits.
 
 <ul>
 <li>The maximum size of a module is 1073741824 bytes (1 GiB).</li>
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -1121,8 +1121,7 @@ Note: ECMAScript doesn't specify any sort of behavior on out-of-memory condition
 
 The WebAssembly core specification allows an implementation to define limits on the syntactic structure of the module.
 While each embedding of WebAssembly may choose to define its own limits, for predictability the standard WebAssembly JavaScript Interface described in this document defines the following exact limits.
-An implementation must reject a module that exceeds these limits with a {{CompileError}}.
-In practice, an implementation may run out of resources for valid modules below these limits.
```
rossberg, line 1125: Perhaps we should keep this sentence here as well, since it may apply even to static limits.

document/js-api/index.bs
```diff
@@ -1134,16 +1133,21 @@ In practice, an implementation may run out of resources for valid modules below
...
 <li>The maximum size of a function body, including locals declarations, is 7654321 bytes.</li>
 <li>The maximum number of locals declared in a function, including implicitly declared as parameters, is 50000.</li>
+</ul>
 
+An implementation must throw a {{RuntimeError}} if one of the following limits is exceeded during runtime:
+In practice, an implementation may run out of resources for valid modules below these limits.
+
+<ul>
+<li>The maximum size of a table is 10000000.</li>
+<li>The initial or maximum number of pages for any memory, declared or imported, is at most 65536.</li>
```
rossberg, line 1150: The static sizes in memory types are already limited to that value by validation. So this could be simplified to: 
```suggestion
<li>The maximum number of pages of a memory is 65536.</li>
```
(FWIW, the reference types proposal refactored the semantics of grow in the core spec and by way of that also adds this as a runtime constraint. It was an oversight that it didn't exist before.)

rossberg:
Thanks!

PR #1198, opened 2020-05-25
[tests] Adjust limits test to dynamic limits

gahaas:
This PR changes the tests in limits.any.js for memory size and table size to validate and compile modules that exceed implementation-defined limits. Only during instantiation and runtime, errors should be thrown.

In https://github.com/WebAssembly/spec/pull/1195 @Ms2ger requested that these tests get adjusted to to changes to the spec.

### DIFF
--- a/test/js-api/limits.any.js
+++ b/test/js-api/limits.any.js
@@ -10,13 +10,13 @@ const kJSEmbeddingMaxExports = 100000;
 const kJSEmbeddingMaxGlobals = 1000000;
 const kJSEmbeddingMaxDataSegments = 100000;
 
-const kJSEmbeddingMaxModuleSize = 1024 * 1024 * 1024;  // = 1 GiB
+const kJSEmbeddingMaxModuleSize = 1024 * 1024 * 1024; // = 1 GiB
 const kJSEmbeddingMaxFunctionSize = 7654321;
 const kJSEmbeddingMaxFunctionLocals = 50000;
 const kJSEmbeddingMaxFunctionParams = 1000;
-const kJSEmbeddingMaxFunctionReturns = 1;
+const kJSEmbeddingMaxFunctionReturns = 1000;
 const kJSEmbeddingMaxElementSegments = 10000000;
-const kJSEmbeddingMaxTables = 1;
+const kJSEmbeddingMaxTables = 100000;
 const kJSEmbeddingMaxMemories = 1;
 
 // Dynamic limits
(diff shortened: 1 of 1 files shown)


gahaas:
ping

Ms2ger:
Trying to get to it this week or next

gahaas:
ping

test/js-api/limits.any.js
```diff
@@ -167,30 +150,81 @@ testLimit("function returns", 0, kJSEmbeddingMaxFunctionReturns, (builder, count
...
+  test(() => { assert_true(WebAssembly.validate(buffer)); },
+       `Validate ${name} beyond its dynamic limit`);
+
+  test(() => { new WebAssembly.Module(buffer); },
+       `Compile ${name} beyond its dynamic limit`);
+
+  promise_test(t => { return WebAssembly.compile(buffer); },
+               `Async compile ${name} beyond its dynamic limit.`);
+
+  if (test_instantiation) {
```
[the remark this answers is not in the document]
gahaas: I tested instantiation only for those values where instantiation should fail, e.g. initial memory size.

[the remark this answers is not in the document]
Ms2ger: Hmm, should we try testing that instantiation *doesn't* fail for the others, or will those fail anyway because the actual limit will be lower?

test/js-api/limits.any.js
```diff
@@ -167,30 +150,81 @@ testLimit("function returns", 0, kJSEmbeddingMaxFunctionReturns, (builder, count
...
+    promise_test(t => {
+      return promise_rejects(t, new WebAssembly.RuntimeError(),
+                             WebAssembly.instantiate(buffer, {}));
+    }, `Async instantiate ${name} over limit`);
+  }
+}
+
+testDynamicLimit(
+    "initial declared memory pages", kJSEmbeddingMaxMemoryPages, true,
+    (builder, count) => { builder.addMemory(count, undefined, false, false); });
```
Ms2ger, line 210: We should also test the case where a memory is initially within the limit, but then is grown past the limit, including with the `memory.grow` instruction in wasm proper, and with `WebAssembly.Memory.grow()`.

However, note that the spec already requires a `RangeError` if `mem_grow` fails. It doesn't seem well-defined whether `RangeError` or `RuntimeError` wins here.

test/js-api/limits.any.js
```diff
@@ -167,30 +150,81 @@ testLimit("function returns", 0, kJSEmbeddingMaxFunctionReturns, (builder, count
...
+    (builder,
+     count) => { builder.addImportedMemory("mod", "mem", count, undefined); });
+
+testDynamicLimit(
+    "maximum imported memory pages", kJSEmbeddingMaxMemoryPages, false,
+    (builder, count) => { builder.addImportedMemory("mod", "mem", 1, count); });
+
+testDynamicLimit(
+    "initial table size", kJSEmbeddingMaxTableSize, true,
+    (builder, count) => { builder.setTableBounds(count, undefined); });
```
Ms2ger, line 227: Same about growing.

test/js-api/limits.any.js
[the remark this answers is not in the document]
gahaas: I added tests now for succeeding instantiation, and also for growing.

test/js-api/limits.any.js
gahaas: I added tests for growing now. I changed the expectation to RangeError.

test/js-api/limits.any.js
gahaas: Done.

gahaas:
I updated the test now to test both failing and succeeding instantiation, and also growing beyond the implementation-defined limits. Sorry that it took so long.

I also changed the values for `kJSEmbeddingMaxFunctionReturns` and `kJSEmbeddingMaxTables`, because with recent additions to the spec, these values are higher now.

gahaas:
I just realized that the reference types proposal is not merged yet. I reverted the number of allowed tables to 1, and removed the table.grow test.

gahaas:
I removed the memory limits tests now. The implementation-defined memory limit is the same as the memory limit of the core spec. Therefore this limit is already covered by the core spec tests.

gahaas:
@Ms2ger ping

gahaas:
@Ms2ger ping

PR #1203, opened 2020-06-05
[interpreter] Simplify ast.ml code by opening Source a bit earlier

ggreif:
Qualifying with `Source.` and later opening completely looks like an oversight to me.

This PR rectifies that. If it was done on purpose, please just close.


rossberg:
It was actually intentional, in order to increase explicitness and thus readability for the type definitions. ;)

PR #1206, opened 2020-06-09
Add some more instructions on how to build the spec

### LINKED ISSUE #1155: Document build instructions
I was trying to repro some errors that I observed in travis, and realized there were a couple of missing steps that don't seem to be documented elsewhere. I'll write some of those down here, and probably can put it in README.md or CONTRIBUTING.md later.

1. when this repo is cloned, we need to clone submodules too, `git clone --recurse-submodules https://github.com/WebAssembly/spec`, this is to bring in document/core/util/katex
2. you need to install yarnpkg (and npm), you will see this in .travis.yml
3. you will need texlive-full (also in .travis)

ngzhian:
Fixed #1155.

### DIFF
--- a/document/README.md
+++ b/document/README.md
@@ -24,6 +24,8 @@ Please make sure to only use that once a change has approval.
 
 ## Step by step guide to building the spec
 
+### Checking out the repository
+
 Make sure this repository was cloned with `--recursive`:
 
 ```
@@ -41,6 +43,8 @@ The rest of these instructions assume you are in the directory where is README i
 ```
 cd spec/document
 ```
+
+### Prerequisites
 
 You will need `python3.7`, and `pip`. `pip` should come with Python, if not follow [these installation instructions for `pip`](https://pip.pypa.io/en/stable/installing/), or check your system package manager for `pip3`.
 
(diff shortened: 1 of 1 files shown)


document/README.md
```diff
@@ -21,3 +21,65 @@ To build everything and update [webassembly.github.io/spec](https://webassembly.
...
+
+```
+pip install Sphinx==2.4.4
+```
+
+You can now build the [multi-page html document](https://webassembly.github.io/spec/core/):
+
+```
+make -C core html
+```
```
rossberg, line 68: Would be good to also include the PDF target. Sphinx and MathJax are rather sloppy and forgiving, so that most errors in LaTeX bits of the doc only surface when you build the PDF. Moreover, the PDF is included in the deployment, so the build needs to work.

document/README.md
```diff
@@ -21,3 +21,65 @@ To build everything and update [webassembly.github.io/spec](https://webassembly.
...
+git submodule update --init --recursive
+```
+
+The rest of these instructions assume you are in the directory where is README is:
+
+```
+cd spec/document
+```
+
+You will need `python3.7`, and `pip`. `pip` should come with Python, if not follow [these installation instructions for `pip`](https://pip.pypa.io/en/stable/installing/), or check your system package manager for `pip3`.
```
rossberg, line 45: Maybe split this into subsections like "Perequisites", "Checking Out the Repository", "Building"?

Ms2ger:
Maybe add something about the jsapi/webapi specs too, and we probably shouldn't put this in the README to avoid clutter on the main https://github.com/WebAssembly/spec page

document/README.md
```diff
@@ -42,6 +44,8 @@ The rest of these instructions assume you are in the directory where is README i
 cd spec/document
 ```
 
+### Prerequisites
```
rossberg, line 47: Nit: Seems more logical to make this section the first one.

ngzhian: Done!

PR #1216, opened 2020-07-01
Add merlin config file

ngzhian:
This is useful to specify our soruce/build structure to Meriln to enable
IDE like features (syntax check in editor, go to definition, etc.)

See https://github.com/ocaml/merlin.


ngzhian:
I found this to be helpful when doing development in the SIMD repository, I think it will a useful addition to the main spec as well.

PR #1223, opened 2020-07-20
Add more i32.trunc_f64_s conversion checks

kripken:
I may be missing something here, sorry if so. Looks like the
existing tests check for `2147483647.0` and `2147483648.0`,
but not the precise fraction where things change. I think that
is fully defined by the spec, so it should be testable?

I noticed this as the binaryen fuzzer found a difference between
the binaryen interpreter and other VMs, so I assume the spec
interpreter is right and these tests are valid?

### DIFF
--- a/test/core/conversions.wast
+++ b/test/core/conversions.wast
@@ -120,6 +120,8 @@
 (assert_return (invoke "i32.trunc_f64_s" (f64.const -2.0)) (i32.const -2))
 (assert_return (invoke "i32.trunc_f64_s" (f64.const 2147483647.0)) (i32.const 2147483647))
 (assert_return (invoke "i32.trunc_f64_s" (f64.const -2147483648.0)) (i32.const -2147483648))
+(assert_return (invoke "i32.trunc_f64_s" (f64.const 2147483647.9999998)) (i32.const 2147483647))
+(assert_trap   (invoke "i32.trunc_f64_s" (f64.const 2147483647.9999999)) "integer overflow")
 (assert_trap (invoke "i32.trunc_f64_s" (f64.const 2147483648.0)) "integer overflow")
 (assert_trap (invoke "i32.trunc_f64_s" (f64.const -2147483649.0)) "integer overflow")
 (assert_trap (invoke "i32.trunc_f64_s" (f64.const inf)) "integer overflow")


test/core/conversions.wast
```diff
@@ -120,6 +120,8 @@
 (assert_return (invoke "i32.trunc_f64_s" (f64.const -2.0)) (i32.const -2))
 (assert_return (invoke "i32.trunc_f64_s" (f64.const 2147483647.0)) (i32.const 2147483647))
 (assert_return (invoke "i32.trunc_f64_s" (f64.const -2147483648.0)) (i32.const -2147483648))
+(assert_return (invoke "i32.trunc_f64_s" (f64.const 2147483647.9999998)) (i32.const 2147483647))
+(assert_trap   (invoke "i32.trunc_f64_s" (f64.const 2147483647.9999999)) "integer overflow")
```
binji, line 124: This value seems to be equivalent to `2147483648.0`, so it's really just testing the wat double parsing; I'm assuming we don't care much about that here.

test/core/conversions.wast
```diff
@@ -120,6 +120,8 @@
 (assert_return (invoke "i32.trunc_f64_s" (f64.const -2.0)) (i32.const -2))
 (assert_return (invoke "i32.trunc_f64_s" (f64.const 2147483647.0)) (i32.const 2147483647))
 (assert_return (invoke "i32.trunc_f64_s" (f64.const -2147483648.0)) (i32.const -2147483648))
+(assert_return (invoke "i32.trunc_f64_s" (f64.const 2147483647.9999998)) (i32.const 2147483647))
```
binji, line 123: We may want to use the hex float format here to make it clearer what the expected float is (again to avoid text rep issues): `0x1.fffffffffffffp+30`

test/core/conversions.wast
binji: Yeah, to be fair I think we've done this for a few other tests too, so perhaps we should leave it as is here too. Personally, I'd prefer to use hex floats everywhere that we're not explicitly testing text parsing, but that should be done as a different change.

kripken:
Ok, given there are no other comments, landing as is, and leaving the question of possibly rewriting all tests to float hex format as a separate topic.

PR #1230, opened 2020-07-29
[test] Add "invalid section id" tests

chfast:
This adds explicit tests for "invalid section id". Previously there was only single test for it in custom.wast.

Some of the test cases are designed in a way to reproduce issues of parsing section ids as LEB128 encoded values.

This reproduces the bug in [wabt](https://github.com/WebAssembly/wabt) fixed by https://github.com/WebAssembly/wabt/pull/1501.

The bug has been discovered by fuzzing wabt wasm validation against [Fizzy](https://github.com/wasmx/fizzy).

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -44,6 +44,13 @@
 (assert_malformed (module binary "\00asm\00\00\01\00") "unknown binary version")
 (assert_malformed (module binary "\00asm\00\00\00\01") "unknown binary version")
 
+;; Invalid section id.
+(assert_malformed (module binary "\00asm" "\01\00\00\00" "\0c\00") "invalid section id")
+(assert_malformed (module binary "\00asm" "\01\00\00\00" "\7f\00") "invalid section id")
+(assert_malformed (module binary "\00asm" "\01\00\00\00" "\80\00\01\00") "invalid section id")
+(assert_malformed (module binary "\00asm" "\01\00\00\00" "\81\00\01\00") "invalid section id")
+(assert_malformed (module binary "\00asm" "\01\00\00\00" "\ff\00\01\00") "invalid section id")
+
 
 ;; call_indirect reserved byte equal to zero.
 (assert_malformed


chfast:
I'm not sure if you noticed, but I submitted this PR to `wg_v1` branch. Is it ok to submit more tests to this branch?

rossberg:
Ah, no, I didn't notice, thanks for pointing out. We probably shouldn't do that before some procedural discussion. In general, you should use master. I'll revert this for now, please create a new PR for master.

chfast:
Can we/you have this "procedural discussion" sooner than later. I cannot use master branch in my downstream project so I'm also little interested in submitting PRs to master as I will have to maintain my own branch of "wasm 1.0" anyway.

binji:
I'm tempted to say that adding new tests for existing behavior to older versions should be OK. But yes, we should discuss this. I'll add it to the CG meeting agenda.

binji:
Added an agenda item here: https://github.com/WebAssembly/meetings/pull/610

Ms2ger:
@binji I won't be able to attend the meeting, but strongly agree that we should make sure that any improvements happen in master too.

chfast:
> @chfast If we decide to allow changes to `wg_v1`, I still think the preferred way to do it would be to merge to `master`, then cherry-pick that PR to `wg_v1`, rather than submit a PR directly to the `wg_v1` branch.

I started submitting tests to master for now, but this is not generic solution. Extensions in master will make previously invalid wasm binaries valid. So I cannot submit tests for such wasm binaries as "invalid in wasm 1.0".

binji:
@chfast That's a good point, there will be some tests that will not be possible to test on master. In those cases we may want to commit directly to a branch directly.

chfast:
I can see in the meeting notes that backporting tests to v1 branch has been agreed on. Is the branch the `wg_v1`?
I will start some of this work as soon as someone pushes the button to accept me in the WebAssembly Community Group, and any other paper work needed to be done after.

PR #1234, opened 2020-07-30
[test] Add "malformed limits flag" tests

chfast:
The valid values are `00` and `01`. These tests checks if other values are reported as malformed.

The error messages reported by the interpreter are suspicious, so please double-check the tests and the interpreter implementation (I cannot ready OCaml).

This corresponds to [wabt](https://github.com/WebAssembly/wabt) bug https://github.com/WebAssembly/wabt/pull/1505, detected by fuzzing against [Fizzy](https://github.com/wasmx/fizzy).

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -510,6 +510,37 @@
   "unexpected end of section or function"
 )
 
+;; Malformed table limits flag
+(assert_malformed
+  (module binary
+      "\00asm" "\01\00\00\00"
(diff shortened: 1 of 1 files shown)


chfast:
@rossberg, Did you check why such error messages are reported? Is this expected?

rossberg:
The interpreter simply decodes a bool byte as an u1, in order to restrict its value range to 0-1. So the error messages consider it a number that's too large.

PR #1235, opened 2020-08-05
[spec/interpreter/test] Use LEB128 for saturating truncation opcodes

rossberg:
As agreed upon in yesterdays CG meeting.


interpreter/exec/int.ml
```diff
@@ -221,6 +223,9 @@ struct
   let ge_s x y = x >= y
   let ge_u x y = cmp_u x (>=) y
 
+  let to_int_s = Rep.to_int
+  let to_int_u i = Rep.to_int i land (Rep.to_int Rep.max_int lsl 1) lor 1
```
binji, line 227: I'm a bit surprised this isn't implemented anywhere else... or is it that the other implementations depend on the int module?

rossberg: Yeah, so was I. I think we never needed this for plain int.

rossberg:
Hm, what's the issue with Travis, and why don't I see a "Restart Build" button?

ngzhian:
I had a similar issue with "Restart Build" not showing up, https://stackoverflow.com/a/17624403 seems like I didn't have the right perms or something. In the end I pushed an empty commit, then in the "Rebase and squash" i just removed the description of the empty commit.

rossberg:
@ngzhian, the Internet suggested closing & reopening the PR as the simplest workaround.

PR #1241, opened 2020-08-28
Add python script to generate index-instructions

binji:
The index-instructions table is particularly annoying to update, since
and it requires the tables to line up properly, so it always creates
merge conflicts. This CL adds a Python script that generates it instead.


document/core/appendix/gen-index-instructions.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+def MathWrap(s, default=None):
+    if s is None:
+        if default is not None:
+            return default
+        else:
+            return ''
+    else:
+        return ':math:`{}`'.format(s)
```
sbc100, line 32: If we are using python3 should we start using f-strings?  `{s}`

document/core/appendix/gen-index-instructions.py
```diff
@@ -0,0 +1,286 @@
...
+COLUMNS = [
+    'Instruction',
+    'Binary Opcode',
+    'Type',
+    'Validation',
+    'Execution',
+]
+
+
+def MathWrap(s, default=None):
```
sbc100, line 25: How about `default=''`?

document/core/appendix/gen-index-instructions.py
```diff
@@ -0,0 +1,286 @@
+#!/usr/bin/env python3
```
sbc100, line 1: Maybe add a comment about why this was added rather than directly maintaining the rst file?

document/core/appendix/index-instructions.rst
```diff
@@ -4,212 +4,212 @@
 Index of Instructions
```
sbc100, line 4: Add an "DO NOT EDIT" comment?

document/core/appendix/gen-index-instructions.py
binji: That makes us require python 3.6, right? I guess that's OK. Do we use them in other Python scripts in WebAssembly org?

ngzhian: [Bikeshed requires 3.7](https://github.com/tabatkins/bikeshed/issues/321) so the version shouldn't be a problem.

document/core/appendix/gen-index-instructions.py
binji: Done.

document/core/appendix/gen-index-instructions.py
binji: Done.

document/core/appendix/index-instructions.rst
binji: Done.

document/core/appendix/gen-index-instructions.py
binji: OK, updated to use f-strings.

rossberg:
Yay, more hacks around markdown. :) But why not invoke this in the Makefile and remove the generated file from the repo?

binji:
Sounds good to me. Anyone else want to make this fix? :smile:

PR #1242, opened 2020-08-28
Merge bulk-memory-operations proposal into spec

binji:
See the bulk-memory-operations proposal here:

https://github.com/WebAssembly/bulk-memory-operations


rossberg:
Hm, we have to be careful to not mess things up. There were plenty of gaps, conflicts, and moved definitions between this repo and the reference types one, all of which I carefully filled in and resolved (I hope) in the reftypes repo. As a consequence, that repo subsumes this one.

Since we need to land both proposals anyway (the spec in this one is incomplete as is), I would strongly recommend that we simply merge reftypes to master in order to land both proposals. Otherwise we'll have to redo all the manual conflict resolution (because the merge history is lost after a squash), which is super error-prone and _a lot_ of work.

More precisely, I suggest to proceed as follows:

1. Merge recent changes from master into bulk.
2. Merge recent changes from bulk (including master) into reftypes.
3. Merge reftypes into master.

Does that make sense?

binji:
Agreed, my plan was not to merge this directly, but to show what changes would be merged independently. Then only merge the reference types branch. But perhaps that's just extra busywork for reviewers and we should instead just skip this one.

aleee1333 (approved):

rossberg:
This is obsolete now.

PR #1259, opened 2020-10-26
[spec] automate instruction index rebuild

rossberg:
Leaving the generated file in the tree for now.


Ms2ger:
Should we have CI to check that the checked-in file remains up to date?

binji (approved):

rossberg:
@Ms2ger, sounds like a good idea, please feel free to add that.

PR #1267, opened 2020-12-14
[test] Add invalid cases with globals access

gumb0:
This adds some cases with `global.get`/`global.set` with invalid index and a case with `global.set` for immutable imported global.


rossberg (approved):
Sorry for the delay.

PR #1268, opened 2020-12-14
[test] Add invalid cases with br_table type mismatch

gumb0:


test/core/br_table.wast
```diff
@@ -1552,6 +1562,35 @@
...
 )
 
+(assert_invalid
+  (module
+    (func (param i32) (result i32)
+      (loop (result i32)
+        (block (result i32)
+          i32.const 1
+          local.get 0
+          br_table 0 1
```
rossberg, line 1572: Nit: use folded form here as rest of file?

test/core/br_table.wast
```diff
@@ -1552,6 +1562,35 @@
...
+  "type mismatch"
+)
+(assert_invalid
+  (module
+    (func (param i32) (result i32)
+      (block (result i32)
+        (loop (result i32)
+          i32.const 1
+          local.get 0
+          br_table 0 1
```
rossberg, line 1586: Same here.

test/core/br_table.wast
gumb0: Changed.

test/core/br_table.wast
gumb0: This is fixed, too.

rossberg (approved):
Thanks!

PR #1270, opened 2021-01-08
[js-api] Better null handling in table.get/set.

Ms2ger:


document/js-api/index.bs
```diff
@@ -792,14 +792,15 @@ Each {{Table}} object has a \[[Table]] internal slot, which is a [=table address
     1. Let |store| be the [=surrounding agent=]'s [=associated store=].
     1. Let |result| be [=table_read=](|store|, |tableaddr|, |index|).
     1. If |result| is [=error=], throw a {{RangeError}} exception.
+    1. If |result| is ε, return null.
```
Ms2ger, line 795: @ioannad pointed out that we haven't used ε before in the js api spec, but it doesn't seem like we have any other case that needs to handle optional values.

ioannad (approved):
Since there is no other place to use ε, lgtm.

PR #1272, opened 2021-01-11
[test] Add cases checking maximum allowed number of locals

gumb0:
This adds a number of cases that check that number of locals + number of parameters is required to be < 2^32, but allowed to be = 2^32 - 1.

Note: The spec nowhere explicitly states that this sum is < 2^32, but I think it is at least implied from the fact that `localidx` is defined to be `u32`, please correct me if I'm wrong.


test/core/binary.wast
```diff
@@ -350,6 +350,62 @@
   "zero flag expected"
 )
 
+;; Local count can be 2^32.
+(module binary
```
gumb0, line 354: WABT currently fails this test, see https://github.com/WebAssembly/wabt/issues/1593

test/core/binary.wast
```diff
@@ -350,7 +350,63 @@
   "zero flag expected"
 )
 
-;; No more than 2^32 locals.
+;; Local count can be 2^32-1.
+(module binary
```
gumb0, line 354: WABT currently fails this test, see WebAssembly/wabt#1593

rossberg:
Unfortunately, it would be problematic to put test like these into the test suite. The spec explicitly allows implementations to impose certain implementation limits (see the Implementation appendix). So an implementation is its right to fail a test like this that uses excessive numbers. And in practice, most will.

> The spec nowhere explicitly states that this sum is < 2^32, but I think it is at least implied from the fact that localidx is defined to be u32, please correct me if I'm wrong.

That's right.

gumb0:
> Unfortunately, it would be problematic to put test like these into the test suite. The spec explicitly allows implementations to impose certain implementation limits (see the Implementation appendix). So an implementation is its right to fail a test like this that uses excessive numbers. And in practice, most will.

I see, I will leave here only the cases with >= 2^32 locals, that must always fail.

rossberg (approved):

rossberg (changes requested):
Actually, it looks like the test fails.

gumb0:
> Actually, it looks like the test fails.

The expected error message was wrong in the first test, I fixed it.

But now it looks like running the second test (2^32-1 locals + 2 parameters) goes out of memory, I'm not sure how to proceed with this.

rossberg:
Yeah, right, that is another case where implementations are expected to impose resource limits. So I think you'll have to remove that one as well.

Maybe replace it with another test where there are locals of multiple types, and their total exceeds 2^32.

gumb0:
> Yeah, right, that is another case where implementations are expected to impose resource limits. So I think you'll have to remove that one as well.

My guess is that the interpreter doesn't report an error and tries to allocate all of 2^32-1 locals (similar to my earlier "valid" tests here), can this be an indication that such module is not considered invalid by the spec? (but would be by most implementations)

> 
> Maybe replace it with another test where there are locals of multiple types, and their total exceeds 2^32.

That already exists for two types here:
https://github.com/WebAssembly/spec/blob/0aecf9908a049d88cfdccbe9afafd71b23d20b22/test/core/binary.wast#L380-L381
I can add another test with more types.

rossberg (approved):
> My guess is that the interpreter doesn't report an error and tries to allocate all of 2^32-1 locals (similar to my earlier "valid" tests here), can this be an indication that such module is not considered invalid by the spec? (but would be by most implementations)

It is not technically invalid, but not guaranteed to be accepted either, per the [spec Appendix](https://webassembly.github.io/spec/core/appendix/implementation.html). That section makes no attempt to specify how implementations handle resource exhaustion, since that is very hard to catch reliably in many cases.

> > Maybe replace it with another test where there are locals of multiple types, and their total exceeds 2^32.
> 
> That already exists for two types here:

Ah, sorry, yes.

PR #1288, opened 2021-02-17
Add testing for globals initialized by othert (imported) globals

sbc100:
Split out from https://github.com/WebAssembly/extended-const/pull/2


rossberg (approved):
Thanks!

PR #1289, opened 2021-02-21
[test] Add tests for functions sharing same implementation

atdrendel:
This pull request adds a test to verify WebAssembly interpreters correctly handle the case where multiple exported functions share a single implementation. I've run into this problem when trying to use Libsodium with some Wasm interpreters. Libsodium exposes a number of different functions that wrap private constants, some of which are equal. These functions are all compiled down into the same WebAssembly function implementation with different exported names.

### DIFF
--- a/test/core/exports.wast
+++ b/test/core/exports.wast
@@ -25,6 +25,16 @@
 (module $Other1)
 (assert_return (invoke $Func "e" (i32.const 42)) (i32.const 43))
 
+(module
+  (type (;0;) (func (result i32)))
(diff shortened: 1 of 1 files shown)


atdrendel:
I've requested access to the WebAssembly Community Group.

rossberg (approved):

atdrendel:
@rossberg My request to join the W3C WebAssembly group has been approved. Please let me know if there is anything you want me to change in this pull request before merging it.

PR #1290, opened 2021-02-26
Add global_i64 to js harness

ngzhian:
PR #1288 is the first use of global_i64, but the generated JS did not
yet define this global, which causes the JS test to fail to link.


Ms2ger:
Doesn't this need to be a bigint?

ngzhian:
Good point, fixed, thanks!

Ms2ger (approved):
Verified locally. Thanks.

PR #1291, opened 2021-02-27
[test] Test cases with imports the same name as exports

axic:
While integrating [wasm3](https://github.com/wasm3/wasm3) into [fizzy](https://github.com/wasmx/fizzy)'s benchmarking suite, we found some issues with wasm3. Namely it did not make a distinction between imported and exported functions where the name matches. Since wasm3 passes the spec tests, I think such cases are not covered. The original test case can be seen at https://github.com/wasm3/wasm3/issues/143.

### DIFF
--- a/test/core/imports.wast
+++ b/test/core/imports.wast
@@ -96,6 +96,26 @@
   "unknown type"
 )
 
+;; Export sharing name with import
+(module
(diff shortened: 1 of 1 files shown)


rossberg (approved):

axic:
Thanks!

PR #1293, opened 2021-03-05
[test] Check that memory is exported/imported with correct min limit after grow

gumb0:
I made this to check my understanding of the [external memory definition in the spec](https://webassembly.github.io/spec/core/exec/modules.html#xref-exec-runtime-syntax-externval-mathsf-mem-a): 

image

IIUC it requires imported memory to have `min` limit equal to it's currently allocated size, which means it can be updated after grow, and that has implications on matching this memory limits against import definition in another module.

The test basically creates a module with `(memory 1)`, grows it to 2 pages, then tries to import it into a module with `(memory 2)`.
Then repeats it another time for the memory that is both imported and exported.

(It [fails on WABT](https://github.com/WebAssembly/wabt/issues/1626).)

### DIFF
--- a/test/core/imports.wast
+++ b/test/core/imports.wast
@@ -565,6 +565,26 @@
 (assert_return (invoke "grow" (i32.const 1)) (i32.const -1))
 (assert_return (invoke "grow" (i32.const 0)) (i32.const 2))
 
+(module $Mgm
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Thanks for this.

gumb0:
@rossberg Is this going to be merged?

PR #1294, opened 2021-03-10
Small fixes in build instructions, fix warnings

ngzhian:
Since the build instructions suggest using pipenv, we should
consistently use pipenv to install the packages.

Also remove some warnings about unavailable static files.

Fix a Sphinx deprecation warning.

### DIFF
--- a/document/core/util/mathdef.py
+++ b/document/core/util/mathdef.py
@@ -1,5 +1,5 @@
 from sphinx.ext.mathbase import math
-from sphinx.ext.mathbase import MathDirective
+from sphinx.directives.patches import MathDirective
 from sphinx.util.texescape import tex_replace_map
 from sphinx.writers.html5 import HTML5Translator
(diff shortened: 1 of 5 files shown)


document/core/util/mathdef.py
```diff
@@ -1,5 +1,5 @@
 from sphinx.ext.mathbase import math
-from sphinx.ext.mathbase import MathDirective
+from sphinx.directives.patches import MathDirective
```
rossberg, line 2: Would this happen to make us compatible with newer versions of Sphinx?

rossberg (approved):
Great, thanks!

document/core/util/mathdef.py
ngzhian: Yes :) I have https://github.com/ngzhian/spec/tree/upgrade-sphinx prepped and will send it out soon.

PR #1295, opened 2021-03-11
[interpreter] Functionality for encoding custom sections

rossberg:
This adds functionality for en/decoding user defined custom sections to the En/Decode modules. Doesn't change the interpreter, but provides useful functionality for the Wasm lib build.

Also, rename `stack_type` to `result_type` to match spec.

The large diff in Encode.ml is misleading, since most of it just changed indentation.


rossberg:
Taking the liberty to land this, since it doesn't affect anything observable.

PR #1298, opened 2021-03-11
Upgrade Sphinx to 3.5.2

ngzhian:
The html_add_permalinks configuration was changed in 3.5
(https://www.sphinx-doc.org/en/3.x/usage/configuration.html#confval-html_add_permalinks),
we don't want any permalink for the heading (that's the existing
behavior).

Fix up bikeshed_fixup.py, we were not correctly replacing the spurious
"navigation section" and also the Unicode reference.

Most of the changes in the output is:

- adding viewport meta tag (in Sphinx#7695, since 3.1.0)
- sorting order of css files in html
- upgrade mathjax to 2.7.7

### DIFF
--- a/document/core/util/bikeshed_fixup.py
+++ b/document/core/util/bikeshed_fixup.py
@@ -41,8 +41,8 @@ def Main():
     <div class="related" role="navigation" aria-label="related navigation">
       <h3>Navigation</h3>
       <ul>
-        <li class="nav-item nav-item-0"><a href="#">WebAssembly 1.0</a> &#187;</li>
-        <li class="nav-item nav-item-this"><a href="">WebAssembly 1.0</a></li> 
+        <li class="nav-item nav-item-0"><a href="#">WebAssembly 1.1</a> &#187;</li>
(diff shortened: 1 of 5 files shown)


document/core/util/bikeshed_fixup.py
```diff
@@ -41,13 +41,14 @@ def Main():
     <div class="related" role="navigation" aria-label="related navigation">
       <h3>Navigation</h3>
       <ul>
-        <li class="nav-item nav-item-0"><a href="index.html#document-index">WebAssembly 1.0</a> &#187;</li> 
+        <li class="nav-item nav-item-0"><a href="#">WebAssembly 1.0</a> &#187;</li>
+        <li class="nav-item nav-item-this"><a href="">WebAssembly 1.0</a></li>
```
rossberg, line 45: While you're here, could you change this to 1.1, to match the version given on the regular front page?

rossberg (approved):
Fantastic, thanks a lot!

document/core/util/bikeshed_fixup.py
ngzhian: Done!

PR #1301, opened 2021-03-30
Escape parentheses in echo line

radekdoulik:
They were breaking the command block, so `make.bat help` output was:

    > ./make.bat help
      help       to see more options


Ms2ger (approved):
I don't know anything about bat files, but this seems plausible. Will give others the chance to comment before merging.

rossberg (approved):
Thanks!

PR #1305, opened 2021-04-07
[spec] Fix oversight in reversal of br_table validation

rossberg:
WebAssembly/reference-types#116 re-reversed the validation of br_table, but only did so in tests, interpreter, and the spec's algorithm appendix. It somehow missed the actual validation rule in the spec. This fixes that oversight.


rossberg:
@gahaas, ping.

gahaas (approved):
LGTM

PR #1308, opened 2021-04-08
[spec] Move all repo-specifics to conf.py

rossberg:
This avoids having to patch `index.rst` and `util/macros.def` when forking a proposal repo. All temporary changes are now defs in `conf.py`.

Will adapt proposal howto accordingly (see WebAssembly/proposals#109).


document/core/conf.py
```diff
@@ -477,6 +486,8 @@
 
 # Macros
 rst_prolog = """
+.. |repolink| replace:: https://github.com/webassembly/""" + repo + """/issues/
```
tlively, line 489: Maybe `|issuelink|` instead, since this points to the issues rather than the main repo page?

tlively (approved):
Nice simplification 👍

document/core/conf.py
rossberg: Done. I originally had defined it as the actual repolink, but Sphinx substitutions are too braindead to express a link constructed by extension.

PR #1315, opened 2021-04-21
[spec] Fix side condition about fresh ids

### LINKED ISSUE #1313: Identifiers can seemingly be replaced with fresh identifiers
In [6.6.5.1](https://webassembly.github.io/spec/core/bikeshed/#abbreviations%E2%91%A0%E2%93%AA), the condition on the rewrite rule is (if `id`′ = `id`<sup>?</sup> !​= ϵ ∨ `id`′ fresh), which suggests that as long as `id`′ is a fresh identifier, the rule can be applied. So even if the declaration already has an identifier, that identifier can be replaced with a fresh one.

That seems wrong and unintentional, as it would break any references to the original identifier. The right side of the condition should probably be changed to something like (`id`′ fresh ∧ `id`<sup>?</sup> = ϵ).

Same thing applies to the many similar rewrite rules for other types of declarations.

rossberg:
Fixes #1313.

### DIFF
--- a/document/core/text/modules.rst
+++ b/document/core/text/modules.rst
@@ -245,7 +245,7 @@ Functions can be defined as :ref:`imports <text-import>` or :ref:`exports <text-
        \text{(}~\text{export}~~\Tname~~\text{(}~\text{func}~~\Tid'~\text{)}~\text{)}~~
        \text{(}~\text{func}~~\Tid'~~\dots~\text{)}
        \\ & \qquad\qquad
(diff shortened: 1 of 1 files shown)


rossberg:
@dtig, friendly ping.

rossberg:
@SpecLad, could you verify the fixes and perhaps approve this PR?

SpecLad (approved):
Sure. Looks good to me.

rossberg:
Thanks!

PR #1325, opened 2021-05-11
Fix the Editor metadata format

jyasskin:
Names and affiliations are separated by a comma: https://tabatkins.github.io/bikeshed/#metadata-editor


Ms2ger (approved):

Ms2ger:
Looks like https://github.com/WebAssembly/spec/blob/master/document/web-api/index.bs has the same issue, btw.

PR #1329, opened 2021-05-29
Factor unreachability tests out of select and br_table

titzer:
This PR factors out tests that deal with validating unreachable code in the presence of the bottom type. Specifically, both `select` and `br_table` have tricky corner cases that probably deserve their own file.

The motivation for this PR is merely to factor out these tricky corner cases into their own test file in anticipation of any more corner cases that are found or any future changes to how unreachable code is validated. This makes it easy to pinpoint the changes introduced and to diagnose the problem in implementations.

### DIFF
--- a/test/core/br_table.wast
+++ b/test/core/br_table.wast
@@ -1251,17 +1251,6 @@
     )
   )
 
-  (func (export "meet-bottom")
-    (block (result f64)
-      (block (result f32)
-        (unreachable)
-        (br_table 0 1 1 (i32.const 1))
-      )
-      (drop)
-      (f64.const 0)
-    )
-    (drop)
-  )
 )
 
 (assert_return (invoke "type-i32"))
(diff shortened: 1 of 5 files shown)


rossberg (commented):
Can you move the assert_invalid tests to `unreached-invalid.wast`, where similar tests live, and rename this file to `unreached-valid.wast`?

rossberg:
Also, it seems you need to register your current email with the CG to get through the IPR check.

titzer:
> Can you move the assert_invalid tests to `unreached-invalid.wast`, where similar tests already live, and rename this file to `unreached-valid.wast`?

Done.



> Also, it seems you need to register your current email with the CG to get through the IPR check.

Yeah, I think I need to sign up to the W3C again :(

rossberg (approved):

PR #1336, opened 2021-06-23
[spec] $canon_N = 2^{signif(N)} - 1$;

mengsuenyan:
[IEEE-754](https://en.wikipedia.org/wiki/IEEE_754)
[NaN](https://en.wikipedia.org/wiki/NaN);


rossberg:
I don't think this is correct, see the prose explanation above.

PR #1346, opened 2021-07-18
Test order of execution

### LINKED ISSUE #1343: Test more side-effects/order of execution
There should be a side-effects test-suite. In particular, trapping is a side-effect.

Code like this:

```
i32.const 0 i32.const 0 i32.div_s i32.const 1 global.set 0 local.set 0
```

breaks on an implementation that currently passes every other test. (the test should set the global to 0, call this code, and assert that the global is still 0.)

More importantly, this example is a fundamental core semantic. That's why it should be part of the spec test.

SoniEx2:
Fixes #1343.

### DIFF
--- a/test/core/order_of_execution.wast
+++ b/test/core/order_of_execution.wast
@@ -1,0 +1,31 @@
+;; Test order of execution/optimizer bait
+
+(module
+  (memory 0)
+  (global $a (mut i32) (i32.const 0))
+
+  (func (export "get-a") (result i32) (global.get $a))
+
+  (func (export "div-by-zero") (local i32)
+    (i32.const 0) (i32.const 0) (i32.div_s) (i32.const 1) (global.set $a) (local.set 0)
+  )
+
+  (func (export "memory-oob-read") (local i32)
+    (i32.load (i32.const 0)) (i32.const 2) (global.set $a) (local.set 0)
+  )
+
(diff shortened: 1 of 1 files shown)


test/core/order_of_execution.wast
```diff
@@ -0,0 +1,31 @@
...
+  (func $set_by_call (param i32)
+    (local.get 0) (global.set $a)
+  )
+
+  (func (export "trapping-div-trapping-call") (local i32)
+    (i32.const 0) (i32.const 0) (i32.div_s) (call $set_by_call (i32.const 3)) (local.set 0)
+  )
+)
+
+(assert_trap (invoke "div-by-zero") "integer divide by zero")
```
ngzhian, line 26: LGTM, maybe a comment here describing what we want these tests to catch? Along the lines of
"Traps have side effect and instructions follow them should not run, optimizations must preserve this side-effect"

rossberg (changes requested):
Except for a few legacy exceptions, the test suite is structured around individual constructs (or groups thereof). Instead of introducing a new file with a random selection of things, can you put each test into the appropriate file for the affected construct?

rossberg:
@SoniEx2, this PR is stale, do you still intend to land it?

SoniEx2:
kinda haven't had the time to work on this anymore, sorry.

PR #1350, opened 2021-07-30
[spec] Fix type notation of `IF` instruction

dulanov:
`IF` is like two `BLOCKs` instructions guarded by a condition, so maybe it can be more obvious to see it explicitly in it's type notation.


rossberg (commented):
Thanks! However, note that this is a generated file. You'll need to change `gen-index-instructions.py` and rerun it.

rossberg (commented):
Did you check in the changes to the generated file? It seems like the spacing is still off (RST/Sphinx is rather anal about that, which is why we have the generator).

dulanov:
Ah now i see it, just forgot to include.

PR #1352, opened 2021-08-16
[web-api] Limit cross-origin sharing of Wasm modules.

### LINKED ISSUE #1303: Deprecate cross-origin Module sharing
Currently, modules can be serialized and shared with other context in the same agent cluster (see spec [here](https://webassembly.github.io/spec/web-api/index.html#serialization)). Agent clusters are scoped to sites instead of origins by default, which means that modules can be shared with a same-site but cross-origin contexts. To improve mitigations against attacks like Spectre, we are aiming to make agent clusters scoped to origins instead. To do this we must deprecate document.domain (see [plan](https://github.com/mikewest/deprecating-document-domain)) and sharing of WASM modules cross-origin. The latter is used by 0.06% of web pages at best according to our metrics, so we could proceed with deprecation right away. The plan would be to store the origin as well as the agent cluster in the serialization step, and to check it when passing the module.

dtig:
This was discussed, and voted on in the June 22nd meeting ([notes](https://github.com/WebAssembly/meetings/blob/main/main/2021/CG-06-22.md)). The suggested solution was to limit sharing modules across origins. There is some discussion in the linked issue  #1303, but it doesn't look like the HTML spec has a way to enforce an origin check in the serialization infrastructure.    

This PR aims to store the origin as well as the agent cluster, and throw if there has been an attempt to post message across origin, the text right doesn't handle opaque origins as they are `null` when serialized. Another option is to be vague and include a [same-origin check](https://html.spec.whatwg.org/#same-origin), but digging into it more, it's not clear how this would be implemented. Opening this PR to gather feedback, I'm also quite unfamiliar with the HTML spec, so links to existing infrastructure to do this correctly appreciated.

Closes #1303.

### DIFF
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -69,7 +69,9 @@ urlPrefix: https://webassembly.github.io/spec/js-api/; spec: WASMJS
         text: asynchronously compile a webassembly module; url: #asynchronously-compile-a-webassembly-module
         text: instantiate a promise of a module; url: #instantiate-a-promise-of-a-module
         text: Exported Function; url: #exported-function
+url:https://html.spec.whatwg.org/#ascii-serialisation-of-an-origin;text:origin-serialization;type:dfn;spec:HTML
 url:https://html.spec.whatwg.org/#cors-same-origin;text:CORS-same-origin;type:dfn;spec:HTML
+url:https://html.spec.whatwg.org/#origin;text:origin;type:dfn;spec:HTML
 url:https://fetch.spec.whatwg.org/#concept-body-consume-body;text:consume body;type:dfn;spec:FETCH
 url:https://w3c.github.io/webappsec-secure-contexts/#environment-settings-object-contextually-secure; text:contextually secure; type: dfn; spec: SECURECONTEXTS
 </pre>
@@ -150,6 +152,8 @@ The [=serialization steps=], given |value|, |serialized|, and |forStorage|, are:
     1. Set |serialized|.\[[Bytes]] to the [=sub-serialization=] of |value|.\[[Bytes]].
 
     1. Set |serialized|.\[[AgentCluster]] to the [=current Realm=]'s corresponding [=agent cluster=].
+    
+    1. Set |serialized|.\[[Origin]] to be the [=origin-serialization=] of the user agent's [=origin=].
 
 The [=deserialization steps=], given |serialized| and |value|, are:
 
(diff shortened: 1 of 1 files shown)


Ms2ger:
I'll try to review this week

document/web-api/index.bs
```diff
@@ -158,6 +162,8 @@ The [=deserialization steps=], given |serialized| and |value|, are:
     1. Set |value|.\[[Bytes]] to |bytes|.
 
     1. If |targetRealm|'s corresponding [=agent cluster=] is not |serialized|.\[[AgentCluster]], then throw a "<a exception>DataCloneError</a>" {{DOMException}}.
+    
+    1. If the [=origin-sereialization=] of the target user agent's [=origin=] is not |serialized|.\[[Origin]], then throw a "<a exception>DataCloneError</a>" {{DOMException}}.
```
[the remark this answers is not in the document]
gahaas: Does `throw a DataCloneError` here mean that if you postmessage a wasm module, and a `DataCloneError` is thrown at the receiver, then the receiver gets a `messageerror` event instead of an `onmessage` event?

dtig:
@Ms2ger Friendly ping for a review.

document/web-api/index.bs
```diff
@@ -150,6 +152,8 @@ The [=serialization steps=], given |value|, |serialized|, and |forStorage|, are:
     1. Set |serialized|.\[[Bytes]] to the [=sub-serialization=] of |value|.\[[Bytes]].
 
     1. Set |serialized|.\[[AgentCluster]] to the [=current Realm=]'s corresponding [=agent cluster=].
+    
+    1. Set |serialized|.\[[Origin]] to be the [=origin-serialization=] of the user agent's [=origin=].
```
Ms2ger, line 156: A user agent doesn't have an origin - most likely this should be the [origin](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-origin) of the [relevant settings object](https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object) for *value*.

Ms2ger, line 156: I'm also wondering if using the serialization is correct - this would perhaps allow transferring between two opaque origins. I'm not sure if we can store the origin itself here - @annevk?

Ms2ger (changes requested):
Sorry for the delay!

annevk:
I'd still like a reply to my question in https://github.com/WebAssembly/spec/issues/1303#issuecomment-829087678.

I read the minutes and it's clear my point there was not considered or addressed as it's a rather high-level discussion. And the minutes also seem to contain something that illustrates a rather fundamental misunderstanding of how this works:

> Change Wasm spec to mention origins instead of agent clusters

As to @Ms2ger's question, normally an opaque origin would also imply a different agent cluster (and this PR keeps the agent cluster in tact, as it should), but there is this edge case that is still unsolved: https://github.com/whatwg/html/issues/5254.

rossberg:
@dtig, this PR is stale, what's the status?

rossberg:
@dtig, any progress on this PR, or should we close it?

dtig:
I'm closing this for now, because without infrastructure that the Wasm web spec can use, and better consensus on the checks themselves, this is out of scope for the Wasm CG. Will reopen if anything changes. Sorry for the slow response, I was unsure about the status of the various parts of the deprecations that are connected to this, and had some trouble with parsing the information across different parts of the spec.

PR #1354, opened 2021-08-20
[interpreter] Fix quadratic behavior when stepping in deeply-nested scopes.

cfallin:
Hi, spec-interpreter maintainers! We've found some quadratic
behavior in the spec interpreter -- and I'm curious how palatable
this fix will be. Happy to discuss any alternative approaches as well.

The reference interpreter defines a `step` function on the program
configuration. This configuration embeds the current Wasm callstack and
block (label) stack, representing the state of execution with a tree of
terms. At each `step`, execution advances by one transition, e.g. one
instruction or one control-flow transfer.

When a `step` occurs on a configuration that has nested callstack and
block frames, the interpreter recurses structurally into the
configuration, eventually stepping in the innermost scope, and then
returning and rebuilding the entire configuration to the root of the
value. Unfortunately, this means that for deeply nested execution, the
interpreter can take `O(n)` time (where `n` is the depth of the stack of
all call and block frames) to execute a single instruction.

In bytecodealliance/wasmtime#3186, we found (via fuzzing against the
reference interpreter) a simple program that causes quadratic behavior
in the interpreter, taking many seconds to run despite a relatively
modest number of steps (1000 recursive calls nested in 18 block
levels per call).

This change to the interpreter adds a `step_many` function that resolves
the quadratic behavior, executing the above test program nearly
instantly. The key idea is to step an inner configuration (e.g., within
a block scope or function scope) many times until it cannot be stepped
"locally" (without the outer context). This prevents the unnecessary
exit and reentry to a deep recursive level (and corresponding
reallocation of the entire configuration tree) on every instruction
step.

### DIFF
--- a/interpreter/exec/eval.ml
+++ b/interpreter/exec/eval.ml
@@ -144,7 +144,34 @@ let elem_oob frame x i n =
   I64.gt_u (I64.add (I64_convert.extend_i32_u i) (I64_convert.extend_i32_u n))
     (I64.of_int_u (List.length !(elem frame.inst x)))
 
-let rec step (c : config) : config =
+(* Perform multiple steps at once, possibly with `code` that is part of a
+   surrounding context (block scope or call frame), until we cannot step without
+   the surrounding context. This is used to advance execution within a nested
+   context without rebuilding the entire configuration from the root, which is
+   necessary to avoid quadratic behavior (O(n) time tp step with n-deep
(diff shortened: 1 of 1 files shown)


cfallin:
Friendly ping -- @rossberg, are you the right person to tag here? Or someone else? Thanks!

PR #1359, opened 2021-09-15
Remove outdated table test

gahaas:
The spec allows a missing `value` parameter in `WebAssembly.Table.set`, see https://webassembly.github.io/spec/js-api/index.html#dom-table-set.  There is a test, however, that `WebAssembly.Table.set` throws if there is no `value` parameter. This PR deletes that test.


Ms2ger (approved):
Thanks, this reminds me that I need to update the reference types tests.

PR #1364, opened 2021-09-17
Add action to mirror main to master

ngzhian:
I created [main branch](https://github.com/WebAssembly/spec/tree/main) on the remote so this should work now.

I don't expect any downstream (forks/proposals) to have issues, since master will still exist. This mirroring will mirror any changes on main to master, so it will be kept up to date.

Follow-up to this will be to update places like https://github.com/WebAssembly/proposals/blob/master/howto.md to refer to main.

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -1,0 +1,17 @@
+on:
+  push:
+    branches:
+      - 'main'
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Thanks!

PR #1372, opened 2021-10-07
Improve parse error message and add null-byte in comments test

### LINKED ISSUE #1355: Bugs in test suite
There are a few bugs in the binary modules:

**binary.wast**
<pre>
  Line 48: (assert_malformed (module binary "\00asm" "\01\00\00\00" "\0c\00") "malformed section id")
    "\0c" is now a valid section id (data count section). It needs changing to e.g. "\0d".

  Various function length fields are wrong:
    Line 98:  "\07\00" should be "\08\00"
    Line 117: "\08\00" should be "\09\00"
    Line 135: "\09\00" should be "\0a\00"
    Line 153: "\0a\00" should be "\0b\00"

  The test starting at Line 943 ("1 br_table target declared, 2 given") is ill-conceived:
    how is the parser to know that "\02" is meant to be the break depth for default? In practice,
    it is interpreted as a block instruction, followed by an invalid block type "\0b". So the
    error message should be "invalid block type" or some such. I think this test should be removed.
  
    This test also has an invalid code section length at Line 950: "\0a\12\01" should be "\0a\13\01".

  Invalid error message at Line 992 ("Multiple start sections"):
    "junk after last section" should be "duplicate section" or some such.
</pre>
**data.wast**
<pre>
  Four modules omit the bitfield from the start of a data item.

    Lines 295,and 306:
      "\0b\06\01"                             ;; data section
    should be:
      "\0b\07\01\02"                          ;; data section

    Lines 320 and 339:
      "\0b\44\01"                             ;; data section
    should be:
      "\0b\45\01\02"                          ;; data section
</pre>
**comments.wast**
  A suggestion. There is a test (Line 39) that checks whether the parser can handle a null byte
  inside a block comment. But the test also includes CR-LF characters, so the end-of-comment delimiter
  is not on the same line as the null byte. This will save invalid implementations that treat
  the null byte as end-of-line. There should be a test that checks specifically for this, with
  a null byte followed immediately by an end-of-comment delimiter, without any intervening
  line breaks.

ngzhian:
Fixes #1355.


interpreter/binary/decode.ml
```diff
@@ -808,7 +808,7 @@ let module_ s =
   iterate custom_section s;
   let datas = data_section s in
   iterate custom_section s;
-  require (pos s = len s) s (len s) "junk after last section";
+  require (pos s = len s) s (len s) "duplicate section";
```
rossberg, line 811: Hm, I don't think this message is adequate. This can be triggered by arbitrary bytes, not just well-formed, repeated sections. Maybe change to `"unexpected content after last section"`?

rossberg (approved):

PR #1373, opened 2021-10-12
Add tests for using nan patterns outside of script

ngzhian:


test/core/f32.wast
```diff
@@ -2531,3 +2531,13 @@
...
+
+;; These float literal patterns are only allowed in scripts, not in text format.
+(assert_malformed
+  (module quote "(func (result f32) (f32.const nan:arithmetic))")
+  "unexpected token"
+)
+(assert_malformed
+  (module quote "(func (result f32) (f32.const nan:canonical))")
+  "unexpected token"
+)
```
sunfishcode, line 2543: Could you add a similar test to f64.wast that tests the same thing for the f64 type, to keep these two files in sync?

ngzhian: Oh yea good suggestion, done!

PR #1379, opened 2021-10-14
Add integer tests for using nan patterns outside of script

ngzhian:
Similar to 5893af5bce241fbb074d482a65706126c031b8ed but for integers,
this is also a type mismatch, but should first be caught at parse time.


rossberg (commented):
Why not avoid the type mismatch, to make the error more on point?

PR #1388, opened 2021-10-29
Only include ref links to index in multi-page html

ngzhian:
This allows us to drop some manual links removal in mathjax2katex.py.

### DIFF
--- a/document/core/index.rst
+++ b/document/core/index.rst
@@ -29,7 +29,11 @@ WebAssembly Specification
       appendix/index-instructions
       appendix/index-rules
 
-.. only:: html
+..
+   Only include these links when using (multi-page) html builder.
+   (The singlepage html builder is called builder_singlehtml.)
+
+.. only:: builder_html
 
    * :ref:`index-type`
    * :ref:`index-instr`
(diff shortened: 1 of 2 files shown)


rossberg (approved):

ericprud:
Here's my plan A for publishing the core spec
  https://www.w3.org/2021/11/wasm-stage/
It's multiple pages. It's generated from [SphinxToTr](https://github.com/w3c/sphinx-to-tr/) , which could possibly be used for other specs.

Would that structure obviate the need for this PR?

ngzhian:
I don't think so, from https://www.w3.org/2021/11/wasm-stage/ I see this in the middle of the page
Screen Shot 2021-11-02 at 10 16 51 AM

It looks out of place.

(Btw it looks like none of the links in the TOC work, e.g. https://www.w3.org/2021/11/wasm-stage/syntax/conventions.html#grammar-notation, I get a page that says "Sorry, Insufficient Access Privileges"

ericprud:
oops, i faild to set ACLs on the subdirs. index refs and search should all work now

PR #1390, opened 2021-11-02
Fix html_last_updated_fmt

ngzhian:
'%F' is not a supported format, see
https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/util/i18n.py#L133,
it is also not supported in Python's strftime/strptime
https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior.

It is however supported in strftime
(https://man7.org/linux/man-pages/man3/strftime.3.html) and defined to
be "%Y-%m-%d", which is supported by sphinx.


rossberg (approved):
Interesting.

Ms2ger (approved):

PR #1396, opened 2021-11-16
Remove use of physical equality on characters

hanshoglund:
The meaning of `!=` is [implementation-dependent](https://www.ocaml.org/api/Stdlib.html) in OCaml. This seems undesirable for a reference interpreter.

This PR switches to to `<>` which checks for structural non-equality rather than physical non-equlity.


rossberg (approved):
Yup, thanks!

Issue #1404, opened 2021-12-07
Missing test for a function that lacks an end marker?


sbc100:
I found this error condition was untested in wabt so I added test here:  https://github.com/WebAssembly/wabt/pull/1775

Presumably it would be good idea to upstream such tests here?

We also have another test for when a function doesn't report the correct body size (https://github.com/WebAssembly/wabt/blob/main/test/binary/bad-function-body-size.txt).

rossberg:
Yes, absolutely. You'll have to write them as binary tests, though.

PR #1405, opened 2021-12-10
Add tests for functions without end marker. NFC

sbc100:
Inspired by this downstream test in wabt: https://github.com/WebAssembly/wabt/pull/1775

Fixes: #1404

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -368,6 +368,62 @@
   "integer too large"
 )
 
+;; Function with missing end marker (between two functions)
+(assert_malformed
+  (module binary
+    "\00asm" "\01\00\00\00"
+    "\01\04\01\60\00\00"       ;; Type section: 1 type
+    "\03\03\02\00\00"          ;; Function section: 2 functions
+    "\0a\0c\02"                ;; Code section: 2 functions
+    ;; function 0
+    "\04\00"                   ;; Function size and local type count
+    "\41\01"                   ;; i32.const 1
+    "\1a"                      ;; drop
+    ;; Missing end marker here
+    ;; function 1
+    "\05\00"                   ;; Function size and local type count
+    "\41\01"                   ;; i32.const 1
+    "\1a"                      ;; drop
+    "\0b"                      ;; end
+  )
+  "END opcode expected"
+)
+
+;; Function with missing end marker (at EOF)
+(assert_malformed
+  (module binary
+    "\00asm" "\01\00\00\00"
+    "\01\04\01\60\00\00"       ;; Type section: 1 type
+    "\03\02\01\00"             ;; Function section: 1 function
+    "\0a\06\01"                ;; Code section: 1 function
+    ;; function 0
+    "\04\00"                   ;; Function size and local type count
+    "\41\01"                   ;; i32.const 1
+    "\1a"                      ;; drop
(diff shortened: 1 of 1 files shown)


sbc100:
I wonder if "section size mismatch" is wrong error here, but its the one that the interpreter currently generates.

rossberg (approved):
Well, the code section in this example is followed by a data section, which starts with the same opcode as END (I assume that's intentional?). So the decoder assumes the prior section size is wrong, not the contents. (The interpreter decodes each section until it is syntactically complete and then checks the current position against the declared size.)

sbc100:
> I didn't pick data section to follow the code for any other reason other than its the only one that is valid after the code section (aside from custom sections). Odd coincidence the END and is the same byte (0xb) as the data sections code.
> 
> I will add a comment as to why error message from the interpreter its somewhat misleading.

I dropped the data section and used a second function instead to signal the early end of the first function.

sbc100:
@rossberg would you consider enabling the "Auto-merge" feature on this repo?   I basically means I can have change like this merge if & when the tests all pass without having to remember to come back and check myself.

test/core/binary.wast
```diff
@@ -368,6 +368,25 @@
   "integer too large"
 )
 
+;; Function with missing end marker
+(assert_malformed
+  (module binary
+    "\00asm" "\01\00\00\00"
+    "\03\02\01\00"             ;; Function section
```
rossberg, line 375: You may want to adjust this to have two versions as well.

Perhaps also add a type section, because the function type now actually points to an undefined type index, which may make most decoders bail out early, before even reaching the intended bug.

test/core/binary.wast
```diff
@@ -368,6 +368,25 @@
...
+    "\0a\06\02"                ;; Code section with 2 functions
```
rossberg, line 376: The section length should be adjusted as well.

rossberg (commented):
He he. It's actually an interesting corner case, so perhaps it's worth having both versions of the test? You could also add a third version where there is nothing past the missing end.

rossberg:
As for auto-merge, I'm not sure it's a good idea for a repo like this. It's frequent practice that reviewers approve a PR despite small comments for change, in order to avoid another day-long turn-around across time zones. I have seen it happen too often that the auto-merge label is overlooked in such cases, causing a PR to be merged unintended.

sbc100:
> As for auto-merge, I'm not sure it's a good idea for a repo like this. It's frequent practice that reviewers approve a PR despite small comments for change, in order to avoid another day-long turn-around across time zones. I have seen it happen too often that the auto-merge label is overlooked in such cases, causing a PR to be merged unintended.

True, I have noticed that too.  

It would be nice if there was way to for the auto-merge to only ever be triggered by the author and never by the approver.  I think that would avoid all the accidental merging (it would be the same as the power the author already has to land things with the current merge button but avoid them having to wait for tests).

i.e. "Auto-merge on test completion" as distinct from "Auto-merge on review approval" (that later being the dangerous on I think).

sbc100:
Expanded to 3 separate tests:

- Missing end at EOF
- Missing end at function boundary
- Missing end at section boundary

rossberg (approved):
Excellent, thanks!

PR #1412, opened 2022-01-27
[spec] Add note regarding parameter names

### LINKED ISSUE #1411: Text Format: Clarify symbolic identifiers on function type parameters aren't to be used
Currently, the text format defines not only a symbolic identifier for a type, but also, via functype, symbolic identifiers for params. There is ambiguity in the Type Uses documentation caused by this. For example, when one references a type by index, shouldn't its param IDs be inherited if not explicitly overridden?

Ex. one could reasonably think they should.
```
(type $filter (func (param $offset i32) (param $count i32)))

(func $nop_filter (type $filter) ...)
```

However, via a [discussion](https://github.com/bytecodealliance/wasm-tools/issues/456#issuecomment-1022410776) with @alexcrichton and @rossberg, it seems the parameter IDs on a type are not intended to be used anywhere. I think there should be a spec or doc change as a result and don't mind either.

a. Remove symbolic param IDs on a function type definition. That removes the ambiguity and eliminates currently useless IDs. If someone wants to document, they can use comments
b. Clarify in the "Type Uses" note that any param IDs on a type referenced by index are to be ignored when binding the symbolic index of locals (and the name section for that matter).

rossberg:
Fix #1411.

### DIFF
--- a/document/core/text/types.rst
+++ b/document/core/text/types.rst
@@ -93,6 +93,10 @@ Function Types
        &\Rightarrow& t \\
    \end{array}
 
+.. note::
+   The optional identifier names for parameters in a function type only have documentation purpose.
+   They cannot be referenced from anywhere.
+
 
 Abbreviations
 .............


codefromthecrypt (approved):
Works for me, thanks!

PR #1420, opened 2022-02-10
Clarifies the note on text format export abbreviation expansion to reflect impact

### LINKED ISSUE #1418: Reword 6.6.5.1. Function Abbreviations to be more specific about impact
The WebAssembly 1.0 (MVP) and 1.1 draft both allow repetition and co-mingling of inlined/abbreviated function import and exports.

Screen Shot 2022-02-07 at 8 42 33 AM

The wording "The latter abbreviation can be applied repeatedly, with “…” containing another import or export." is more wide than necessary. The only possible grammar seems `(export)* (import)?` because the "import", if present, must be present at the end.

For example, the valid combinations besides no inlined import/export

* only import -> this means the func cannot have a body or locals, it is effectively a different way of writing import func
* N exports -> this is a module-defined func where N exports point to the current position in the index namespace
* N exports then import -> this is an imported function re-exported N times

To figure these points out from the spec requires too much thinking, so I think the section can be worded, ya?

codefromthecrypt:
While implementing this specification, I became confused as note clarifying the expansion grammar was confusing me.

"The latter abbreviation can be applied repeatedly, with “…” containing another import or export."

^^ seems to hint that you can do (export) (import) (export) (import) resulting in multiple imports that won't work as there can be at most one and really only exports can repeat.

This changes the text to hopefully prevent future folks from reaching the same misconclusion.

Fixes #1418

### DIFF
--- a/document/core/text/modules.rst
+++ b/document/core/text/modules.rst
@@ -250,7 +250,9 @@ Functions can be defined as :ref:`imports <text-import>` or :ref:`exports <text-
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export.
+.. note::
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a function declaration can contain any number of exports, possibly followed by an import.
 
 
 .. index:: table, table type, identifier
@@ -319,7 +321,9 @@ Tables can be defined as :ref:`imports <text-import>` or :ref:`exports <text-exp
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export or an inline elements segment.
+.. note::
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a table declaration can contain any number of exports, possibly followed by an import.
 
 
 .. index:: memory, memory type, identifier
@@ -379,7 +383,9 @@ Memories can be defined as :ref:`imports <text-import>` or :ref:`exports <text-e
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export or an inline data segment.
+.. note::
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a memory declaration can contain any number of exports, possibly followed by an import.
 
 
 .. index:: global, global type, identifier, expression
@@ -423,7 +429,9 @@ Globals can be defined as :ref:`imports <text-import>` or :ref:`exports <text-ex
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export.
+.. note::
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a global declaration can contain any number of exports, possibly followed by an import.
 
 
 .. index:: export, name, index, function index, table index, memory index, global index


codefromthecrypt:
one last note is that while some may assume the spec is only for implementors of runtimes and compilers, end users of the text format will hit similar snags. Personally, I will write in an error on `(memory (data "x")(export "memory"))` something like "export abbreviations must be first" because bombing with "unexpected token" or similar will be even harder to understand by end users vs folks like me who are literally writing the text compiler. In general, the point if I'm able to make it is let's try to help people understand quickly the intent as a larger audience than we might know may end up looking at this spec to figure out for example why their code isn't compiling.

document/core/text/modules.rst
```diff
@@ -250,7 +250,7 @@ Functions can be defined as :ref:`imports <text-import>` or :ref:`exports <text-
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export.
+Only the export abbreviation can be repeated, with ":math:`\dots`". This means expansion can result in any number of exports followed by up to one import.
```
rossberg, line 253: Hm, not sure this is much clearer than the original wording. How about:
```suggestion
.. note::
   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
   Consequently, a function definition can contain any number of exports, followed by either an import or a function body.
```

rossberg (commented):
Thanks for looking into this, see my suggestion below.

codefromthecrypt:
works well. I will apply your rewording to the other points. thanks for the look!

document/core/text/modules.rst
```diff
@@ -425,7 +427,8 @@ Globals can be defined as :ref:`imports <text-import>` or :ref:`exports <text-ex
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-Only the export abbreviation can be repeated, with ":math:`\dots`". This means expansion can result in any number of exports followed by up to one import.
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a function definition can contain any number of exports, followed by either an import or a global type.
```
codefromthecrypt, line 431: not sure "global type" is the best wording here, so if you have a better one let me know!

document/core/text/modules.rst
```diff
@@ -250,7 +250,9 @@ Functions can be defined as :ref:`imports <text-import>` or :ref:`exports <text-
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export.
+.. note::
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a function definition can contain any number of exports, followed by either an import or a function body.
```
rossberg, line 255:
```suggestion
   Consequently, a function declaration can contain any number of exports, possibly followed by an import.
```

document/core/text/modules.rst
```diff
@@ -319,7 +321,8 @@ Tables can be defined as :ref:`imports <text-import>` or :ref:`exports <text-exp
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export or an inline elements segment.
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a function definition can contain any number of exports, followed by either an import or an inline element segment.
```
rossberg, line 325:
```suggestion
   Consequently, a table declaration can contain any number of exports, possibly followed by an import.
```

document/core/text/modules.rst
```diff
@@ -319,7 +321,8 @@ Tables can be defined as :ref:`imports <text-import>` or :ref:`exports <text-exp
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
```
rossberg, line 323:
```suggestion

.. note::
```

document/core/text/modules.rst
```diff
@@ -423,7 +427,8 @@ Globals can be defined as :ref:`imports <text-import>` or :ref:`exports <text-ex
...
```
rossberg, line 429:
```suggestion

.. note::
```

document/core/text/modules.rst
```diff
@@ -379,7 +382,8 @@ Memories can be defined as :ref:`imports <text-import>` or :ref:`exports <text-e
...
```
rossberg, line 384:
```suggestion

.. note::
```

document/core/text/modules.rst
```diff
@@ -379,7 +382,8 @@ Memories can be defined as :ref:`imports <text-import>` or :ref:`exports <text-e
...
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export or an inline data segment.
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a function definition can contain any number of exports, followed by either an import or an inline data segment.
```
rossberg, line 386:
```suggestion
   Consequently, a memory declaration can contain any number of exports, possibly followed by an import.
```

document/core/text/modules.rst
```diff
@@ -423,7 +427,8 @@ Globals can be defined as :ref:`imports <text-import>` or :ref:`exports <text-ex
        (\iff \Tid^? \neq \epsilon \wedge \Tid' = \Tid^? \vee \Tid^? = \epsilon \wedge \Tid' \idfresh) \\
    \end{array}
 
-The latter abbreviation can be applied repeatedly, with ":math:`\dots`" containing another import or export.
+   The latter abbreviation can be applied repeatedly, if ":math:`\dots`" contains additional export clauses.
+   Consequently, a function definition can contain any number of exports, followed by either an import or a global type.
```
rossberg, line 431:
```suggestion
   Consequently, a global's declaration can contain any number of exports, possibly followed by an import.
```

document/core/text/modules.rst
rossberg: The type is present in either case, the difference would be the initialiser after the type. But see my suggestion to drop that part.

rossberg (commented):
Thanks! After seeing how it transfers to the other forms, I realised my suggestion gets complicated. For example, to be accurate, for tables we'd need to say:
```
Consequently, a table declaration can contain
any number of exports, possibly followed by
either an import and a table type or a table
type and an optional element segment. If the 
element segment is present, the table type's
limits are omitted. :)
```
(since the segment comes after the type, and is optional, and and...) This is tedious explaining of the whole grammar in prose, which would bury the main point of the note. So I made suggestions below to simplify my previous wording and focus on the im/exports.

codefromthecrypt:
thanks and I agree that focus > expanding all possible outcomes!

codefromthecrypt:
squashed

rossberg (approved):
Thanks @codefromthecrypt! But I'm afraid you'll still need to sign up with the W3C CG, using the same email as on Github. Otherwise the IPR check will keep blocking this PR.

codefromthecrypt:
ok I signed up https://www.w3.org/accounts/request (confirmed email) then https://www.w3.org/community/webassembly/join (as company). My w3c profile shows I'm in the CG now and I also linked github and clicked refresh status. IPR still doesn't like me, maybe I have to wait a bit.

codefromthecrypt:
maybe it is ok now! with the paperwork done I'm happy to help in something else that comes up to (where help is have Andreas chisel something good out of the rough)

rossberg:
Thanks!

Issue #1386, opened 2021-10-27
[spec] Potential accessibility issues with singlehtml due to missing MathML


ngzhian:
I was looking at https://github.com/WebAssembly/spec/blob/941c6f37cb13de20f54b89b73145f802b87b8155/document/core/util/mathjax2katex.py#L117

By default, katex will generate something like:
```
<span class="katex-display"><span class="katex"><span class="katex-mathml"><math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><semantics><mrow><mrow><mn mathvariant="monospace">0</mn><mi mathvariant="monospace">x</mi><mrow><mi mathvariant="monospace">F</mi><mi mathvariant="monospace">C</mi></mrow></mrow><mtext> </mtext><mrow><mn mathvariant="monospace">0</mn><mi mathvariant="monospace">x</mi><mn mathvariant="monospace">11</mn></mrow></mrow><annotation encoding="application/x-tex">\mathrm{{\mathtt{0x{FC}}}~{\mathtt{0x{11}}}}</annotation></semantics></math></span><span class="katex-html" aria-hidden="true"><span class="base"><span class="strut" style="height:0.61111em;vertical-align:0em;"></span><span class="mord"><span class="mord"><span class="mord"><span class="mord mathtt">0x</span><span class="mord"><span class="mord mathtt">FC</span></span></span></span><span class="mspace nobreak"> </span><span class="mord"><span class="mord"><span class="mord mathtt">0x</span><span class="mord"><span class="mord mathtt">11</span></span></span></span></span></span></span></span></span>
```

Notice the `katex-mathml` and `<math>` DOM nodes. We strip those away by skipping directly to `<span class="katex-html">`.

We can can actually skip this post-processing with a [katex option](https://katex.org/docs/options.html#:~:text=output%3A%20string.%20Determines,is%20the%20default.) but it seems like MathML is for accessibility.

I don't now how the singlehtml doc currently is like under, say, a screen reader, and whether the missing MathML is acceptable, but I will go read a bit more about it.

ericprud:
On Monday I'll post a pointer to the output of the sphinxToTr, which I hope to replace the one-pager. It basically packages the sphinx output so if it's in sphinx, it's in this (these) doc

ericprud:
It's one day later than Monday, but here's a snapshot of my plan A for publishing the core spec as a multi-page doc:
  https://www.w3.org/2021/11/wasm-stage/
[...]

PR #1429, opened 2022-03-14
[bikeshed] WIP new w3c html

ngzhian:
This uses a multihtml Sphinx builder, together with custom HTML layout
(layout.html and page.html), to build a document that matches w3c
requirements.

Currently the toctree directive does not work right, so it is duplicated
in the side bar (https://github.com/sphinx-doc/sphinx/pull/9830). The
W3C styles also don't apply fully to the side bar (because the Sphinx
generated TOC doesn't have the right class name, but this might be fixed
by some JS).

layout.html is used for all pages, index + individual pages, so the
front matter is guarded by a check that the current page is index.
layout.html is rather large, most of it is the front matter copied from
https://github.com/WebAssembly/spec/issues/1386#issuecomment-957347678.
page.html is used by all the individual rst pages.

Right now all the configuration lives in a new folder, w3chtml, but it
can presumably replace the w3chtml folder.


ericprud:
Many thanks for the W3C doc work.

Just as an FYI, W3C specs don't have to be monolithic; it's up to us to publish as a single doc (maybe nice for folks who want to print?) or a multiple pages (more responsive). It might be possible to publish both, but I imagine that simply publishing a multi-page doc with a link to a printable version would be simpler.

Also, we don't have to work all this out before publishing.

document/core/index.rst
```diff
@@ -10,16 +10,19 @@ WebAssembly Specification
...
-   intro/index
-   syntax/index
-   valid/index
-   exec/index
-   binary/index
-   text/index
-   appendix/index
+.. only:: builder_html
+
+  .. toctree::
```
rossberg, line 15: Won't this remove the toctree from the PDF?

document/core/Makefile
```diff
@@ -120,6 +120,22 @@ html: index
 	@echo
 	@echo "Build finished. The HTML pages are in `pwd`/$(BUILDDIR)/html/."
 
+.PHONY: newhtml
```
rossberg, line 123: Should we call this `w3chtml` perhaps?

document/core/conf.py
```diff
@@ -62,8 +62,12 @@
 title = u'WebAssembly Specification'
 copyright = u'2017, WebAssembly Community Group'
 author = u'WebAssembly Community Group'
-editor = u'Andreas Rossberg (editor)'
+editor_name = u'Andreas Rossberg'
+editor = editor_name + u' (editor)'
```
rossberg, line 66: Nit: rename editor_name to editor and inline the "(editor)" where needed? (I see only one occurrence anyway)

ngzhian:
> Many thanks for the W3C doc work.
> 
> Just as an FYI, W3C specs don't have to be monolithic; it's up to us to publish as a single doc (maybe nice for folks who want to print?) or a multiple pages (more responsive). It might be possible to publish both, but I imagine that simply publishing a multi-page doc with a link to a printable version would be simpler.
> 
> Also, we don't have to work all this out before publishing now; we can add features later.

the PR title is a misnomer, it should just be "new html". I think I called it new single html because the build step followed the singlehtml. Let me fix that.

rossberg:
@ngzhian, what's the status of this PR, is it still needed?

ngzhian:
Not actively working on this, you can close it if you'ld like.

PR #1431, opened 2022-03-18
Add --generate-js-only flag to test runner

### LINKED ISSUE #1430: Proposal phase process requirements for test suite
In phase 2, a test suite is added.
The entry requirement for phase 3 is:
> Test suite has been updated to cover the feature in its forked repo.

But it doesn't need to run against the reference interpreter.
> The test suite should run against some implementation, though it need not be the reference interpreter.

It is actually quite difficult to have tests but not have reference interpreter implementation because:
- most web engines run JS tests
- JS tests are generated from .wast files by the interpreter
- in order to get to generating JS, the test runner [first runs the wast file to make sure that it passes](https://github.com/WebAssembly/spec/blob/main/test/core/run.py#L68)

So in the current state, to be able to generate JS test files, you will need to implement it in the reference interpreter first.

I suggest that we make a tiny tweak to https://github.com/WebAssembly/spec/blob/main/test/core/run.py#L68 to take a flag to just generate js, and skip running the tests (and round tripping).

ngzhian:
This will return early right after generating JS from the wast test
files. It will not attempt to run the tests, or do the round trip
conversion from wasm <-> wast.

This is convenient for proposals to add tests without having to
update the reference interpreter with implementation, and generate those
tests to JS to run in other Wasm engines.

Fixes #1430.

### DIFF
--- a/test/core/run.py
+++ b/test/core/run.py
@@ -17,6 +17,7 @@ outputDir = os.path.join(inputDir, "_output")
 parser = argparse.ArgumentParser()
 parser.add_argument("--wasm", metavar="<wasm-command>", default=os.path.join(os.getcwd(), "wasm"))
 parser.add_argument("--js", metavar="<js-command>")
+parser.add_argument("--generate-js-only", action='store_true')
 parser.add_argument("--out", metavar="<out-dir>", default=outputDir)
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):
lgtm

PR #1434, opened 2022-03-26
[spec] Fix single-table limitation in module instantiation

keithw:
The current spec (in 4.5.4 Instantiation) asserts that active elem segments may only reference table #0. (The reference-types proposal removed this limitation from 2.5.7 Element Segments, but left the corresponding assertion in 4.5.4 Instantiation. I think this was probably unintended and the assertion is vestigial?)

This PR removes the assertion and allows active elem segments to reference any tableidx.


rossberg (approved):
Thanks!

Issue #1439, opened 2022-03-31
Encoding of prefix in data segment entries


nagisa:
Hi,

In WebAssembly 1.0 the data segment encoding was specified as

```
datasec ::= vec(data)
data ::= memidx expr vec(byte)
```

and with the merge of `bulk-memory` proposal (AFAIU) it has become

```
datasec ::= vec(data)
data ::= 0x00 expr vec(byte)
       | 0x01 vec(byte)
       | 0x02 memidx expr vec(byte)
```

This sort of seems backwards-compatible at a first glance, but it isn't actually due to `memidx` using the LEB128 encoding. This means that in the original specification `00`, `80 00`, `80 80 00`, `80 80 80 00`, `80 80 80 80 00` were all valid prefix bitstream patterns for the `data`. As a result binary encoded modules that were valid as per the currently published WebAssembly specification are no longer valid in the draft.

I wonder if it would make sense to adjust the description of the `data` as such:

```
datasec ::= vec(data)
data ::= 0x00 expr vec(byte)
       | 0x01 vec(byte)
       | 0x02 memidx expr vec(byte)
       | 0x80 0x00 expr vec(byte)
       | 0x80 0x80 0x00 expr vec(byte)
       | 0x80 0x80 0x80 0x00 expr vec(byte)
       | 0x80 0x80 0x80 0x80 0x00 expr vec(byte)
```

or perhaps even

```
datasec ::= vec(data)
data ::= 0x00 expr vec(byte)
       | 0x01 vec(byte)
       | 0x02 memidx expr vec(byte)
       | 0x80 memidx expr vec(byte) // note, though that it ends up accepting `80 80 80 80 80 00` as an additional valid encoding.
```

in order to maintain the backwards compatibility? Seems like it'd be a pretty painless change to make.

nagisa:
I haven't checked too closely if there are other occurrences of this, but at least the element section entries are also effected in much the same way.

nagisa:
All implementations of what, specifically? Producers? Consumers? On the consumer side of things the wasm-tools' wasmparser will happily validate any LEB encoding of `0` for `memidx` since the very first iterations of the project, as far as I can tell. On the producer side… a fuzzer comes to mind as something that could produce arbitrary LEB encoding of `0`.

rossberg:
The interpreter actually maintains backwards compatibility by deliberately using the equivalent of
```
data ::= i:u32 expr vec(byte)          (if i = 0)
[...]

PR #1440, opened 2022-04-01
Update decoding of element/data segments in spec interpreter

alexcrichton:
This commit updates the spec interpreter after the merge of the bulk
memory proposal to align with the textual specification for the encoding
of data and element segments. In original MVP wasm a data/element
segmented started with a leb128 for the memory/table index, but this
leb128 was repurposed as a flags byte in the bulk-memory/reference-types
proposals since it was always zero in practice (and never an over-long
zero such as `"\80\00"`). The spec interpreter, however, hasn't been
updated and was still reading a u32 for the flags byte, so this commit
updates it to instead read a single byte.

The tests have been updated accordinatly. Tests for overlong or invalid
index encodings were updated to use an encoding that explicitly
specifies the index (the prefix `"\02"` byte for both data and element
segments). New tests were added to ensure that an overlong encoding of
0, which was previously valid, is no longer valid.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -951,41 +951,41 @@ let elem_kind s =
   | _ -> error s (pos s - 1) "malformed element kind"
 
 let elem s =
-  match vu32 s with
-  | 0x00l ->
+  match u8 s with
+  | 0x00 ->
     let emode = at active_zero s in
     let einit = vec (at elem_index) s in
     {etype = FuncRefType; einit; emode}
-  | 0x01l ->
+  | 0x01 ->
     let emode = at passive s in
     let etype = elem_kind s in
     let einit = vec (at elem_index) s in
     {etype; einit; emode}
-  | 0x02l ->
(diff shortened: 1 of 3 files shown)


alexcrichton:
While #1439 is somewhat related this is intended to fix the interpreter to match the current text spec. The text spec does not mention a leb for the leading byte discriminant for element/data segments but the spec interpreter uses a leb. This means that parsers which strictly implement the textual spec cannot pass the current spec tests and this PR is trying to fix that.

I recall that the decision at the time when this came up was to intentionally break any modules using overlong encodings for index 0. I found https://github.com/WebAssembly/bulk-memory-operations/issues/153 as some discussion of this but I don't recall where other discussion happened. I thought, though, that #1439 was "intended breakage" and action wasn't necessary.

rossberg:
Well, the question raised on #1439 is whether this is a bug in the spec. But your recollection makes sense to me. Perhaps point this out on #1439, so it can be resolved?

PR #1441, opened 2022-04-04
[spec] exec/instructions.rst: fix missing immediate on table.set

keithw:
The table.set instruction was missing its immediate `x` in the execution spec heading.


rossberg (approved):
Thanks!

PR #1442, opened 2022-04-07
multi-value: fixes compilation errors in examples

codefromthecrypt:
This fixes compilation errors in examples. It also changes case format of phis to PHIs so that people don't first wonder if they are reading a typo.

### DIFF
--- a/proposals/multi-value/Overview.md
+++ b/proposals/multi-value/Overview.md
@@ -43,7 +43,7 @@
 A simple swap function.
 ```wasm
 (func $swap (param i32 i32) (result i32 i32)
-	(get_local 1) (get_local 0)
+	(local.get 1) (local.get 0)
 )
 ```
 
@@ -51,10 +51,10 @@ An addition function returning an additional carry bit.
 ```wasm
 (func $add64_u_with_carry (param $i i64) (param $j i64) (param $c i32) (result i64 i32)
 	(local $k i64)
-	(set_local $k
-		(i64.add (i64.add (get_local $i) (get_local $j)) (i64.extend_u/i32 (get_local $c)))
+	(local.set $k
+		(i64.add (i64.add (local.get $i) (local.get $j)) (i64.extend_i32_u (local.get $c)))
 	)
-	(return (get_local $k) (i64.lt_u (get_local $k) (get_local $i)))
+	(return (local.get $k) (i64.lt_u (local.get $k) (local.get $i)))
 )
 ```
 
@@ -71,7 +71,7 @@ An addition function returning an additional carry bit.
 Conditionally manipulating a stack operand without using a local.
 ```wasm
 (func $add64_u_saturated (param i64 i64) (result i64)
-	($i64.add_u_carry (get_local 0) (get_local 1) (i32.const 0))
+	(call $add64_u_with_carry (local.get 0) (local.get 1) (i32.const 0))
 	(if (param i64) (result i64)
 		(then (drop) (i64.const 0xffff_ffff_ffff_ffff))
 	)
@@ -80,14 +80,22 @@ Conditionally manipulating a stack operand without using a local.
 
 An iterative factorial function whose loop doesn't use locals, but uses arguments like phis.
 ```wasm
+(func $pick0 (param i64) (result i64 i64)
+	(local.get 0) (local.get 0)
+)
+
+(func $pick1 (param i64 i64) (result i64 i64 i64)
+	(local.get 0) (local.get 1) (local.get 0)
+)
+
 (func $fac (param i64) (result i64)
-	(i64.const 1) (get_local 0)
+	(i64.const 1) (local.get 0)
 	(loop $l (param i64 i64) (result i64)
-		(pick 1) (pick 1) (i64.mul)
-		(pick 1) (i64.const 1) (i64.sub)
-		(pick 0) (i64.const 0) (i64.gt_u)
+		(call $pick1) (call $pick1) (i64.mul)
+		(call $pick1) (i64.const 1) (i64.sub)
+		(call $pick0) (i64.const 0) (i64.gt_u)
 		(br_if $l)
-		(pick 1) (return)
+		(call $pick1) (return)
 	)
 )
 ```


codefromthecrypt:
I took this code from some existing examples in this org, and it compiles with wabt, but I want to verify it actually works

proposals/multi-value/Overview.md
```diff
@@ -71,23 +71,31 @@ An addition function returning an additional carry bit.
...
+	(local.get 0) (local.get 0)
+)
+
+(func $pick1 (param i64 i64) (result i64 i64 i64)
+	(local.get 0) (local.get 1) (local.get 0)
+)
+
 (func $fac (param i64) (result i64)
-	(i64.const 1) (get_local 0)
+	(i64.const 1) (local.get 0)
```
codefromthecrypt, line 92: this doesn't handle zero, but maybe it is too complicated to explain code to fix that
```
    ;; ensure the loop completes on input zero
     i64.eqz if (result i64) i64.const 1 else local.get 0 end
```

proposals/multi-value/Overview.md
```diff
@@ -30,7 +30,7 @@
 
 * Inputs to blocks:
   - loop labels can have arguments
-  - can represent phis on backward edges
+  - can represent PHIs on backward edges
```
rossberg, line 33: "phi" is a greek letter, not an acronym, so this should remain lowercase.

proposals/multi-value/Overview.md
```diff
@@ -71,23 +71,31 @@ An addition function returning an additional carry bit.
...
-	($i64.add_u_carry (get_local 0) (get_local 1) (i32.const 0))
+	(call $add64_u_with_carry (local.get 0) (local.get 1) (i32.const 0))
 	(if (param i64) (result i64)
 		(then (drop) (i64.const 0xffff_ffff_ffff_ffff))
 	)
 )
 ```
 
-An iterative factorial function whose loop doesn't use locals, but uses arguments like phis.
+An iterative factorial function whose loop doesn't use locals, but uses arguments like PHIs.
```
rossberg, line 81: Same here.

proposals/multi-value/Overview.md
```diff
@@ -71,23 +71,31 @@ An addition function returning an additional carry bit.
...
 	(loop $l (param i64 i64) (result i64)
-		(pick 1) (pick 1) (i64.mul)
-		(pick 1) (i64.const 1) (i64.sub)
-		(pick 0) (i64.const 0) (i64.gt_u)
+		(call $pick1) (call $pick1) (i64.mul)
+		(call $pick1) (i64.const 1) (i64.sub)
+		(call $pick0) (i64.const 0) (i64.gt_u)
 		(br_if $l)
-		(pick 1) (return)
+		(drop) (return)
```
rossberg, line 98:
```suggestion
		(call $pick1) (return)
```

rossberg (commented):
Thanks. I'm fine with this, though in general, the design docs under proposals/ are historic, so they naturally might be out of date. I think there is no expectation that we keep them updated.

proposals/multi-value/Overview.md
codefromthecrypt: okie

proposals/multi-value/Overview.md
codefromthecrypt: yeah main thing I was hoping was to avoid someone seeing phis and thinking did that mean this? as you'll notice the example follows. Anyway hopefully new specs can try to avoid unnecessary jargon in examples.

codefromthecrypt:
>  the design docs under proposals/ are historic, so they naturally might be out of date. I think there is no expectation that we keep them updated.

Not trying to be fussy, but I did need to look at this to figure out what this feature needs. Is there an authoritative way to get the same content? https://webassembly.github.io/spec/core/bikeshed/#multiple-values%E2%91%A0 has very little content and also links here..

proposals/multi-value/Overview.md
codefromthecrypt: done

codefromthecrypt:
FWIW to implement this feature, the essential things I needed were to read this proposal which gave more detail about the general expectations. To actually do the work was reading the commit diff.

The summary at the bottom of the spec was helpful but not really for implementation. For implementation, at least for me, it is this doc and the commit that changed the spec tests because spec tests are not grouped by feature otherwise.

codefromthecrypt:
also note to any future reader that there's typically an assumption of LLVM experience when looking at specs here. one shortcut is when in doubt about something search LLVM (that something), ex to get to the phi instruction which the "phi" here is about https://llvm.org/docs/LangRef.html#phi-instruction

rossberg (approved):
Thanks!

To clarify, phi nodes are part of SSA representation, so relatively generic compiler tech, nothing specific to LLVM (I have very little knowledge of LLVM myself :) ).

codefromthecrypt:
The main point is that this spec shouldn't have prerequisites on both knowing jargon and also knowing what scope the jargon is in. 

It would kill no one to say "phi nodes (in SSA representation)"

saying "phis" is basically suggesting everyone whose implementing this can jump to that context by looking at factorial code. That's an unnecessary narrowing.

We can't change the past, but I really hope we can at least recognize that having a review by someone with less experience can help with accessibility of the spec. I don't think the goal of this spec is everyone basically depending on one impl, and to change that, where the spec is highly reused, things like this help.

I'm an end user and it is places like this, where I feel there's basically some secret club who know what all the background needed is, as things are almost intentionally terse. Great example is using word like phis as if it were a normal programming term, to describe factorial!

codefromthecrypt:
added a commit you can feel free to have me revert it, but would have saved what I feel was unnecessary jump to wondering if a sentence in front of code that says "phis" might have meant "this"

---

This also adds the context necessary (SSA) that makes learning why this is important possible for someone who doesn't already know why. Remember there are a lot of different types of engineers who can help on code around WebAssembly.

Also, the text format is used for things besides implementing compilers. People are using it for "user defined functions", general purpose code written in the text format that's intent is higher level. This doesn't negate the value in implementing things with a lot of context, but it does say that there is value being decoupled also. I've had to personally reverse document a lot of things because there are norms like no comments in spec tests, which assume people reading know tons. Anyway I feel the last commit is more empathetic even if it has some grammatical glitches as it moves from someone thinking why am I too dumb to know what "phi" is to, oh. that's what phi is! because if you try to search "phi" on its own good luck, "phis" is worse.

PS I hesitated from making a link on SSA, as I noticed web links aren't really used except to refer to GitHub. However, if this were my codebase I would (and do) when using jargon as that saves the search engine step.

rossberg:
Well, this isn't the spec document, but a historic design doc. It is preserved in the repo to document and reflect the original discussion. Perhaps the README should have a disclaimer clarifying that.

I understand your intention, but I'm not sure I'm comfortable changing these docs beyond fixing simple typos/syntax, good intentions notwithstanding. That would have a touch of "rewriting history".

codefromthecrypt:
ok I'll kill the last commit, thanks for listening

rossberg:
Thanks!

PR #1454, opened 2022-04-27
Adds missing start from overview example

codefromthecrypt:
Lost some time not noticing this. While it is possible others will guess sooner than I did
by function name and it isn't exported that they should add it as a start function, better
to be explicit I think.


rossberg (approved):
Fine by me, though this is probably not essential to the example, since there may be an application-specific policy to invoke the $start function.

PR #1458, opened 2022-05-02
Replace of with or

tpmccallum:
I think this paragraph was meaning to say **or** i.e.
`number types | vectors | reference types`


rossberg (approved):
Thanks!

PR #1459, opened 2022-05-03
Turn interpreter into a Dune package

vapourismo:
Add a files necessary to build the package with Dune.

I had to delete `.merlin` file to get the LSP working. Merlin and Dune communicate directly.


interpreter/dune
```diff
@@ -0,0 +1,29 @@
...
+   (chdir
+    %{workspace_root}
+    (run %{bin:ocamllex} -ml -q -o %{target} %{deps}))))
+ (ocamlyacc
+  (modules parser)))
+
+(env
+ (dev
+  (flags
+   (-w -27-8))))
```
rossberg, line 29: Can we make this default to the same flags used in the Makefile?

rossberg (commented):
Thanks, this looks good to me. My only question is: isn't the module `Interpreter_lib` that gets produced structurally identical to the wrapper module `Wasm` produced by the opam target in the Makefile? If so, maybe call it the same?

(Also, would this allow to build that opam package with your dune file?)

interpreter/dune
vapourismo: Good idea. Done.

vapourismo:
> isn't the module `Interpreter_lib` that gets produced structurally identical to the wrapper module `Wasm` produced by the opam target in the Makefile? If so, maybe call it the same?

I believe they are structurally equal, but naming it `Wasm` has the unfortunate side-effect that modules with the same name as the library are considered the library interface. I did rename it, however, I had to work around it with a module exclusion. Please have a look if that is ok.

> (Also, would this allow to build that opam package with your dune file?)

It is the basis for it. To do so, you would have to give the [library entry](https://dune.readthedocs.io/en/stable/dune-files.html#library-1) a `public_name` and extend the `dune-project` to enable [generate_opam_files](https://dune.readthedocs.io/en/stable/dune-files.html#generate-opam-files-1) for a [package description](https://dune.readthedocs.io/en/stable/dune-files.html#package-1).

I'd happily make the change, if you need it.

rossberg (approved):
Thanks!

Isn't the existing Wasm module exactly that module for the Opam package, and not used otherwise? That is, if we were to build the package with Dune, there would be no conflict?

I'd be happy to have that build with Dune, if you are up for changing that.

vapourismo:
As far as I can tell they expose the same interface so there would be no conflict.

PR #1461, opened 2022-05-04
[spec] Fix binary format of data/elem tags to allow LEB

### LINKED ISSUE #1439: Encoding of prefix in data segment entries

rossberg:
Change the encoding of data/element segment tags to be a full LEB u32 instead of a single byte. This matches test suite, reference interpreter, and major engines like V8, JSC(*), Spidermonkey, and Wasmtime. It also reinstantiates backwards compatibility with the Wasm 1.0 format.

Fixes #1439.

(*) With the one exception that JSC [apparently](https://github.com/WebAssembly/spec/issues/1439#issuecomment-1088922572) requires a single byte for element segments but not data segments, an inconsistency that seems more like accident than intent.

### DIFF
--- a/document/core/binary/modules.rst
+++ b/document/core/binary/modules.rst
@@ -327,28 +327,28 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
    \production{element section} & \Belemsec &::=&
      \X{seg}^\ast{:}\Bsection_9(\Bvec(\Belem)) &\Rightarrow& \X{seg} \\
    \production{element segment} & \Belem &::=&
-     \hex{00}~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
+     0{:}\Bu32~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
        &\Rightarrow& \\&&&\quad
        \{ \ETYPE~\FUNCREF, \EINIT~((\REFFUNC~y)~\END)^\ast, \EMODE~\EACTIVE~\{ \ETABLE~0, \EOFFSET~e \} \} \\ &&|&
-     \hex{01}~~\X{et}:\Belemkind~~y^\ast{:}\Bvec(\Bfuncidx)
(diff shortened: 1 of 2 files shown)


gahaas (approved):

rossberg:
Thanks @gahaas. I added a couple more tests to cover more 2.0 cases, PTAL.

I will wait with landing this until discussion on #1439 has converged.

gahaas:
Still LGTM

PR #1464, opened 2022-05-05
Add missing default for vector types

ia0:
Each value type has an associated default value, however only numeric types and reference types are listed, while there are also values of vector types.


document/core/exec/runtime.rst
```diff
@@ -51,11 +51,11 @@ or *external references* pointing to an uninterpreted form of :ref:`extern addre
 .. _default-val:
 
 Each :ref:`value type <syntax-valtype>` has an associated *default value*;
-it is the respective value :math:`0` for :ref:`number types <syntax-numtype>` and null for :ref:`reference types <syntax-reftype>`.
+it is the respective value :math:`0` for :ref:`number types <syntax-numtype>` and :ref:`vector types <syntax-vectype>`, and null for :ref:`reference types <syntax-reftype>`.
 
 .. math::
    \begin{array}{lcl@{\qquad}l}
-   \default_t &=& t{.}\CONST~0 & (\iff t = \numtype) \\
+   \default_t &=& t{.}\CONST~0 & (\iff t = \numtype \vee t = \vectype) \\
```
rossberg, line 58: This doesn't quite work, since the syntax for vector constants is slightly different and includes a shape. You'll need to make it a separate case.

rossberg: Ah, you are right, sorry, I was confusing abstract syntax with the text format.

Just for stylistic reasons, I'd still prefer a separate line for the vectype case.

ia0: I see, I didn't read binary and text format yet.

I used a separate line. There was 2 options:
- Inline `\vectype` as `v128`
- Keep generic over `\vectype`

I opted for the second option assuming that if additional vector types are added in the future they will probably only differ in value range and thus have a constant initialization taking a natural number. But I can change to the first option in case we don't foresee additional vector types or they won't be shaped as natural numbers.

rossberg (approved):
Thank you for all the fixes, much appreciated!

One remark: please do not use force-push to a branch under review, since that makes it hard for reviewers to identify relative changes. It also messes up review comments. ;)

PR #1465, opened 2022-05-05
Fix syntax error in element segments validation rule

ia0:
Expressions are valid with a result type and not simply ok. It is also important that elements are well-typed for soundness.

### DIFF
--- a/document/core/valid/modules.rst
+++ b/document/core/valid/modules.rst
@@ -149,7 +149,7 @@ Element segments :math:`\elem` are classified by the :ref:`reference type <synta
 
 * For each :math:`e_i` in :math:`e^\ast`,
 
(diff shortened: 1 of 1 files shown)


document/core/valid/modules.rst
```diff
@@ -149,7 +149,7 @@ Element segments :math:`\elem` are classified by the :ref:`reference type <synta
 
 * For each :math:`e_i` in :math:`e^\ast`,
 
-  * The expression :math:`e_i` must be :ref:`valid <valid-expr>`.
+  * The expression :math:`e_i` must be valid with :ref:`result type <syntax-resulttype>` :math:`[t]`.
```
rossberg, line 152:
```suggestion
  * The expression :math:`e_i` must be :ref:`valid <valid-expr>` with some :ref:`result type <syntax-resulttype>` :math:`[t]`.
```

PR #1472, opened 2022-05-12
TypeError for Global constructor with v128

gahaas:
At the moment the spec requires a `LinkError` to be thrown when the `WebAssembly.Global` constructor is called for type `v128`. This was introduced in https://github.com/WebAssembly/simd/pull/360, but according to the PR description, actually a `TypeError` should be thrown. The PR refers to https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#javascript-api-and-simd-values, and there a `TypeError` is required.


Ms2ger (approved):
Seems to have been done in https://github.com/WebAssembly/simd/pull/360; cc @ngzhian @dtig 

No objection from me, but I'll wait a little while to merge in case anyone disagrees.

ngzhian:
LGTM, thanks!

PR #1478, opened 2022-05-17
[interpreter] Dune test suite

vapourismo:
This PR adds a few tweaks to allow us to run the test suite against the reference interpreter with a simple call to `dune test`.


interpreter/dune
```diff
@@ -6,7 +6,8 @@
  ; WASM REPL every time in all the dependencies.
```
rossberg, line 6:
```suggestion
 ; Wasm REPL every time in all the dependencies.
```

rossberg (approved):
Looks good to me. Would it be possible to have wrapper targets in the Makefile that invoke these?

vapourismo:
> Would it be possible to have wrapper targets in the Makefile that invoke these?

Sure. Like `make dune-test` for instance?

interpreter/Makefile
```diff
@@ -136,7 +136,7 @@ TESTDIR =	../test/core
 TESTFILES =	$(shell cd $(TESTDIR); ls *.wast; ls [a-z]*/*.wast)
 TESTS =		$(TESTFILES:%.wast=%)
 
-.PHONY:		test debugtest partest
+.PHONY:		test debugtest partest dune-test
```
rossberg, line 139: Nit: `dunetest` for consistency.

Can you also add such a target for the dune build itself?

rossberg (approved):

interpreter/Makefile
vapourismo: Fixed and added.

rossberg (approved):
Thanks!

PR #1490, opened 2022-06-07
[spec] Clarify that strings can separate tokens

tlively:
Since `"` is not included in `idchar`, it can mark the end of a token. For
example, `0"foo"$x` would parse as `0`, `"foo"`, `$x`.

### DIFF
--- a/document/core/text/lexical.rst
+++ b/document/core/text/lexical.rst
@@ -63,7 +63,8 @@ The set of *keyword* tokens is defined implicitly, by all occurrences of a :ref:
 Any token that does not fall into any of the other categories is considered *reserved*, and cannot occur in source text.
 
 .. note::
-   The effect of defining the set of reserved tokens is that all tokens must be separated by either parentheses or :ref:`white space <text-space>`.
+   The effect of defining the set of reserved tokens is that all tokens must be separated by either parentheses,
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Oh, interesting. That is kind of unintentional and inconsistent (e.g., I remember that we explicitly decided to make `0$a` a single token and tweaked the lexical grammar for that purpose at some point in the past).

I wonder if we can afford fixing it? That is, add `"` to the characters that can occur in `reserved`? Technically, that would be a breaking change to the text format, but it's probably unlikely to affect real-world code.

tlively:
Yeah, I can see that it might be surprising that `"` starts a new token but `$` does not. I wonder how many parsers in the wild actually implement this lexing correctly anyway. It's probably worth asking more broadly whether anyone cares before making a breaking change, but I also suspect this wouldn't matter in practice.

rossberg:
Yes, absolutely. Shall we put it on the next meeting's agenda?

How did you run into this in the first place?

tlively:
We had unanimous consent in the CG meeting this morning to make a change to the grammar to disallow strings from separating tokens, although we didn't get into the details of how such a change might be accomplished.

rossberg:
Closing this, since it has been superseded by a spec change.

PR #1492, opened 2022-06-08
[interpreter] Factor data and element segments into modules

rossberg:
In order to enable replacing them with a different representation.

@vapourismo, PTAL.


vapourismo:
Looks good. Thanks for the heads up.

PR #1504, opened 2022-07-11
[test] Replace the non-existent "test" module

### LINKED ISSUE #1502: question about the "test" module
Hi,

In some tests (`elem.wasm`, `data.wasm`...) there are some values imported from a `"test"` module. [Example from elem.wast](https://github.com/WebAssembly/spec/blob/main/test/core/elem.wast#L379=):

```wasm
(assert_invalid
  (module
    (global (import "test" "global-i32") i32)
    (table 1 funcref)
    (elem (offset (global.get 0) (global.get 0)))
  )
  "type mismatch"
)
```

I tried to find such a `"test"` module but couldn't find any definition or documentation anywhere. Thus I'm wondering if I really missed it or  if it's indeed not supposed to exist because the various tests referencing it should fail before they actually need it.

If it's the latter, I must admit that I find it quite confusing and I'm wondering if it may not be better to use a real module (I'm thinking of `"spectest"`) ?

zapashcanon:
Fix #1502.

Where possible I used `"spectest"`. In some cases I needed a mutable i32 global which doesn't exists in `"spectest"` so I added a module and registered it.

cc @rossberg

### DIFF
--- a/test/core/data.wast
+++ b/test/core/data.wast
@@ -391,7 +391,7 @@
 )
 
 (assert_invalid
-  (module 
+  (module
     (memory 1)
     (data (offset (;empty instruction sequence;)))
   )
@@ -408,7 +408,7 @@
 
 (assert_invalid
   (module
-    (global (import "test" "global-i32") i32)
+    (global (import "spectest" "global_i32") i32)
     (memory 1)
     (data (offset (global.get 0) (global.get 0)))
   )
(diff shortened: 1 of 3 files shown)


sunfishcode:
An implementation should be able to perform validation without linking, and without any knowledge of the specific things instances will eventually be linked with. That seems valuable to continue to test.

rossberg (approved):
Thanks

rossberg:
@sunfishcode, hah, good point!

zapashcanon:
@sunfishcode yes sure, I can add dedicated tests for that, with a comment explaining their goal ?

sunfishcode:
Comments would be great, however in any case I'd still recommend not switching these tests to "spectest", because implementations shouldn't actually be consulting the actual "spectest" here, and it would seem to risk confusion about whether they should or could.

Issue #1506, opened 2022-07-16
[interpreter] wast.js build appears to be broken


takikawa:
There currently appear to be several issues with the wast.js build mode for the reference interpreter.

One is that Makefile steps for wast.js are a bit broken due to some changes from around Nov 2021. This is relatively easy to fix and the build can proceed to running Bucklescript. (see this commit for a fix: https://github.com/WebAssembly/spec/commit/4148f109a4a210be2ce9e30793f7d67e3186d5a6)

The Bucklescript build, however, doesn't work for me. It looks like, among other things, Bucklescript doesn't support nullary variants introduced in Ocaml 4.07. It also looks like Bucklescript has [rebranded as Rescript](https://rescript-lang.org/blog/bucklescript-is-rebranding) and is de-emphasizing the OCaml support (though they are attempting backwards compat). Recent OCaml release notes have stopped mentioning Bucklescript/Rescript as well: https://v2.ocaml.org/releases/4.14.0.html#Alternative-Compilers

Here's an excerpt of build errors that I get:

```
==== Compiling ====
bsb: [40/226] src/ast.ast
FAILED: src/ast.ast
File "/home/asumu/spec/interpreter/_build/jslib/src/ast.ml", line 44, characters 2-6:
Error: Syntax error
bsb: [41/226] src/lib.ast
FAILED: src/lib.ast
File "/home/asumu/spec/interpreter/_build/jslib/src/lib.ml", line 3, characters 0-6:
Error: Syntax error
bsb: [42/226] src/lib.iast
FAILED: src/lib.iast
```

I'm curious if the bsb build works for anyone else.

Given the above, I am wondering if it could make sense to use Js_of_Ocaml instead to provide the wast.js build mode. I'll upload a draft PR that implements this approach (I haven't extensively tested it, but it works for some simple encode/decode examples) to show what it would look like.

rossberg:
Fixed by #1507.

PR #1507, opened 2022-07-16
[interpreter] Change how wast.js is built by using Js_of_ocaml

takikawa:
See issue #1506 for a discussion on the wast.js build mode state.

This PR replaces the current JS build mode for wast.js by using dune and [Js_of_ocaml](https://ocsigen.org/js_of_ocaml/latest/manual/overview) to compile the reference interpreter to JS.


takikawa:
Looks like this overlaps a bit with #1408 but is more minimally targeted at just the JS support. This PR also does the ArrayBuffer conversion for the JS interface, which wasn't in the other PR AFAIK.

interpreter/Makefile
```diff
@@ -109,12 +109,9 @@ _build/$(LIB).cmxa: $(FILES) $(LIB).mllib _tags Makefile
...
 .PHONY:		$(JSLIB)
-$(JSLIB):	$(UNOPT)
-		mkdir -p _build/jslib/src
-		cp meta/jslib/* _build/jslib
-		cp $(DIRS:%=_build/%/*.ml*) meta/jslib/*.ml _build/jslib/src
-		rm _build/jslib/src/*.ml[^i]
-		(cd _build/jslib; ./build.sh ../../$@)
+$(JSLIB):
+		dune build wast.bc.js
+		cp ../_build/default/interpreter/wast.bc.js ./wast.js
```
rossberg, line 114: Can you introduce proper build rules for wast.bc.js, so that dependencies and cleanup work?

interpreter/meta/jslib/wast.ml
```diff
@@ -0,0 +1,25 @@
...
+        let bs =
+          match def.Source.it with
+          | Script.Textual m -> (Encode.encode m)
+          | Script.Encoded (_, bs) -> bs in
+        let buf = new%js Typed_array.arrayBuffer (String.length bs) in
+        let u8arr = new%js Typed_array.uint8Array_fromBuffer buf in
+        String.iteri (fun i c -> Typed_array.set u8arr i (int_of_char c)) bs; buf
+
+      method decode buf = decode buf 80
+      method decode2 buf width = decode buf width
```
rossberg, line 23: Can we drop this duplication and always require the width? I don't think the minor convenience justifies it, and 80 is a fairly random default these days.

interpreter/meta/jslib/wast.ml
```diff
@@ -0,0 +1,25 @@
+open Js_of_ocaml
```
rossberg, line 1: If you introduce this as wast.ml, shouldn't you remove the prior wasm.ml?

rossberg (commented):
Thanks for looking into this! I'm on board with replacing BuckleScript.

interpreter/Makefile
takikawa: This should be addressed now. I redid the Makefile rule to avoid using dune and just use `ocamlbuild` and calling `js_of_ocaml` directly (there is an ocamlbuild plugin for `js_of_ocaml` but I was unable to get it to work correctly). It should do cleanup now for example.

Happy to fine tune the build rules if anything looks off still.

interpreter/meta/jslib/wast.ml
takikawa: Done.

interpreter/meta/jslib/wast.ml
takikawa: I've now removed `wasm.ml` and associated build files in the latest commits.

takikawa:
I think the latest push should address the review comments. In addition, I modified the README to match the new setup and added CI integration that will build `wast.js` along with the normal interpreter build.

interpreter/README.md
```diff
@@ -160,6 +163,20 @@ WebAssemblyText.decode(binary)
 // )
 ```
 
+Depending on how you load the library, the object may be accessed in different ways. For example, using `require` in node.js:
+
+```
+let wast = require("./wast.js");
+let binary = wast.WebAssemblyText.encode("(module"));
```
rossberg, line 170:
```suggestion
let binary = wast.WebAssemblyText.encode("(module)");
```

interpreter/README.md
```diff
@@ -160,6 +163,20 @@ WebAssemblyText.decode(binary)
...
+```
+
+Or using `load` from a JavaScript shell:
+
+```
+load("./wast.js");
+let binary = WebAssemblyText.encode("(module"));
```
rossberg, line 177:
```suggestion
let binary = WebAssemblyText.encode("(module)");
```

interpreter/meta/jslib/wast.ml
```diff
@@ -0,0 +1,26 @@
...
+        let bs =
+          match def.Source.it with
+          | Script.Textual m -> (Encode.encode m)
+          | Script.Encoded (_, bs) -> bs
+          | Script.Quoted (_, _) -> failwith "Unsupported" in
+        let buf = new%js Typed_array.arrayBuffer (String.length bs) in
+        let u8arr = new%js Typed_array.uint8Array_fromBuffer buf in
+        String.iteri (fun i c -> Typed_array.set u8arr i (int_of_char c)) bs; buf
+
+      method decode (buf : Typed_array.arrayBuffer Js.t) width : (Js.js_string Js.t) =
```
rossberg, line 21: Perhaps by declaring the parameter
```
(optwidth : int Js.optdef)
```
and then doing
```
let width = Js.Optdef.get optwidth (fun () -> 80) in
```
it is possible to emulate the previous default parameter?

rossberg (commented):
This is great!

interpreter/Makefile
```diff
@@ -108,14 +109,12 @@ _build/$(LIB).cmxa: $(FILES) $(LIB).mllib _tags Makefile
...
-.PHONY:		$(JSLIB)
-$(JSLIB):	$(UNOPT)
-		mkdir -p _build/jslib/src
-		cp meta/jslib/* _build/jslib
-		cp $(DIRS:%=_build/%/*.ml*) meta/jslib/*.ml _build/jslib/src
-		rm _build/jslib/src/*.ml[^i]
-		(cd _build/jslib; ./build.sh ../../$@)
+.INTERMEDIATE:	$(JSLIB).byte
+$(JSLIB).byte:	meta/jslib/$(JSLIB).ml
+		$(OCBA) -use-ocamlfind -pkg js_of_ocaml -pkg js_of_ocaml-ppx meta/jslib/$(JSLIB).byte
```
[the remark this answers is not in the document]
takikawa: I was able to refactor this a bit so that it uses `$@` as the target now.

rossberg (approved):
Sorry, missed your last reply.

LGTM, thanks!

interpreter/meta/jslib/wast.ml
hhugo: You can achieve that behavior using JS.Unsafe.* functions.
Maybe with `val decode = Js.Unsafe.callback (fun bud width -> ...)`

PR #1512, opened 2022-08-01
Minor copy edit suggestions

surma:
I properly read the spec and found some minor suggestions I wanted to provide for your consideration.

### DIFF
--- a/document/core/syntax/modules.rst
+++ b/document/core/syntax/modules.rst
@@ -328,7 +328,7 @@ The |MSTART| component of a module declares the :ref:`function index <syntax-fun
 
 .. note::
    The start function is intended for initializing the state of a module.
-   The module and its exports are not accessible to the host before this initialization has completed.
+   The module and its exports are not accessible externally before this initialization has completed.
 
 
 .. index:: ! export, name, index, function index, table index, memory index, global index, function, table, memory, global, instantiation
--- a/document/core/syntax/modules.rst
+++ b/document/core/syntax/modules.rst
@@ -328,7 +328,7 @@ The |MSTART| component of a module declares the :ref:`function index <syntax-fun
 
 .. note::
    The start function is intended for initializing the state of a module.
-   The module and its exports are not accessible before this initialization has completed.
+   The module and its exports are not accessible to the host before this initialization has completed.
 
 
 .. index:: ! export, name, index, function index, table index, memory index, global index, function, table, memory, global, instantiation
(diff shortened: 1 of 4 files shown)


document/core/valid/types.rst
```diff
@@ -279,6 +279,8 @@ Limits
      \vdashlimitsmatch \{ \LMIN~n_1, \LMAX~m_1 \} \matcheslimits \{ \LMIN~n_2, \LMAX~m_2 \}
    }
 
+.. note::
+  In this context, read :math:`\matcheslimits` as “is the same type as or a subtype of”.
```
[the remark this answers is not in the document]
rossberg: Makes sense to point this out, but given that the subtyping symbol is used for all relations in this section, this seems like the wrong place to put this note. Instead of this note, I'd suggest changing line 243 to the following:
```
In some cases, this allows for a simple form of subtyping (written ":math:`\matches`" formally), as defined here.
```

rossberg (commented):
Thanks, makes sense to me, but see comment.

document/core/syntax/modules.rst
```diff
@@ -328,7 +328,7 @@ The |MSTART| component of a module declares the :ref:`function index <syntax-fun
 
 .. note::
    The start function is intended for initializing the state of a module.
-   The module and its exports are not accessible before this initialization has completed.
+   The module and its exports are not accessible to the host before this initialization has completed.
```
rossberg, line 331: Depending on the environment, it's not just the host, but other Wasm modules may be clients as well.
```suggestion
   The module and its exports are not accessible to clients before this initialization has completed.
```

surma:
Thanks for taking the time to review. I implemented your comments.

document/core/valid/types.rst
```diff
@@ -240,7 +240,7 @@ Import Subtyping
 
 When :ref:`instantiating <exec-module>` a module,
 :ref:`external values <syntax-externval>` must be provided whose :ref:`types <valid-externval>` are *matched* against the respective :ref:`external types <syntax-externtype>` classifying each import.
-In some cases, this allows for a simple form of subtyping, as defined here.
+In some cases, this allows for a simple form of subtyping (written ":math:`\matchesexterntype`" formally), as defined here.
```
rossberg, line 243:
```suggestion
In some cases, this allows for a simple form of subtyping (written ":math:`\matches`" formally), as defined here.
```

rossberg: Sorry, scratch that, that only works on a later repo.

rossberg (approved):

document/core/valid/types.rst
surma: Yeah I tried it and it doesn’t work, which is why I went for this one 😅

PR #1516, opened 2022-08-08
[spec] Fix typing rule for module instances

rossberg:
The appendix was not updated for the bulk op extension, so missed the addition of elem and data components to the context.

### DIFF
--- a/document/core/appendix/properties.rst
+++ b/document/core/appendix/properties.rst
@@ -285,7 +285,7 @@ Module instances are classified by *module contexts*, which are regular :ref:`co
 
   * The :ref:`reference <syntax-ref>` :math:`\reff_i` must be :ref:`valid <valid-ref>` with :ref:`reference type <syntax-reftype>` :math:`t`.
 
-* Then the table instance is valid with :ref:`reference type <syntax-reftype>` :math:`t`.
+* Then the element instance is valid with :ref:`reference type <syntax-reftype>` :math:`t`.
 
 .. math::
    \frac{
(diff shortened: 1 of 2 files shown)


document/core/appendix/properties.rst
```diff
@@ -346,7 +346,7 @@ Module instances are classified by *module contexts*, which are regular :ref:`co
 
 * For each :ref:`element address <syntax-elemaddr>` :math:`\elemaddr_i` in :math:`\moduleinst.\MIELEMS`, the :ref:`element instance <syntax-eleminst>` :math:`S.\SELEMS[\elemaddr_i]` must be :ref:`valid <valid-eleminst>`.
 
-* For each :ref:`data address <syntax-dataaddr>` :math:`\dataaddr_i` in :math:`\moduleinst.\MIDATAS`, the :ref:`data instance <syntax-datainst>` :math:`S.\SDATAS[\dataaddr_i]` must be :ref:`valid <valid-datainst>`.
+* For each :ref:`data address <syntax-dataaddr>` :math:`\dataaddr_i` in :math:`\moduleinst.\MIDATAS`, the :ref:`data instance <syntax-datainst>` :math:`S.\SDATAS[\dataaddr_i]` must be :ref:`valid <valid-datainst>` with some :ref:`reference type <syntax-reftype>` :math:`\reftype_i`.
```
conrad-watt, line 349: I think a small muddle here. The elems in the line above should be valid with some reference type; the datas here should be checked to have length `n` to match the datacount section.

conrad-watt (changes requested):

document/core/appendix/properties.rst
rossberg: Oops, yes, moved the new text to elem.

The data count section isn't relevant here, as it's handled as a detail of the binary encoding, not appearing in the AST.

rossberg:
Just noticed that store validity needed a tweak to match the eleminst judgement as well. Done.

conrad-watt (approved):
lgtm

PR #1517, opened 2022-08-10
[spec] Simplify exec rule for if

rossberg:
This is how the interpreter implemented it, which should be in sync. It's the simpler formulation, so changing the spec instead of the interpreter.


conrad-watt (approved):
LGTM

PR #1518, opened 2022-08-10
[interpreter] use a correct module_inst /w invoke

nagisa:
With an `empty_module_inst` the very first `Invoke` instruction would be
executed with an empty module instance before the handling of the
`Invoke` would grab the instance stored within the function.

Today this isn’t doesn’t produce a wrong result – execution of the
`invoke` does not depend on access to the “correct” module instance.
This is true by the virtue of `Invoke` instruction, and the `invoke`
function, referencing a `func_inst`. However, if they did reference a
`funcaddr` instead, as prescribed in the specification, having an access
to the correct store / module instance would be a hard requirement for the
correct operation.

### DIFF
--- a/interpreter/exec/eval.ml
+++ b/interpreter/exec/eval.ml
@@ -668,14 +668,14 @@ let rec eval (c : config) : value stack =
 
 (* Functions & Constants *)
 
-let invoke (func : func_inst) (vs : value list) : value list =
+let invoke (mod_inst : module_inst) (func : func_inst) (vs : value list) : value list =
   let at = match func with Func.AstFunc (_, _, f) -> f.at | _ -> no_region in
   let FuncType (ins, out) = Func.type_of func in
   if List.length vs <> List.length ins then
     Crash.error at "wrong number of arguments";
   if not (List.for_all2 (fun v -> (=) (type_of_value v)) vs ins) then
     Crash.error at "wrong types of arguments";
-  let c = config empty_module_inst (List.rev vs) [Invoke func @@ at] in
+  let c = config mod_inst (List.rev vs) [Invoke func @@ at] in
   try List.rev (eval c) with Stack_overflow ->
     Exhaustion.error at "call stack exhausted"
 
--- a/interpreter/exec/eval.mli
+++ b/interpreter/exec/eval.mli
@@ -7,4 +7,4 @@ exception Crash of Source.region * string
 exception Exhaustion of Source.region * string
 
 val init : Ast.module_ -> extern list -> module_inst (* raises Link, Trap *)
-val invoke : func_inst -> value list -> value list (* raises Trap *)
+val invoke : module_inst -> func_inst -> value list -> value list (* raises Trap *)
--- a/interpreter/script/run.ml
+++ b/interpreter/script/run.ml
@@ -354,7 +354,7 @@ let run_action act : Values.value list =
         if Values.type_of_value v.it <> t then
           Script.error v.at "wrong type of argument"
       ) vs ins;
-      Eval.invoke f (List.map (fun v -> v.it) vs)
+      Eval.invoke inst f (List.map (fun v -> v.it) vs)
     | Some _ -> Assert.error act.at "export is not a function"
     | None -> Assert.error act.at "undefined export"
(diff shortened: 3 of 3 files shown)


nagisa:
If you think it would be a good idea to change the implementation of `Invoke` to refer to `funcaddr` rather than `func_inst`, I’d be happy to explore adjusting the interpreter for that as well.

conrad-watt:
I think there's a slight confusion here, which is caused by the interpreter diverging from the formal specification. As far as the spec is concerned, the index of the `invoke` instruction refers to the `funcinst` at the indexed position in the _`funcs`_ field of the [store](https://webassembly.github.io/spec/core/exec/runtime.html#store), in contrast with `call`, which references a position in the _`funcaddrs`_ field of the current [instance](https://webassembly.github.io/spec/core/exec/runtime.html#module-instances) (with the indexed `funcaddr` in turn referring to a `funcinst` in the store).

In particular, you can see how the execution of `call` is defined in terms of `invoke` [here](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-call-x). Note the difference between the `call` index `x` and the `funcaddr` `a` used by `invoke`.

The reference interpreter chooses not to explicitly represent the global store, instead using the OCaml heap as an "implicit" store, with indices into the store replaced with references to OCaml allocations. So the idea of the `Invoke` instruction as present in the interpreter is that the reference to the `func_inst` should be thought of as standing in for an index into the Wasm global store where the spec says that `funcinst`s live

If we wanted to change the interpreter's `Invoke` to use a `funcaddr` and be more in line with the spec, we'd need to explicitly represent the Wasm global store in the state of the interpreter. The `funcaddr` is _not_ an index into the instance - `invoke` works directly with respect to the store and is instance-independent (in contrast to `call`), so having it execute in the context of a different instance should have no effect either way.

TL:DR: I don't think this change is correct, but the confusion is understandable given how the interpreter diverges from the spec

rossberg:
Yeah, not making the store explicit in the interpreter in favour of a "shallow embedding" that equates spec addresses with OCaml references was a conscious choice to avoid too much clutter – as well as more quadratic behaviour that might make it unfit to even run the test suite :).

nagisa:
> The reference interpreter chooses not to explicitly represent the global store, instead using the OCaml heap as an "implicit" store, with indices into the store replaced with references to OCaml allocations.

Ah, that’s probably the large chunk of my misunderstanding here. I was largely conflating the store with the `module_inst` rather than with the global heap.

> TL:DR: I don't think this change is correct, but the confusion is understandable given how the interpreter diverges from the spec

I see why a bulk of the reasoning behind this change was flawed, but from what I can tell it doesn’t invalidate the change itself. Passing in a proper `module_inst` to `invoke` still seems like strictly an improvement to me. The interpreter is no longer conjuring random `module_inst`s from the thin air after these changes and is no longer implicitly relying on the fact that executing an `Invoke` doesn’t actually inspect the instance. Is there something else I’m missing?

To give more context behind the motivation for this change: I’m exploring changes downstream that would result in execution of `Invoke` needing an access to the corresponding `module_inst`. That is why I submitted this PR, so if there is a reason why this change is wrong in isolation of the incorrect motivation presented above, I would appreciate an explanation a lot.

rossberg:
Well, you are supposed to be able to invoke a function externally, from outside any instance. The function would no longer model this properly if it required an instance parameter. The empty module essentially models the empty global scope.

nagisa:
Aha, got it.

conrad-watt:
We've tended to specify "top level" things as executing inside an empty instance (I think this becomes more explicit with the threads proposal), so having the interpreter mirror this seems more consistent to me and communicates that `invoke` is instance-independent. Executing the function call inside the instance of the exporter might make someone believe that the execution of `invoke` relies on this instance.

I don't know the details of the extension you're experimenting with, but my knee-jerk reaction would be that such reliance on an instance should be kept away from `invoke` and instead handled in the instructions that produce it (e.g. `call`). My email is in my GH profile and I'd be happy to talk in more detail if you'd like to drop me a line.

PR #1531, opened 2022-09-16
fix unpacked example

whirlicote:
please feel free to ignore this pull request in case of a misunderstanding on my part


rossberg:
Yes, I think this is a misunderstanding. The purpose of `unpacked` isn't to extract the lane type, but to map a lane's packed integer type (i8 or i16) to an unpacked proper type (i32 or i64).

whirlicote:
Ah I see. I missed that when two rules contradict each other the one that is written first gets chosen:

```
unpacked(i8x16)=i32
```
contradicts to
```
unpacked(txN)=t
```
if `t` is replaced with `i8`.

rossberg:
Yes, rules are tried in order until first match.

PR #1532, opened 2022-09-16
The variable t looks like it is not defined in the exaplanation text.

whirlicote:
In the other definitions for allocations type is defined as the type of the thing that is to be allocated


rossberg (approved):
Thanks!

PR #1535, opened 2022-09-20
[js-api] Try to clarify NaN/infinity handling.

### LINKED ISSUE #1496: [js-api] Conversion from JS values to floats doesn't specify how to handle NaNs
The current algorithm within `ToWebAssemblyValue` to convert from JS values to floats looks like this:

> If type is [f64](https://webassembly.github.io/spec/core/syntax/types.html#syntax-numtype),
> > Let f64 be ? [ToNumber](https://tc39.github.io/ecma262/#sec-tonumber)(v).
> > Return [f64.const](https://webassembly.github.io/spec/core/exec/runtime.html#values) f64.

It's very unclear how NaNs should be handled here. `ToNumber` returns a [JS number](https://tc39.es/ecma262/#sec-ecmascript-language-types-number-type), which isn't quite the same as a regular `f64`; it's only defined to have a single NaN value, with an unknown bit pattern. The spec should say something about how that NaN value gets converted, whether it's defined to result in a particular value or just left as implementation-defined.

The actual implementation of this currently differs between browsers. Using this code to test:

```wat
;; float-bits.wasm
(module
    (func $f64toi64 (param $float f64) (result i64)
        local.get $float
        i64.reinterpret_f64
    )
    (func $i64tof64 (param $bits i64) (result f64)
        local.get $bits
        f64.reinterpret_i64
    )
    (export "f64toi64" (func $f64toi64))
    (export "i64tof64" (func $i64tof64))
)

```

```js
const wasm = fetch("./float-bits.wasm");
const { instance } = await WebAssembly.instantiateStreaming(wasm);
const { f64toi64, i64tof64 } = instance.exports;

document.body.textContent = f64toi64(i64tof64(0x7ff8000000001234n)).toString(16);
```

Chromium preserves the bits and shows `7ff8000000001234`, whereas Firefox and Safari pick a single NaN bit pattern of `7ff8000000000000`.

---

In the opposite direction, `ToJSValue` also doesn't specify how NaNs should be handled:

> 5. If w is of the form [f64.const](https://webassembly.github.io/spec/core/exec/runtime.html#values) f64, return [the Number value](https://tc39.github.io/ecma262/#sec-ecmascript-language-types-number-type) for f64.

I don't find this as bad, since it's pretty obvious that any NaN should map to the JS NaN, but it wouldn't hurt to add a note about it here too.

Ms2ger:
Fixes #1496.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -170,6 +170,7 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
         text: −∞
         text: nan
         text: canon
+        text: signif
     text: function element; url: exec/runtime.html#syntax-funcelem
     text: import component; url: syntax/modules.html#imports
     text: external value; url: exec/runtime.html#syntax-externval
@@ -1134,7 +1135,7 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
 1. If |type| is [=f32=],
     1. Let |number| be [=?=] [=ToNumber=](|v|).
     1. If |number| is **NaN**,
-        1. Let |n| be an implementation-defined integer such that [=canon=]<sub>32</sub> ≤ |n| < 2<sup>32</sup>.
+        1. Let |n| be an implementation-defined integer such that [=canon=]<sub>32</sub> ≤ |n| < 2<sup>[=signif=](32)</sup>.
         1. Let |f32| be [=nan=](n).
     1. Otherwise,
         1. Let |f32| be |number| rounded to the nearest representable value using IEEE 754-2008 round to nearest, ties to even mode. [[IEEE-754]]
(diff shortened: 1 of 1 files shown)


Ms2ger:
CC @Liamolucko @gahaas @syg

document/js-api/index.bs
```diff
@@ -1115,10 +1128,20 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+        1. Let |n| be an implementation-defined integer such that [=canon=]<sub>32</sub> ≤ |n| < 2<sup>32</sup>.
+        1. Let |f32| be [=nan=](n).
+    1. Otherwise,
+        1. Let |f32| be |number| rounded to the nearest representable value using IEEE 754-2008 round to nearest, ties to even mode. [[IEEE-754]]
     1. Return [=f32.const=] |f32|.
 1. If |type| is [=f64=],
-    1. Let |f64| be ? [=ToNumber=](|v|).
+    1. Let |number| be ? [=ToNumber=](|v|).
+    1. If |number| is **NaN**,
+        1. Let |n| be an implementation-defined integer such that [=canon=]<sub>64</sub> ≤ |n| < 2<sup>64</sup>.
```
Liamolucko, line 1141: Shouldn't this be < 2<sup>52</sup>, not 2<sup>64</sup>?

Ms2ger: I misread *M* as *N* in the core spec, thanks for pointing that out.

rossberg (approved):
Sorry, I seem to have missed this. LGTM

PR #1537, opened 2022-09-23
[web-api] Correct MIME type check.

### LINKED ISSUE #1138: [web-api] MIME type check in "compile a potential WebAssembly response" imprecise and ambiguous
> 2. Let mimeType be the result of extracting a MIME type from response’s header list.
>
> 3. If mimeType is not `application/wasm`, reject returnValue with a TypeError and abort these substeps.

Here _mimeType_ is a [MIME type record](https://mimesniff.spec.whatwg.org/#mime-type), as returned by [extract a MIME type](https://fetch.spec.whatwg.org/#concept-header-extract-mime-type).

However in step 3 it's compared against a byte sequence.

It's unclear how to interpret the intended check here:

- Do you want to get the raw value of the `Content-Type` header and check that against the byte sequence? In this case `application/wasm;` would not match.
- Do you want to serialize the MIME type and compare it against the string (not byte sequence) "application/wasm"? In this case `application/wasm;` would match, as would <code>&nbsp;&nbsp;&nbsp;application/wasm&nbsp;&nbsp;&nbsp;</code>.
- Do you want to check against the MIME type's essence? In that case `application/wasm;foo=bar` would match.
- Note that for any case except getting the raw value of the Content-Type header, using the "extract a MIME type" algorithm means that multiple headers (or, equivalently, one header with comma-separate values) will be treated forgivingly. E.g. `application/javascript, application/wasm` and `application/wasm, unparseable` would both end up as `application/wasm`.

https://github.com/web-platform-tests/wpt/blob/master/wasm/webapi/contenttype.any.js seems to pick "check the raw value of the `Content-Type` header".

/cc @Surma @annevk

Ms2ger:
Fixes #1138.

### DIFF
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -122,8 +122,7 @@ Note: This algorithm accepts a {{Response}} object, or a
         1. Let |mimeType| be the result of [=header list/getting=] `` `Content-Type` `` from |response|'s [=response/header list=].
         1. If |mimeType| is null, reject |returnValue| with a {{TypeError}} and abort these substeps.
         1. Remove all [=HTTP tab or space byte=] from the start and end of |mimeType|.
-        1. [=ASCII lowercase=] |mimeType|.
-        1. If |mimeType| is not `` `application/wasm` ``, reject |returnValue| with a {{TypeError}} and abort these substeps.
+        1. If |mimeType| is not a [=byte-case-insensitive=] match for `` `application/wasm` ``, reject |returnValue| with a {{TypeError}} and abort these substeps.
 
             Note: extra parameters are not allowed, including the empty `` `application/wasm;` ``.
 
--- a/document/web-api/index.bs
+++ b/document/web-api/index.bs
@@ -76,6 +76,7 @@ url:https://w3c.github.io/webappsec-secure-contexts/#environment-settings-object
 
 <pre class='link-defaults'>
 spec:ecma-262; type:exception; for:ECMAScript; text:TypeError
+spec:fetch; type:dfn; text:get
 spec:webidl; type:dfn; text:resolve
 </pre>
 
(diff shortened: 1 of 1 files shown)


rossberg (approved):
This looks reasonable to me, but my knowledge about how to spec this kind of thing is limited.

document/web-api/index.bs
```diff
@@ -118,7 +119,9 @@ Note: This algorithm accepts a {{Response}} object, or a
     1. Let |returnValue| be [=a new promise=]
     1. [=Upon fulfillment=] of |source| with value |unwrappedSource|:
         1. Let |response| be |unwrappedSource|'s [=Response/response=].
-        1. Let |mimeType| be the result of [=extracting a MIME type=] from |response|'s [=response/header list=].
+        1. Let |mimeType| be the result of [=getting=] `` `Content-Type` `` from |response|'s [=response/header list=].
```
annevk, line 122: Use for=header list to avoid confusion in the future?

document/web-api/index.bs
```diff
@@ -118,7 +119,9 @@ Note: This algorithm accepts a {{Response}} object, or a
...
+        1. Remove all [=HTTP tab or space=] from the start and end of |mimeType|.
```
annevk, line 123: What if _mimeType_ is null?

annevk: Also, if it's not null it's a byte sequence...

andreubotella: Also, header values should not have leading or trailing HTTP tabs or spaces, since those [make the header value invalid](https://fetch.spec.whatwg.org/#header-value), and are stripped when appending from the `Headers` interface.

document/web-api/index.bs
```diff
@@ -118,7 +119,9 @@ Note: This algorithm accepts a {{Response}} object, or a
...
+        1. [=ASCII lowercase=] |mimeType|.
```
andreubotella, line 124: You might want to use "if `mimeType` is not a [byte-case-insensitive](https://infra.spec.whatwg.org/#byte-case-insensitive) match".

document/web-api/index.bs
annevk: That happens when you set the values, but that's not necessarily true for something coming from the server. Stripping as OP did is correct. Just need to strip bytes and not code points.

document/web-api/index.bs
annevk: You need to use this suggestion or use byte-lowercase.

annevk (approved):

annevk:
Thanks for getting this fixed @Ms2ger!

PR #1546, opened 2022-10-05
test/meta/generate_*.js: sync upstream JS with tests

keithw:
The error strings in the core testsuite were cleaned up in https://github.com/WebAssembly/reference-types/pull/117 (see https://github.com/WebAssembly/reference-types/issues/113), but some of the tests (for table/memory init and copy, and memory_fill) are auto-generated by upstream JS that wasn't edited at the same time.

The PR syncs the JS scripts so that they will generate the current tests.


rossberg (approved):
Thanks!

keithw:
I assumed it was a cat!

PR #1548, opened 2022-10-05
[interpreter] Remove duplicated token declarations.

dhil:
This patch tidies up the `parser.mly` module by removing two instances of `%prec` that were never useful. More importantly, this patch also removes duplicated token declarations as `menhir` considers such an error in debug mode. Consequently, it is now possible to debug the grammar using `menhir --explain` or similar.


dhil:
I guess that may because I am first time contributor to this particular repository...

PR #1550, opened 2022-10-12
[js-api] Expose everywhere

Ms2ger:


Ms2ger:
This exposes the APIs in ShadowRealms; Firefox and V8 already do this. I didn't check WebKit.

rossberg (approved):
Not sure what this does, but LGTM :)

PR #1555, opened 2022-10-25
[test] Fix tests in binary-leb128 and simd_lane

eutro:
This PR fixes some tests in a couple scripts, where the module contains multiple errors, presumably by mistake.

- `binary-leb128` test had the wrong section length

- `simd_lane` tests had the wrong argument to f32x4.replace_lane


rossberg (approved):
Thanks!

PR #1561, opened 2022-11-11
[test] elem.wast: force to use exprs in a element

yamt:
This change makes the test detect this bug in toywasm. https://github.com/yamt/toywasm/commit/89465d57a8ad07a2e159711c92d3625bf66c41c7

(without this change, all tests were passing even with the bug.)

Note: toywasm CI uses tests converted with wast2json.

Note: wast2json doesn't use exprs unless the element contains exprs other than ref.func.
https://github.com/WebAssembly/wabt/blob/712bb5c908c35de7e4878463bb26ca0c89dac479/src/ir.cc#L680-L682

### DIFF
--- a/test/core/elem.wast
+++ b/test/core/elem.wast
@@ -148,6 +148,25 @@
 (assert_return (invoke "call-7") (i32.const 65))
 (assert_return (invoke "call-9") (i32.const 66))
 
+;; Same as the above, but use ref.null to ensure the elements use exprs.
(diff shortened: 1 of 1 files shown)


rossberg:
Thanks, but can you perhaps test this case with a separate module?

yamt:
> Thanks, but can you perhaps test this case with a separate module?

do you mean something like the last push?

rossberg (approved):
Thanks!

PR #1564, opened 2022-12-07
[spec] Fix minor errors and inconsistencies

rossberg:
This fixes a bunch of minor oversights and notational inconsistencies that I discovered while writing up a rule summary. Most of them occur in the execution rules for bulk instructions.


conrad-watt (approved):
lgtm

PR #1567, opened 2022-12-12
[interpreter] Refactor parser to handle select & call_indirect correctly

### LINKED ISSUE #1566: interpreter: fails to parse a valid-looking wat
```wast
(module
  (type (func (param i32) (result i32)))
  (func (type 0) (param i32) (result i32)
    i32.const 2
    i32.const 3
    i32.const 0
    i32.const 1
    local.get 0
    select
    select
  )
)
```

is a different, but equivalent, way to write https://github.com/WebAssembly/spec/blob/937fc7d63efb9e18f36334a9e761a8a040ac44b7/test/core/select.wast#L47-L49

but it fails to parse with the following error when fed to the reference interpreter

```
11.3-11.4: syntax error: unexpected token
```

If we parenthesize the last select, the parse succeeds. This appears to be the only such function in the spec-test suite where such an equivalent rewrite fails to parse in the test suite, as far as I can tell.

rossberg:
Fix #1566.

### DIFF
--- a/test/core/select.wast
+++ b/test/core/select.wast
@@ -512,3 +512,20 @@
   "type mismatch"
 )
 
+
+;; Flat syntax
+
+(module
+  (table 1 funcref)
+  (func (result i32) unreachable select)
+  (func (result i32) unreachable select nop)
+  (func (result i32) unreachable select (select))
+  (func (result i32) unreachable select select)
+  (func (result i32) unreachable select select select)
+  (func (result i32) unreachable select (result i32))
+  (func (result i32) unreachable select (result i32) (result))
+  (func (result i32) unreachable select (result i32) (result) select)
+  (func (result i32) unreachable select (result) (result i32) select (result i32))
+  (func (result i32) unreachable select call_indirect)
+  (func (result i32) unreachable select call_indirect select)
+)
--- a/interpreter/text/parser.mly
+++ b/interpreter/text/parser.mly
@@ -366,12 +366,16 @@ align_opt :
 
 /* Instructions & Expressions */
 
-instr :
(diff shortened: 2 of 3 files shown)


nagisa:
I’ll test this out shortly (hopefully tomorrow.)

tlively (approved):
rubberstamp lgtm

test/core/select.wast
```diff
@@ -512,3 +512,20 @@
...
+  (table 1 funcref)
+  (func (result i32) unreachable select)
+  (func (result i32) unreachable select nop)
+  (func (result i32) unreachable select (select))
+  (func (result i32) unreachable select select)
+  (func (result i32) unreachable select select select)
+  (func (result i32) unreachable select (result i32))
+  (func (result i32) unreachable select (result i32) (result))
+  (func (result i32) unreachable select (result i32) (result) select)
+  (func (result i32) unreachable select (result) (result i32) select (result i32))
```
yamt, line 528: wast2json 1.0.31 can't parse this. is it expected?
```
../select.wast:526:55: error: unexpected token "result", expected an instr.
  (func (result i32) unreachable select (result i32) (result))
                                                      ^^^^^^
../select.wast:527:55: error: unexpected token "result", expected an instr.
  (func (result i32) unreachable select (result i32) (result) select)
                                                      ^^^^^^
../select.wast:528:51: error: unexpected token "result", expected an instr.
...nc (result i32) unreachable select (result) (result i32) select (result i32))
                                                ^^^^^^
```

yamt: https://github.com/WebAssembly/wabt/pull/2104

PR #1575, opened 2023-01-10
Update the 'spectest' module definition

devoncarew:
This updates the the 'spectest' module definition from the readme. I'd tried to compile it via wat2wasm and got errors; w/ these changes the module now compiles (`wat2wasm` v1.0.31).


rossberg:
Ah, these are actually meant to specify module _types_. Those do not currently exist in Wasm, but the component model defines an [analogous syntax](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions). Types of course do not contain initialiser values (or function bodies, for that matter).

devoncarew:
OK, gotcha, sounds like the docs for the spectest module are WAI? I'll close this PR then.

PR #1577, opened 2023-01-19
[spec] Fix abstract grammar for const immediate

### LINKED ISSUE #1570: [spec] br_table has unspecified semantics for negative branch target
```
(module
  (func (export "test") 
    (block $default
      (block $1
        (block $2
          (br_table $1 $2 $default (i32.const -5))))))

  )

(assert_return (invoke "test"))
```
The spec says `l*` for the `br_table` instruction here is `[$1 $2]`, with indices [0, 1], and `l_N` is `$default`. The two rules for `br_table` based on the value `i`, which is `-5` here, are:
```
br l_i ;; if i < len(l*)
br l_N ;; otherwise
```
In this case, `i < len(l*)` is true since -5 < 2, which would mean the machine would try to lookup `l*[-5]`, which is meaningless.

The reference spec interpreter seems use the `$default` label here, which suggests it requires that i >= 0 before the `br l_i` rule is invoked.
```
br l_i ;; if i >= 0 && i < len(l*)
```

rossberg:
Fix #1570.

### DIFF
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -45,7 +45,7 @@ These operations closely match respective operations available in hardware.
    \production{signedness} & \sx &::=&
      \K{u} ~|~ \K{s} \\
    \production{instruction} & \instr &::=&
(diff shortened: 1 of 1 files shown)


conrad-watt (approved):
lgtm

PR #1578, opened 2023-01-19
[spec] Fix context composition in text format

### LINKED ISSUE #1574: Spec is unclear on function body identifier context
As far as I can tell, the text format spec doesn’t explain how a module’s initial identifier context should reach the body of a function definition, which I assume it must somehow.

The [`func` production](https://webassembly.github.io/spec/core/text/modules.html#functions) synthesises an identifier context $I''$ for the function body by composing $I'$ (synthesised by `typeuse`) with a context containing the function definition’s local variable identifiers, so the inherited attribute $I$ containing the module’s initial context is not directly used. But the [`typeuse` production](https://webassembly.github.io/spec/core/text/modules.html#type-uses) itself also discards $I$: the $I'$ in its synthesised attribute contains only the parameter identifiers. The only syntactic occurrence of $I$ in these productions is when `typeuse` looks up the definition of the function’s type.

It’s therefore unclear how the remaining contents of $I$ are made available to the function body. The prose accompanying the `typeuse` production does mention “the _updated_ identifier context including possible parameter identifiers” (emphasis mine), which perhaps implies that $I'$ is intended to be formed by extending the inherited attribute $I$ rather than discarding it (e.g. $I' = I \oplus \lbrace \mathsf{locals}~(\epsilon)^n \rbrace$ etc), but that’s just a guess.

How is this intended to work? Can the spec be updated to clarify?

rossberg:
Fix #1574.


conrad-watt (approved):
lgtm

PR #1579, opened 2023-01-19
[spec] Fix label shadowing

### LINKED ISSUE #1572: Spec is unclear on label shadowing
As discussed in #210 and #303, it’s not clear from the spec whether a nested block may redefine (“shadow”) a label from one of its ancestors, although the test introduced by #216 makes it unambiguous that this is allowed in practice.

Subjectively, I got the impression from reading the spec that label shadowing is explicitly _not_ allowed, because:

1. the production which synthesises an updated identifier context for labels has a [side condition](https://webassembly.github.io/spec/core/text/instructions.html#labels) which prevents the use of any identifier which already appears in the labels index space of the current context; and
2. [well-formedness](https://webassembly.github.io/spec/core/text/conventions.html#text-context) of identifier contexts prevents duplicate identifiers in all index spaces.

I can explain away point 2 by noting that — itself somewhat counterintuitively — well-formedness appears to be asserted only selectively, i.e. in special circumstances where we expect the labels index space to be empty anyway, but I have no explanation for point 1.

It’s entirely possible I’ve misunderstood the spec, but either way, might it be clarified to make the situation more explicit?

rossberg:
Fix #1572.

### DIFF
--- a/document/core/text/instructions.rst
+++ b/document/core/text/instructions.rst
@@ -34,6 +34,8 @@ The following grammar handles the corresponding update to the :ref:`identifier c
    \production{label} & \Tlabel_I &::=&
      v{:}\Tid &\Rightarrow& \{\ILABELS~v\} \compose I
        & (\iff v \notin I.\ILABELS) \\ &&|&
+     v{:}\Tid &\Rightarrow& \{\ILABELS~v\} \compose (I \WITH \ILABELS[i] = \epsilon)
+       & (\iff I.\ILABELS[i] = v) \\ &&|&
      \epsilon &\Rightarrow& \{\ILABELS~(\epsilon)\} \compose I \\
    \end{array}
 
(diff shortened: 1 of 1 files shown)


document/core/text/instructions.rst
```diff
@@ -34,6 +34,8 @@ The following grammar handles the corresponding update to the :ref:`identifier c
    \production{label} & \Tlabel_I &::=&
      v{:}\Tid &\Rightarrow& \{\ILABELS~v\} \compose I
        & (\iff v \notin I.\ILABELS) \\ &&|&
+     v{:}\Tid &\Rightarrow& \{\ILABELS~v\} \compose (I \WITH \ILABELS[i] = \epsilon)
```
tomstuart, line 37:
```suggestion
     v{:}\Tid &\Rightarrow& \{\ILABELS~v\} \compose (I \with \ILABELS[i] = \epsilon)
```

conrad-watt (approved):
lgtm

PR #1582, opened 2023-01-23
Test the abbreviated forms of table.{get,set,size,grow,fill}

keithw:
These weren't being exercised in the testsuite, and some parsers (including WABT) can't currently parse them.

(The abbreviations for table.copy and table.init are already tested in bulk.wast.)


rossberg (approved):
Thanks!

PR #1584, opened 2023-01-27
[spec] Fix types of v128.store{8,16,32,64}_lane in index

candymate:
Mistakes in types of v128.store{8,16,32,64}_lane in Appendix: instruction index. Types should be `[i32, v128] -> []` instead of `[i32, v128] -> [v128]`.


ngzhian:
Thanks for the fix, this rst is generated from here: https://github.com/WebAssembly/spec/blob/main/document/core/appendix/gen-index-instructions.py#L430-L433 so maybe update the python file and regenerate the rst, or make the same changes to the python file.

rossberg (approved):
Thanks!

PR #1590, opened 2023-02-08
[interpreter] Support nested invoke/get in test scripts

### LINKED ISSUE #1568: webassembly.github.io links all 404
Other than https://webassembly.github.io/spec/ most links appear to be broken. This is what MDN links to when needed. What is the right thing to use instead, or will this be fixed?

The only alternative I'm finding on google search is https://www.w3.org/TR/wasm-core-2/ - most links go to webassembly.github.io

This came up in https://github.com/mdn/content/issues/22908

rossberg:
Also adds `set` action to mutate globals. See script.wast for examples.

@titzer, PTAL. Should fix #1568.

This was straightforward to implement, except for the JS conversion, which caused quite some headache because wrapper modules generated for actions on types inexpressible in JS are no longer closed now but need to take their arguments as imports. This is largely untested.

### DIFF
--- a/test/core/script.wast
+++ b/test/core/script.wast
@@ -7,7 +7,7 @@
 )
 
 (module $m2
-  (global (export "g") i32 (i32.const 42))
+  (global (export "g") (mut i32) (i32.const 42))
   (func (export "f") (result i32) (i32.const 42))
   (func (export "add3") (param i32 i32 i32) (result i32)
     (i32.add (i32.add (local.get 0) (local.get 1)) (local.get 2))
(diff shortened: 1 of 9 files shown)


rossberg:
Okay, JS conversion is completely broken right now. To support multiple imports in Wasm wrappers, the implementation needs a complete refactor. I'll look into it, but might no find time for a little while.

test/core/script.wast
```diff
@@ -0,0 +1,100 @@
...
+(assert_return (get $m2 "g") (i32.const 45))
+
+(assert_return (invoke "f") (i32.const 42))
+(assert_return (invoke $m1 "f") (i32.const 41))
+(assert_return (invoke $m2 "f") (i32.const 42))
+
+(assert_return (invoke $m1 "inc" (i32.const 2)) (i32.const 3))
+(assert_return (invoke $m1 "inc" (get $m1 "g")) (i32.const 42))
+(assert_return (invoke $m1 "inc" (get $m2 "g")) (i32.const 46))
+(assert_return (invoke $m1 "inc" (invoke $m1 "inc" (get "g"))) (i32.const 47))
```
titzer, line 41: I'm wondering if intentionally limiting the grammar to not allow nested invokes (or even `set`) would be good. One the one hand, it simplifies the test runner so it doesn't have to keep an internal stack, and also forces tests to flatten out their nested calls, saving their intermediate results in globals.

rossberg:
I did a rewrite of most of the JS converter and it now can handle wrapper modules with an arbitrary number of imports. It also made the code more readable.

test/core/script.wast
```diff
@@ -0,0 +1,100 @@
...
+    (local.get 1) (local.get 0)
+  )
+)
+
+(assert_return (get "g") (i32.const 42))
+(assert_return (get $m1 "g") (i32.const 41))
+(assert_return (get $m2 "g") (i32.const 42))
+
+(set "g" (i32.const 43))
+(assert_return (set "g" (i32.const 43)))
```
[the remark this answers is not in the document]
rossberg: Right, it's like global.set.

test/core/script.wast
rossberg: Actually, I have often missed the ability to simply plug two function calls together for an assertion. Not being able to do that on script level requires contortions exactly like those from not being able to access a global. So it would be a pity to exclude it. It's just a trivial recursive expression evaluator after all.

The real complexity with this extension is in translating tests to JS. But that all has to do with the fact that synthesised Wasm modules for assertions now require an arbitrary number of imports, which isn't changed by limiting expression depth.

(Also, I ran into a couple of V8/node bugs, such as #1597, and worse, deterministic bus errors on my new Arm MacBook. :( )

rossberg:
@titzer, are you content with this? If so, can you approve the PR?

PR #1591, opened 2023-02-08
Merge extended-const proposal

sbc100:


sbc100:
I've rebased here.. is that the right thing to do?  I can merge back if you prefer?

rossberg:
@sbc100, my understanding is that merging a proposal into the main repo requires WG clearance, i.e. phase 5 -- which I hope we can have at the next WG meeting. A PR should then include the actual spec and the proposal docs.

If you just intended to sync the proposal with upstream, then I think that ought to be done in the proposal repo. But it would be good to sync up the spec doc as well in that case.

sbc100:
Got it, thanks.

PR #1597, opened 2023-02-10
[test] Import v128 global

rossberg:
Adds a simple test for ex/importing a global of type v128.

@ajklein, @jakobkummerow, this makes V8's InstanceBuilder crash, presumably because v128 isn't handled [here](https://github.com/v8/v8/blob/main/src/wasm/module-instantiate.cc#L1446) (tested with node 18.14.0, the CI failure is caused by that as well).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 18.x
-      - run: cd interpreter && opam exec make all
+      - run: cd interpreter && opam exec make JS=node all
 
   ref-interpreter-js-library:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


ajklein:
Thanks, filed https://crbug.com/v8/13732 to track the V8 crash.

jakobkummerow (approved):

rossberg:
Thanks. @jakobkummerow, do you happen to know the rough turn-around time for a V8 fix to make it downstream into node? I'm hesitant to land this PR right now, since I'd have to disable running JS tests in CI to make it green.

rossberg:
> As a short-term workaround, you could make the global mutable (with a `TODO` comment to add a test for an immutable global as well, once the V8 fix has propagated far enough).

Done.

tomstuart:
It’s surprising that this appears in a `test/core/*.wast` test script instead of `test/core/simd/*.wast`. All other occurrences of `v128` appear in `test/core/simd/*.wast`.

rossberg:
Good point, will fix.

PR #1602, opened 2023-02-15
[spec] Clarify in conventions that atoms can be symbolic

rossberg:
Addresses discussion point [here](https://github.com/WebAssembly/spec/issues/726#issuecomment-1431227194).


bonjune:
I think it is a good clarification.

PR #1606, opened 2023-02-18
[spec] Normative: Permit work in parallel during instantiation

tomstuart:
This is a recreation of @littledan’s #745, which was erroneously merged into `master` instead of `main` because it’s five years old.


Ms2ger (approved):
Doh, thanks for pointing that out.

PR #1607, opened 2023-02-18
[spec] Add missing initialization value to table.grow reduction rule

tomstuart:
This rule was probably copied from the (very similar) memory.grow rule when it was introduced by #1287 in c3d5cbc1f7c6b166b7a312adc4efdb7e34746b3d. While memory.grow doesn’t have an initialization value, table.grow does.


rossberg:
Strike that, I need coffee. :)

rossberg (approved):

rossberg:
Thanks!

tomstuart:
Right, we still need to remove it from the stack. Sorry, should’ve been clearer!

PR #1614, opened 2023-02-22
[spec] Clarifying note about canonical NaNs

### LINKED ISSUE #1613: Confusion about the word "canonical" for NaNs
The Wasm spec uses the word ["canonical"](https://webassembly.github.io/spec/core/syntax/values.html#canonical-nan) to describe NaNs in a particular form. I recently spoke with someone who was confused because IEEE 754 also has a concept of a "canonical" representation for a NaN, and it's not the same as the Wasm spec's.

Please consider adding a note to the spec mentioning that IEEE 754's "canonical" and Wasm's "canonical" are different things.

rossberg:
Fix #1613.


sunfishcode (approved):
Thanks!

PR #1616, opened 2023-03-01
[js-api] Remove some anchor definitions

Ms2ger:
WebRef has recently added definitions for ECMAScript, which means most of the explicit anchor definitions are no longer needed. Also, some abstract operations had their names changed.

This change should be entirely editorial.


rossberg (approved):
Rubber-stamping

PR #1624, opened 2023-03-16
[spec] Fix execution of control instructions, regarding current frame

f52985:
This pull request will
1. Fix some typos regarding periods.
2. Append the step `Let F be the current frame <syntax-frame>.`, when needed for `expand_F`. (block, loop-block)
3. Remove use of `F` in formal notation if unnecessary. (if-block)


rossberg (approved):
Thanks!

PR #1627, opened 2023-03-22
[ci] Split CI jobs

rossberg:
This splits up the Github CI workflows for interpreter and spec, and sets up path filters to avoid unnecessary work.

Also adds a bit more friendly output.


rossberg:
Turns out I screwed up the path filters (missing `/**`). Should be fixed now.

PR #1630, opened 2023-03-23
Make "large" compatible with wasm2kotlin

### LINKED ISSUE #1619: Is "large" test required for compliance?
Is the br_table "large" test required? Some platforms (e.g. JVM bytecode) simply do not support jump tables that large.

SoniEx2:
Closes #1619


rossberg (approved):
Thanks!

rossberg:
@sunfishcode, sorry, I merged before seeing your comment (some github hickup).

But I think this is fine, because (a) the test still is pretty large, and (b) it is not a negative test -- nothing prevents larger examples from passing.

We generally expect different embedding environments to impose different limits. There is something to be said about broader standards for minimum limits, but the core test suite is not the right place to enforce this. I don't think it has a mandate for that, as long as such limits are not actually stated in the core spec (which they intentionally aren't). Rather, the test suite should cater to a diverse range of embedders.

SoniEx2:
@sunfishcode for context, even something like blastem, whose opcodes are 16-bit, doesn't use anywhere near that many table entries.

PR #1636, opened 2023-03-29
[spec] Fix typo in module instantiation of execution

jaehyun1ee:
A closing parenthesis seems to be missing for the condition of module instantiation.

image

So this pull request simply adds a closing parenthesis for it.


rossberg (approved):
Thanks!

PR #1639, opened 2023-04-08
[spec] Use the correct link to in element sections

TethysSvensson:
The link for `The expression ei must be constant.` points to the text for `t.const` instructions, not to the text for constant expressions.


rossberg (approved):
Thanks.

PR #1648, opened 2023-05-04
[test] Remove malformed tests that become valid with memory64

backes:
With memory64, memory offsets will be decoded as u64. An engine implementing this will fail tests that test that we reject such module. Thus remove those tests from the main spec repo now; they should be added back in the memory64 repo, but changed to `assert_trap` tests.

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -504,25 +504,6 @@
     "\7f"                      ;; i32
     "\41\00"                   ;; i32.const 0
     "\28"                      ;; i32.load
-    "\02"                      ;; alignment 2
-    "\82\80\80\80\80\00"       ;; offset 2 with one byte too many
-    "\1a"                      ;; drop
-    "\0b"                      ;; end
-  )
-  "integer representation too long"
-)
-(assert_malformed
-  (module binary
-    "\00asm" "\01\00\00\00"
-    "\01\04\01\60\00\00"       ;; Type section
-    "\03\02\01\00"             ;; Function section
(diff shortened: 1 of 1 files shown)


rossberg (commented):
Instead of removing the tests, can you comment them out with a TODO to reactivate them with assert_trap in the Memory64 proposal?

Also, if you could then downstream this to the mem64 repo and address the TODO there that would be awesome. ;)

backes:
Thanks, Andreas, that makes sense. First part done, will address the second part now (in the memory64 repo).

rossberg (approved):
Thanks!

backes:
Sorry, I just noticed that something is off here when preparing the change in the memory64 repo. The tests were moved from `binary.wast` to `binary-leb128.wast` in #1019, but then came back when merging the bulk instructions & reference types proposals (#1287).

That would mean that we should indeed remove the tests from `binary.wast`, and maybe just already merge those few changes to `binary-leb128.wast` to make it fail both before and after memory64.

Not sure how to check if there were more such merge errors in #1287.

sbc100:
The memory64 repo is currently quite out-of-date with upstream I'm afraid.  I've been struggling to find time to resolve the merge conflicts (my OCaml skills is not too sharp yet) to get it up-to-date.

backes:
@sbc100 I guess that's fine for now, and is not what's causing the problems.

I uploaded #1649 to remove the duplicated tests (merge error in #1287) and #1650 to make the tests memory64-ready, which I think is better than commenting them out.

PR #1652, opened 2023-05-11
Update README.md

alcantara-m:
It was necessary to install the ocaml-base-compiler to build the interpreter

### DIFF
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -15,7 +15,6 @@ The text format defines modules in S-expression syntax. Moreover, it is generali
 
 ## Building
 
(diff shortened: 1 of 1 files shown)


rossberg:
Thanks, but I'm still a bit confused. Isn't ocaml-base-compiler always installed, at least when following the linked instructions? Can an OCaml installation even work without it? What method did you use that failed?

zapashcanon:
The only thing I can think of is installing opam without ocaml on e.g. Debian with the following `apt install --no-install-recommends opam`.

rossberg:
I'm taking the liberty to close this PR.

PR #1654, opened 2023-05-22
[spec] Fix typo in introduction of element section's bitfield

am009:
The bit 0 should indicates a passive or active segment.


rossberg:
I don't think this fix is correct. When bit 0 is set, it is either passive or declarative, like the note is saying. Active segments require a cleared bit 0.

am009:
You are correct.

But now I think it is a little bit misleading. Bit 0 **being set to 1** indicates a passive or declarative segment. What about changing it to "Bit 0 indicates a passive or declarative segment instead of an active segment".

rossberg:
Note that the current wording is mirrors the one for data segments, so if we change it, we should probably change both.

PR #1662, opened 2023-06-06
Fix copypaste error for V128.Load*_Zero instructions in index

rdunnington:
The website instruction index contains a error for these two instructions:
* V128.Load32_Zero
* V128.Load64_Zero

It specifies both memarg and laneidx immediates, but in reality it isn't supposed to have a laneidx since the instructions assume they set the lowest bits of the vector. It is correct in the Structure/Instructions/Memory Instructions part of the spec.


rossberg (approved):
Thanks!

PR #1665, opened 2023-06-14
use dune in the Makefile and remove ocamlbuild

zapashcanon:
Hi,

As discussed in #1664, I switched everything from `ocamlbuild` to `dune`, I removed winmake.bat along the way. I made a lot of simplifications in the Makefile but also removed a few things: we don't build the bytecode version anymore and we don't copy the executable at the root anymore.


interpreter/README.md
ngzhian, line 1: I think we should add instructions here on how to get `dune` https://github.com/WebAssembly/spec/tree/main/interpreter#building

interpreter/Makefile
```diff
@@ -9,131 +9,39 @@
...
-DIRS =		util syntax binary text valid runtime exec script host main tests
-LIBS =		
-FLAGS = 	-lexflags -ml -cflags '-w +a-4-27-42-44-45-70 -warn-error +a-3'
-OCBA =		ocamlbuild $(FLAGS) $(DIRS:%=-I %)
-OCB =		$(OCBA) $(LIBS:%=-libs %)
-JSO =		js_of_ocaml -q --opt 3
-JS =		# set to JS shell command to run JS tests, empty to skip
+OPT =		wasm.exe
+ZIP =		wasm.zip
+JSLIB =		wast.bc.js
```
rossberg, line 14: The end result here should still be named wast.js. May require renaming it in the build target.

interpreter/README.md
```diff
@@ -22,10 +22,7 @@ Once you have OCaml, simply do
 ```
 make
 ```
-You'll get an executable named `./wasm`. This is a byte code executable. If you want a (faster) native code executable, do
-```
-make opt
-```
+You'll get an executable named `_build/default/wasm.exe`.
```
rossberg, line 25: Let's not break people's workflow. Can you tweak the makefile such that this is symlinked to ./wasm?

interpreter/Makefile
```diff
@@ -9,131 +9,39 @@
 
 # Configuration
 
-NAME =		wasm
```
rossberg, line 12: Can we keep the NAME variable? That makes it easier to reuse the Makefile for derivatives.

interpreter/Makefile
```diff
@@ -182,27 +83,8 @@ dunetest:
...
-		git archive --format=zip --prefix=$(NAME)/ -o $@ HEAD
+$(ZIP):
+	git archive --format=zip --prefix=wasm/ -o $@ HEAD
 
 clean:
-		rm -rf _build/jslib $(LIB).mlpack _tags $(JSLIB).js
-		$(OCB) -clean
-
-
-# Opam support
```
rossberg, line 193: Why did you remove the opam support?

interpreter/README.md
rossberg: Agreed.

interpreter/Makefile
```diff
@@ -9,131 +9,39 @@
...
-		$(OCB) -quiet $@
-
-.PHONY:		smallint.byte smallint.native
-smallint.byte: _tags
-		$(OCB) -quiet $@
-smallint.native: _tags
-		$(OCB) -quiet $@
-
-
-# Building library
```
rossberg, line 82: This is used to create the Wasm opam package. Perhaps it can be simplified, but we still need it.

rossberg (commented):
Changes look good mostly, but you may have been a bit over-aggressive in deleting stuff. At least I wouldn't know how to build the opam library package anymore after this. ;)

Is there a way to avoid and work around the renaming main.ml→wasm.ml? IMO it's bad design that dune assumes this.

interpreter/Makefile
zapashcanon: It's broken currently. I'll need to make a few fixes to get it back (e.g. move the `dune-project` at the root of the repository again, see [ocaml/dune#7962](https://github.com/ocaml/dune/issues/7962). Currently it crashes, when it'll be fixed, we'll still get an error saying that dependencies outside of workspace is not supported. You can reproduce easily by doing `dune runtest` from the `interpreter` directory on current main.

interpreter/README.md
zapashcanon: I'm fine with having `./wasm` but IIRC, currently it's `./wasm.exe` and we get an error from dune saying that we're asking to generate `./wasm.exe` which is already in the source tree.

interpreter/Makefile
zapashcanon: If you're OK with it, I can bring back proper opam support in another PR, but it requires more fixes (e.g. having a `public_name` and not simply a `name` etc.).

interpreter/Makefile
zapashcanon: Fixed.

interpreter/Makefile
zapashcanon: I made a minimal opam support.

zapashcanon:
> Is there a way to avoid and work around the renaming main.ml→wasm.ml? IMO it's bad design that dune assumes this.

I did something for this. Everything should be OK now @rossberg. :)

interpreter/Makefile
```diff
@@ -142,67 +55,45 @@ TESTDIR =	../test/core
...
+	dune install
 
 # Miscellaneous targets
 
 .PHONY:		clean
 
-$(ZIP):		$(WINMAKE)
-		git archive --format=zip --prefix=$(NAME)/ -o $@ HEAD
+$(ZIP):
+	git archive --format=zip --prefix=wasm/ -o $@ HEAD
```
rossberg, line 95:
```suggestion
	git archive --format=zip --prefix=$(NAME)/ -o $@ HEAD
```

interpreter/Makefile
```diff
@@ -142,67 +55,45 @@ TESTDIR =	../test/core
...
-
-install:	_build/$(LIB).cmx _build/$(LIB).cmo
-		ocamlfind install $(LIB) meta/findlib/META _build/$(LIB).o \
-		  $(wildcard _build/$(LIB).cm*) \
-		  $(wildcard $(DIRS:%=%/*.mli))
-
-uninstall:
-		ocamlfind remove $(LIB)
+	dune clean
+	rm -f $(NAME) $(JSLIB)
```
rossberg, line 99: This should probably be in a separate distclean target.

interpreter/Makefile
```diff
@@ -9,131 +9,44 @@
...
-		echo if \'%1\' neq \'\' set NAME=%1 >>$@
-		$(OCB) main.byte \
-		| grep -v ocamldep \
-		| grep -v mkdir \
-		| sed s:`which ocaml`:ocaml:g \
-		| sed s:main/main.d.byte:%NAME%.exe: \
-		>>$@
-
+$(JSLIB):
+	rm -f $@
```
rossberg, line 47: That is odd, why is this necessary? You probably want to make the copy via a target, so that dependency tracking works properly.

interpreter/wasm.opam
```diff
@@ -0,0 +1,22 @@
+# This file is generated by dune, edit dune-project instead
```
rossberg, line 1: If this file is generated, is there a reason to check it in?

interpreter/.gitignore
```diff
@@ -1,13 +1,3 @@
-*.cmo
-*.cmx
-*.native
-*.byte
-*.opt
-*.unopt
-*.js
-*.zip
```
rossberg, line 8: You probably want to keep *.js and *.zip

rossberg (commented):
Cool, thanks!

interpreter/Makefile
zapashcanon: Fixed.

interpreter/Makefile
zapashcanon: Fixed.

interpreter/wasm.opam
zapashcanon: Yes, that's standard. The main reason is that, you usually want to be able to run things like: `opam pin git@github.com:WebAssembly/spec.git`. But opam doesn't know that the package is to be built with `dune` and can't call `dune build wasm.opam`, thus we have to commit the file. There are plans to fix this but I don't know if it'll happen soon.

interpreter/.gitignore
zapashcanon: Fixed.

interpreter/Makefile
```diff
@@ -92,8 +92,10 @@ install:
 .PHONY:		clean
 
 $(ZIP):
-	git archive --format=zip --prefix=wasm/ -o $@ HEAD
+	git archive --format=zip --prefix=$(NAME)/ -o $@ HEAD
 
 clean:
 	dune clean
+
+distclean:
```
rossberg, line 100:
```suggestion
distclean: clean
```

rossberg (approved):

interpreter/Makefile
zapashcanon: Sorry, fixed.

interpreter/wasm.opam
avsm: It's not a bug, so no plans to fix it :-)

Checking in the `opam` file ensures that the package manager can just inspect that file from the release archive to know what to do with it.  If code needs to be run to generate the file, and that fails or otherwise has complex constraints (e.g. a different dune), then figuring out an installation plan becomes quite complex.

So this looks fine the way it is...

interpreter/Makefile
rossberg: This should work:
```
$(JSLIB):	$(BUILDDIR)/$(JSLIB)
	cp $< $@

$(BUILDDIR)/$(JSLIB):
	dune build $(JSLIB)
```

interpreter/Makefile
```diff
@@ -142,67 +55,47 @@ TESTDIR =	../test/core
...
-# Opam support
-
-.PHONY:		check install uninstall
-
-check:
-		# Check that we can find all relevant libraries
-		# when using ocamlfind
-		ocamlfind query $(LIBS)
-
-install:	_build/$(LIB).cmx _build/$(LIB).cmo
```
rossberg, line 202: This target (as well as opt and unopt) is currently used by the `opam` package file. I'm not sure what that needs to be replaced with.

As part of a doing a new 2.0.1 package this morning, I checked in a template for the opam file under meta/opam. Can you merge main and adapt that file to the changes here?

interpreter/wasm.opam
rossberg: I see. Somehow you folks must have a higher itch threshold against cluttered directories than I... ;)

rossberg (commented):
Sorry, I had somehow lost this PR.

interpreter/wasm.opam
zapashcanon: > we might even skip the generation step altogether by having opam invoke dune to get the generated opam file directly

From [discuss]. But I guess the plans changed one more time. ;-)

interpreter/Makefile
zapashcanon: They won't be needed anymore once we use dune.

I updated the `dune-project` to add the new metadata from `meta/opam`.

zapashcanon:
It should be good now @rossberg ?

interpreter/Makefile
rossberg: But what does that have to do with the Makefile? Wouldn't you get that error either way?

zapashcanon: The Makefile is the one putting `wast.js` at the root of the repository. It works with your fix as long as one doesn't try to run `make wast.js` and then `dune build wast.js`.

I'm not going to use the Makefile anyway so I don't really care. And I believe people using the Makefile are not going to run dune manually so this is probably fine...

interpreter/Makefile
```diff
@@ -142,78 +55,47 @@ TESTDIR =	../test/core
...
-opam-release/%:
-		git tag opam-$*
-		git push --tags
-		rm -f opam-$*.zip
-		wget https://github.com/WebAssembly/spec/archive/opam-$*.zip
-		cp meta/opam/opam .
-		sed -i ".tmp" s/@VERSION/$*/g opam
-		sed -i ".tmp" s/@MD5/`md5 -q opam-$*.zip`/g opam
-		rm opam.tmp opam-$*.zip
-		@echo Created file ./opam, submit to github opam-repository/packages/wasm/wasm.$*/opam
```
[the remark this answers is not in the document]
rossberg: Thanks for volunteering, I am happy to pass that on to you. :)

Yet, others should still be able to do that easily in case, to avoid single points of failure. A makefile target is a useful abstraction that is both automating and documenting the steps necessary, and tends to be a more stable interface than the tools behind it. (Once a person has worked on a dozen diverse systems that each have gone through multiple generations of custom build tools, they start to appreciate having that abstraction. ;) )

If opam-publish or dune-release are the best way of implementing that process now (I admit I haven't used those before), can you adjust the Makefile target to call into those?

interpreter/Makefile
```diff
@@ -88,6 +88,13 @@ install:
 	dune build -p $(NAME) @install
 	dune install
 
+.PHONY: opam-release
```
rossberg, line 91: I don't think this works, since the target contains a wildcard.

rossberg (approved):
Thanks!

interpreter/Makefile
zapashcanon: Oh yes sorry, it's because at first I had a `opam-release:` target without a Makefile.

interpreter/Makefile
```diff
@@ -88,6 +88,13 @@ install:
...
+
+opam-release/%:
+	git tag opam-$*
+	git push --tags
+	opam publish
```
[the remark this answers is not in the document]
zapashcanon: IIRC it's going to ask.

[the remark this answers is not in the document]
rossberg: I just tried opam-publish with all variations and available options, but it always said "Not enough information to publish", even when I set the version explicitly. My suspicion is that it bakes in assumptions like package-name = repo-name.

The implementation details don't really matter to me, but this target should continue working.

rossberg (changes requested):

interpreter/Makefile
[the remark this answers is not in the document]
zapashcanon: Indeed, it breaks also `dune-release`. This is because we don't have `wasm.opam` at the root of the repository...
Should I move `dune-project` and `wasm.opam` at the root (I know you don't like it) ? Otherwise we can keep the previous version.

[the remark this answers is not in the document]
zapashcanon: Also note that having them at the root of the repository will make `opam pin https://github.com/WebAssembly/spec` to work properly.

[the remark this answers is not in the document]
rossberg: By root you mean the root of the repo? I'm afraid that wouldn't be adequate, since the interpreter is only a secondary subcomponent of this project that shouldn't affect the root dir. It would also slightly break path-dependent CI.

Would perhaps opams-publish's --packages-directory=interpreter option work in this case, by any chance?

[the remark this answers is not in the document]
rossberg: That's unfortunate, but thanks for poking. Let's revert to what was there before then. Maybe opam will support it at some point in the future.

(CI is set up per toplevel directory, i.e., interpreter builds only run when something under interpreter/ changes. Moving dune and opam files out of there would then fail to capture them. Although that could probably be hacked.)

interpreter/Makefile
```diff
@@ -91,7 +91,17 @@ install:
...
+	rm -f opam-$*.zip
+	wget https://github.com/WebAssembly/spec/archive/opam-$*.zip
+	cp wasm.opam opam
+	echo 'url {' >> opam
+	echo '  src: "https://github.com/WebAssembly/spec/archive/opam-@VERSION.zip"' >> opam
+	echo '  checksum: "md5=@MD5"' >> opam
+	echo '}' >> opam
+	sed -i ".tmp" s/@VERSION/$*/g opam
+	sed -i ".tmp" s/@MD5/`md5 -q opam-$*.zip`/g opam
+	rm opam.tmp opam-$*.zip
```
rossberg, line 103: If we are producing this via echo, we don't need sed but can splice in the right strings right here:
```suggestion
	echo "  src: \"https://github.com/WebAssembly/spec/archive/opam-$*.zip\"" >> opam
	echo "  checksum: \"md5=`md5 -q opam-$*.zip`\"" >> opam
	echo "}" >> opam
        rm opam-$*.zip
```

rossberg (approved):

interpreter/Makefile
zapashcanon: Indeed. Also the `sed -i ".tmp" ...` part is not working on linux, this is a macOS thing I guess. To be portable you'd do something like `sed -i.tmp ...`.

interpreter/Makefile
```diff
@@ -95,12 +95,10 @@ opam-release/%:
 	wget https://github.com/WebAssembly/spec/archive/opam-$*.zip
 	cp wasm.opam opam
 	echo 'url {' >> opam
```
rossberg, line 97: Nit: maybe also use `"` here for consistency.

rossberg (approved):

interpreter/Makefile
zapashcanon: done :)

rossberg:
I merged it, thanks for taking this.

PR #1668, opened 2023-06-30
Add type to elem.drop and element segment store soundness

### LINKED ISSUE #1663: [spec] `type` field of element instance seems unnecessary
Currently, the element instance consists of two fields, `type` and `elem`:

스크린샷 2023-06-13 오후 4 42 50

However, it seems that the `type` field of the element instance is unnecessary.
As far as i know, none of the reduction rules for the execution actually access the `type` field of a element instance (only `elem` field is being accessed).
Also, the `elem.drop` instruction is treating element instance as if it has only one field:

스크린샷 2023-06-13 오후 4 51 56

Does this `type` field of element instance have a specific purpose? If not, should it be removed from the spec?

bvisness:
The execution section for `elem.drop` was missing the `type` field of element instances. In addition, the heading of the Element Instances section in the Soundness appendix was also missing the type. I have added the type to both places.

This is my first contribution to the spec repo, so apologies if I got any conventions wrong.

Would close #1663 if it wasn't already closed for some reason.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1676,15 +1676,9 @@ Table Instructions
 
 1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
 
-2. Assert: due to :ref:`validation <valid-elem.drop>`, :math:`F.\AMODULE.\MIELEMS[x]` exists.
+2. Assert: due to :ref:`validation <valid-elem.drop>`, :math:`S.\SELEMS[F.\AMODULE.\MIELEMS[x]]` exists.
 
-3. Let :math:`a` be the :ref:`element address <syntax-elemaddr>` :math:`F.\AMODULE.\MIELEMS[x]`.
-
-4. Assert: due to :ref:`validation <valid-elem.drop>`, :math:`S.\SELEMS[a]` exists.
-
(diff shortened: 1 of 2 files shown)


document/core/exec/instructions.rst
```diff
@@ -1691,7 +1693,11 @@ Table Instructions
    S; F; (\ELEMDROP~x) &\stepto& S'; F; \epsilon
    \end{array}
    \\ \qquad
-     (\iff S' = S \with \SELEMS[F.\AMODULE.\MIELEMS[x]] = \{ \EIELEM~\epsilon \}) \\
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & F.\AMODULE.\MIELEMS[x] = a \\
+     \wedge & \X{t} = S.\SELEMS[a].\EITYPE \\
+     \wedge & S' = S \with \SELEMS[a] = \{ \EITYPE~t, \EIELEM~\epsilon \}) \\
+     \end{array}
```
rossberg, line 1700: I think we can actually simplify this to:
```suggestion
     (\iff S' = S \with \SELEMS[F.\AMODULE.\MIELEMS[x]].\EIELEM = \epsilon) \\
```
(and similarly shorten the prose)

rossberg (commented):
Ah yes, thanks!

document/core/exec/instructions.rst
bvisness: Done, let me know what you think.

rossberg (approved):
Thanks!

PR #1678, opened 2023-09-08
Fix typos

yamt:
I can't think of any reasons to repeat identical functions with different names here. I guess they are typos.


rossberg (approved):
Thanks!

PR #926, opened 2018-11-29
[spec/interpreter/tests] Rename instructions

rossberg:
Implement instruction renamings from [this list](https://github.com/WebAssembly/spec/issues/884#issuecomment-426433329) on #884, as decided at [2018/10/02 CG meeting](https://github.com/WebAssembly/meetings/blob/e43a829840fed8fe17c1061decdc3260d6323e04/2018/CG-10-02.md).

PR #1680, opened 2023-09-20
[test] text format: drop support for some pre-standard keywords

keithw:
In 2018 (https://github.com/WebAssembly/spec/pull/720 and https://github.com/WebAssembly/spec/pull/926), the spec renamed some keywords before WebAssembly was standardized (`get_local` -> `local.get`, `anyfunc` -> `funcref`, etc.).

WABT dropped support for the old names in 2021 (https://github.com/WebAssembly/wabt/pull/1792), MDN revised its documentation in 2022 (https://github.com/mdn/interactive-examples/pull/2298), and wasm-tools dropped support last month (https://github.com/bytecodealliance/wasm-tools/pull/1184). However, users do get confused when they find a tutorial that uses the old syntax and the code works in more-tolerant implementations but is rejected by other conforming consumers. (E.g. https://github.com/WebAssembly/wabt/issues/1802, https://github.com/WebAssembly/wabt/issues/1820, https://github.com/WebAssembly/wabt/issues/1883, https://github.com/WebAssembly/wabt/issues/2010, ...)

Now that it's been five years and tutorials using the old syntax are increasingly rare, this PR would check that the old keywords are now considered malformed.


rossberg (approved):
LGTM, except nit: can we rename the test to `obsolete-keywords`?

keithw:
> LGTM, except nit: can we rename the test to `obsolete-keywords`?

Sure, pushed.

PR #1684, opened 2023-09-25
[spec/interpreter/test] Align definition of newline with Unicode reco…

rossberg:
So far, the Wasm text format only recognises LF as a newline. This PR tweaks that definition to align with Section 5.8 of the [Unicode standard](https://www.unicode.org/versions/latest/), which recommends treating CR, LF, CRLF, and NEL as equivalent newlines in text files, in order to be agnostic to newline translations performed by various tools and conversions between OSes and character sets.

This only affects the interpretation of line comments, where previously, code like
```
;; comment<CR>
(instr)
```
would (surprisingly) treat the instruction as part of the comment, but no longer does now. While theoretically a breaking change, it shouldn't affect any well-formatted code. It hence seems okay to consider this a bug fix.


sunfishcode:
I propose not recognizing NEL (U+85) as a newline character. It is not recognized as a newline in popular text formats, tools, or editors, so it would be inconvenient to work with if files using NEL newlines were in circulation. An incomplete list of popular text formats that don't recognize NEL:
 - [CommonMark](https://spec.commonmark.org/0.30/#characters-and-lines)
 - [XML](https://www.w3.org/TR/xml/#sec-line-ends)
 - [JSON](https://datatracker.ietf.org/doc/html/rfc7159#section-2)
 - [TOML](https://toml.io/en/v1.0.0#spec)
 - [YAML](https://yaml.org/spec/1.2.2/#54-line-break-characters)

rossberg:
Yeah, I'm not feeling strongly. I'd assume there is value in following the official Unicode recommendation, even though NEL hardly matters in practice. I believe the motivation for that recommendation is maximising compatibility with legacy character sets, but I'd be fine limiting that concern to ASCII.

That said, there are various other codepoints in Unicode that represent forms of line breaks, and there likely is no consistency in how editors and tools treat those. So it's doubtful if that form of alignment is even possible, independent of NEL.

tlively:
Do NEL characters ever appear in practice? If not, following unicode would not cause any additional problems in practice and we would get the benefits of just doing what unicode recommends (at least for this particular sliver of the design space).

sunfishcode:
> Do NEL characters ever appear in practice? If not, following unicode would not cause any additional problems in practice and we would get the benefits of just doing what unicode recommends (at least for this particular sliver of the design space).

I'm not aware of any practical benefits of following this particular Unicode recommendation here. NEL's only known use in practice is in text converted from EBCDIC. My understanding is that Unicode added the recommendation about NEL for the benefit of interoperability with EBCDIC documents. I don't believe this is relevant to Wasm.

Additional examples of things that don't interpret U+85 (NEL) as a newline include [JS](https://262.ecma-international.org/14.0/#sec-line-terminators), [CSS](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#what_is_whitespace), [HTML](https://www.w3.org/TR/css-text-3/#white-space), [Python](https://docs.python.org/3/library/functions.html#open), and many more.

tlively:
I do think that being able to offload all decision making in this area to unicode is an advantage, but I'll admit it's a small one. I guess being consistent with so many existing text formats is probably a larger advantage, so not treating NEL as a newline sounds fine to me.

rossberg:
Okay, I don't care enough, so I just removed NEL. PTAL.

document/core/text/lexical.rst
```diff
@@ -85,7 +85,9 @@ The allowed formatting characters correspond to a subset of the |ASCII|_ *format
    \production{white space} & \Tspace &::=&
      (\text{~~} ~|~ \Tformat ~|~ \Tcomment)^\ast \\
    \production{format} & \Tformat &::=&
-     \unicode{09} ~|~ \unicode{0A} ~|~ \unicode{0D} \\
+     \Tnewline ~|~ \unicode{09} \\
+   \production{newline} & \Tnewline &::=&
+     \unicode{0A} ~|~ \unicode{0D} ~|~ \unicode{0D}~\unicode{0A} \\
```
tlively, line 90: If we got rid of the last case here, then it would parse as two newlines in a row, which I don't think would break anything. Is it worth making that simplification, or is it better to keep that case for consistency with the standard definition of newline?

test/core/comments.wast
tlively, line 1: Interesting and unfortunate that git and GitHub treat this as a binary file.

tlively (approved):

document/core/text/lexical.rst
rossberg: Indeed, currently that should not observably change anything (though an implementation taking that literally would report incorrect line numbers). But yeah, I think it is preferable to use the standard definition for clarity.

test/core/comments.wast
rossberg: I assume their heuristic is based on the appearance of 0x00 and other control characters.

PR #1686, opened 2023-10-06
Don't install ocamlbuild in CI

foolip:
It ought to not be needed after https://github.com/WebAssembly/spec/pull/1665.


rossberg (approved):
Thanks!

PR #1696, opened 2023-10-24
[spec] Fix definition of iextend

rossberg:
Fix #1695.

### DIFF
--- a/document/core/exec/numerics.rst
+++ b/document/core/exec/numerics.rst
@@ -719,7 +719,7 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
 
 .. math::
    \begin{array}{lll@{\qquad}l}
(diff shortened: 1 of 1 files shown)


document/core/exec/numerics.rst
```diff
@@ -719,7 +719,7 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
 
 .. math::
    \begin{array}{lll@{\qquad}l}
-   \iextendMs_{N}(i) &=& \extends_{M,N}(i) \\
+   \iextendMs_{N}(i) &=& \extends_{M,N}(\wrap_{N,M}(i)) \\
```
tlively, line 722: Looks like the prose above needs the same fix.

tlively (approved):

document/core/exec/numerics.rst
rossberg: Oh, indeed, fixed.

PR #1697, opened 2023-10-25
[spec] Cleaner definition of frames

### LINKED ISSUE #1693: Activation frames: where's the return arity?
According to the section on Activation Frames:

image

"Activation frames carry the return arity _n_ of the respective function..."

However, the definition of the frame given there doesn't mention _n_ at all.

rossberg:
Fix #1693 by separating the syntactic categories of frames and of their state.


conrad-watt (approved):
lgtm

PR #1699, opened 2023-10-27
[test] introduce script scopes around expensive memory tests

### LINKED ISSUE #1698: [JS Generator] Add scope for each wasm instance
The live range of the wasm instance created by JS Generator is global, likes:

> let $1 = instance(...);
> assert...($1, ...);
> let $2 = instance(...);
> assert...($2, ...);

Tests such as memory_copy.wast and memory_init.wast, usually contain many instances. When executing these tests on some machines, out of memory fails may be reported.

Can you add a scope to each instance so that they are destroyed after use ?

rossberg:
Fix #1698. This tweaks the generators for bulk memory/table tests to insert `(script ...)` scopes around tests and extends the JS converter to turn these into block scopes.

This is likely to break existing .wast consumers that do not recognise `script` scopes. The simplest fix is to ignore them by treating them as equivalent to their contents.


titzer:
Does it help to split up these tests into multiple `.wast`s? In general, splitting up the many large spec tests helps ergonomics of the test suite.

rossberg:
Personally, I'm not a big fan of having a gazillion micro test files. But in any case, these are generated, so that would require a more significant rewrite of the JS generator scripts, that somebody would have to volunteer doing.

ivomusk (approved):

ivomusk (approved):

ivomusk (approved):

PR #1700, opened 2023-10-27
[spec] Fix definition of lane ops + better xrefs

rossberg:
I noticed that the definition for mapping numeric ops over lanes was incorrect. This should fix it, and also improves some of the cross-referencing for numeric ops.

@ngzhian, can you take a look?


ngzhian (approved):
Looks good, thanks for the fixes!

PR #1705, opened 2023-11-10
switch to menhir

zapashcanon:
Hi @rossberg,

As discussed in [WebAssembly/gc#474](https://github.com/WebAssembly/gc/pull/474), I switched to Menhir.

In a second commit I demonstrate some of the facilities offered by Menhir. There are plenty others we could use to remove some code. But I already did this in Owi and I'm not feeling like rewriting a whole parser for the second time right now. :-)


rossberg (commented):
Thanks, that's nicer. I'm not entirely convinced the functorisation of the parse function is worth it, but I'm fine merging this for now.

rossberg (approved):

PR #1709, opened 2023-11-21
[spec] Add missing length condition to br_table prose

### LINKED ISSUE #1708: The definition of validation of br_table is inadequate
The spec defines the validation of br_table as follows and  I think this definition is inadequate. 

validate_br_table

When I first saw this definition, I took it to mean "all labels have the same type". However, I realised my mistake when I saw the following test in unreached-valid.wast.

https://github.com/WebAssembly/spec/blob/main/test/core/unreached-valid.wast#L49

unreached-valid

In my first interpretation, this module is invalid because [f32] and [f64] are not the same. 
In order for this module to be valid, t* must be something like [f32 f64].

The problem is that it is unclear what is intended by "corresponding" in the spec.

rossberg:
Fix #1708.

### DIFF
--- a/document/core/valid/instructions.rst
+++ b/document/core/valid/instructions.rst
@@ -1412,17 +1412,22 @@ Control Instructions
 ...........................
 
 
-* The label :math:`C.\CLABELS[l_N]` must be defined in the context.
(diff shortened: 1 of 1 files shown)


conrad-watt (approved):
lgtm

PR #1710, opened 2023-11-24
[spec] Fix typo in shape.all_true and txN.bitmask instruction execution

HoseongLee:
There seems to be some typos in shape.all_true instruction and txN.bitmask execution.
The parameter is labeled as $c_1$, but the some of the ones used are just labeled $c$.

image

This pull request changes all the incorrect $c$ to $c_1$.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -685,7 +685,7 @@ Most other vector instructions are defined in terms of numeric operators that ar
    \end{array}
    \\ \qquad
      \begin{array}[t]{@{}r@{~}l@{}}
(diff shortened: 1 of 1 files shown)


rossberg (commented):
Thanks! Instead of changing these occurrences to c_1 , can you change the other to plain c? The index is redundant in the absence of other c's.

HoseongLee:
I've changed all the c_1s to plain c for the above two instructions. However, the return value for txN.bitmask is also c, so I've changed it from c to i since it returns an i32.const.

rossberg (approved):
Thanks!

PR #1712, opened 2023-11-27
[spec] Fix module context for constants

### LINKED ISSUE #1711: Question about validation of {element, data} segments
The [spec](https://webassembly.github.io/spec/core/valid/modules.html#valid-module) tells that we have to validate element and data segments under context C' defined as follows.

context_C

However, that means the active element and data segments are always invalid, because C'.tables and C'.mems are always empty. Is this a specification error? Or am I missing something?

validate_elemet_segment

validate_data

rossberg:
Fix #1711.


rossberg:
@conrad-watt, ping.

rossberg:
@conrad-watt, ping again.

conrad-watt (approved):
LGTM - this will change again with the GC proposal, where global initialisers can refer to previously-declared globals, right?

rossberg:
Yes, that's right.

PR #1715, opened 2024-01-01
test(binary.wast): adjust br_table length test to work with streaming validation

rvagg:
Current implementation fails on streaming validation due to interpreting `0x020x0b` as block with type 11 which doesn't exist; so add enough types such that `0x0b` resolves to a valid one, thereby passing validation and erroring at the expected state.

This was previously reported in https://github.com/WebAssembly/spec/issues/1355 but never fixed, and I've just discovered it for myself.

Suggestions for making this more realistic or safe welcome! I'm a bit concerned about adding so many duplicate types in here but I don't believe that's illegal according to spec. My first pass at this changed the first `0x0b` to an `0x00` but that breaks the realism a bit too much since you need the 3 breaks and the wild `0x00` makes no sense. The test is trying to get at the br_table having a missing byte.

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -1085,8 +1085,19 @@
 (assert_malformed
   (module binary
     "\00asm" "\01\00\00\00"
-    "\01\04\01"                             ;; type section
+    "\01\25\0c"                             ;; type section
     "\60\00\00"                             ;; type 0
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Thanks!

PR #1717, opened 2024-01-04
[interpreter]: Add depends to dune-project

muqiuhan:
Add depends to `dune-project` so that dependencies can be installed by calling `opam install . --deps-only --with-doc --with-test` and then build the interpreter from source :)

### DIFF
--- a/interpreter/dune-project
+++ b/interpreter/dune-project
@@ -3,21 +3,32 @@
 (name wasm)
 
 (generate_opam_files true)
+
 (using menhir 2.1)
+
 (implicit_transitive_deps false)
 
 (license Apache-2.0)
 
 (source
-  (github WebAssembly/spec))
+ (github WebAssembly/spec))
 
 (authors "Andreas Rossberg <rossberg@mpi-sws.org")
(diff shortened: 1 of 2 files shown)


rossberg (commented):
Thanks, looks good to me. Can you minimise the PR such that it avoids reformatting unrelated lines?

muqiuhan:
Thank you for your reply, I have removed the formatting related lines.

rossberg (approved):
Thanks!

zapashcanon:
I believe this is not correct and should be reverted.

The interpreter (`wasm` executable) does not depends on `js_on_ocaml` or `js_of_ocaml-ppx`, only the `wast` executable does. This executable is not part of the opam package and shouldn't impose its dependencies.

People won't be happy about having to install some ppx trash for this package...

One solution would be to mark them as development dependencies with e.g. `(js_of_ocaml :dev)`, but I'd rather simply remove them as they're not really development dependencies...

rossberg:
Fair enough, I reverted the change.

PR #1720, opened 2024-01-17
[spec] Changed sx to sx? in vcvtop_zero

702fbtngus:
image
image
As seen above, the instruction 'demote' is classified as vcvtop, and 'f32x4.demote_f64x2_zero' should be executed with below rule.

image
However, this rule is only applied with the presence of 'sx', while 'f32x4.demote_f64x2_zero' does not contain 'sx', Thus I changed the rule $`t_2\times N.vcvtop\_t_1\times M\_sx\_zero`$ with $`t_2\times N.vcvtop\_t_1\times M\_sx^?\_zero`$.


document/core/exec/instructions.rst
```diff
@@ -837,7 +837,7 @@ where:
    \end{array}
 
 
-:math:`t_2\K{x}N\K{.}\vcvtop\K{\_}t_1\K{x}M\K{\_}\sx\K{\_zero}`
+:math:`t_2\K{x}N\K{.}\vcvtop\K{\_}t_1\K{x}M\K{\_}\sx^?\K{\_zero}`
 ...............................................................
```
rossberg, line 841:
```suggestion
.................................................................
```
(otherwise Sphinx will annoy with warnings)

rossberg (approved):
Makes sense, thanks!

PR #1721, opened 2024-01-17
[spec] Fix argument order in shape.replace_lane instruction of execution

HoseongLee:
The order of the arguments in the instruction shape.replace_lane is different to the order of the pop in the execution.
The execution pops t1.const c1 first even though it is presented as the left-most argument. 

image*Image of the original spec*

This pull request changes the order of the arguments to match the pop operations and the subscripts of variables accordingly.


rossberg (approved):
Thanks!

PR #1722, opened 2024-02-05
[spec] Add missing text format for `f32x4` and `f64x2`

gfgafn:
These instructions were include in abstract syntax instructions  but missing in text format instructions.
reference [syntax-vfunop](https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-vfunop)


rossberg (approved):
Thanks!

PR #1725, opened 2024-02-11
[test] Tests for period-less float literals

rossberg:
Tests for parsing of float literals that are integers. Related to [this wabt bug](https://github.com/WebAssembly/wabt/issues/2388).


keithw (approved):
lgtm. Two of the new assertions (`f64.min_int32` on line 184 and `f64_dec.min_int32` on line 198) fail with wasm2c->MSVC (https://github.com/WebAssembly/wabt/actions/runs/7864206734/job/21455684596) and pass with the proposed fix.

PR #1729, opened 2024-02-17
Remove failure case with ambiguous error options

rvagg:
This failure case could be equally valid as "type mismatch" if a validator is validating one opcode at a time rather than consuming them all then validating the return type.

It wants a `ref.func` but the first opcode it finds is an `i32.const` so a streaming validator should be able to error on the first opcode rather than wait for more to decide that it's not a constant expression.

I can't think of an alternative to replace this with, so I'm suggesting this case is unnecessary and too ambiguous.

### DIFF
--- a/test/core/elem.wast
+++ b/test/core/elem.wast
@@ -532,14 +532,6 @@
   "constant expression required"
 )
 
-(assert_invalid
-  (module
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Fair enough.

PR #1737, opened 2024-03-02
[spec] Fixing module typing for active elem/data segments

raoxiaojia:
Under the current module typing rules, if a module defines any module elem/data segment with an active mode, then it cannot be well-typed: this is because the definition requires each elem/data to be well-typed under the C' context, but no tables and memories exist there.

This proposed change fixes this issue by allowing the module elem and data segments to access the full tables and memories accessible in the module.

### DIFF
--- a/document/core/valid/modules.rst
+++ b/document/core/valid/modules.rst
@@ -542,7 +542,12 @@ Instead, the context :math:`C` for validation of the module's content is constru
 
   * :math:`C'.\CFUNCS` is the same as :math:`C.\CFUNCS`,
 
+  * :math:`C'.\CTABLES` is the same as :math:`C.\CTABLES`,
+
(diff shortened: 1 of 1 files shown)


rossberg (commented):
Thanks! I think you need to fix it in the Latex rule as well.

raoxiaojia:
> Thanks! I think you need to fix it in the Latex rule as well.

You are right -- done unless I've missed anything else!

rossberg (approved):

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

PR #1738, opened 2024-03-05
[CI] Publish to W3C only on push to main

sideshowbarker:
This change splits the publish-to-W3C-TR job out into its own separate workflow, and only runs that workflow for pushes to the main branch — but not for pushes to pull-request branches.


rossberg (approved):
Thanks!

PR #1741, opened 2024-04-02
[spec] Fix refs field in context in appendix

rossberg:
@raoxiaojia, please take a look again.


raoxiaojia:
This should work indeed!

PR #1746, opened 2024-04-27
[spec] Check and fix cross-references

rossberg:
Since neither Sphinx nor MathJax provide any diagnostics for it, this PR adds a simple script to check that all xrefs in the macros.def file are well-defined — and fixes various definitions on the way.


tlively (approved):
Can't say I've fully grokked the bash script, but having this check looks like a massive improvement.

PR #1751, opened 2024-05-08
Add tests for alignment overflow

### LINKED ISSUE #1749: Test extreme alignments
Currently tests only test alignment slightly greater than usual. It would be useful to test extreme alignments like 9223372036854775808, we think. These can often fail in UBSAN.

SoniEx2:
let's see what CI says about these...

closes #1749

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -220,7 +220,7 @@ let zero s = expect 0x00 s "zero byte expected"
 
 let memop s =
   let align = u32 s in
-  require (I32.le_u align 32l) s (pos s - 1) "malformed memop flags";
+  require (I32.lt_u align 32l) s (pos s - 1) "malformed memop flags";
   let offset = u32 s in
   Int32.to_int align, offset
 
--- a/test/core/align.wast
+++ b/test/core/align.wast
@@ -864,3 +864,120 @@
 (assert_trap (invoke "store" (i32.const 65532) (i64.const -1)) "out of bounds memory access")
 ;; No memory was changed
 (assert_return (invoke "load" (i32.const 65532)) (i32.const 0))
+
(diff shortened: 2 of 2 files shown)


SoniEx2:
heh. looks like the test is working as intended. ^^

but also. uh, how do you fix that one?

```
../test/core/_output/align.wast.bin.wast.wast:1269.47-1269.54: syntax error: alignment must be a power of two
```

```wast
(assert_invalid
  (module
    (type $0 (func))
    (memory $0 1)
    (func $0 (type 0) (i32.const 0) (i32.load align=0) (drop))
  )
  "alignment must not be larger than natural"
)
```

rossberg:
To assert malformed text format, you'll need to use `module quote`, which defers parsing.

SoniEx2:
take a look at the filename again, it's a roundtrip(?) one. we found a bug in the roundtrip(?) function.

rossberg:
Ah, actually, the bug is in the decoder not considering 2^32 malformed (which then overflows into 0 when converting to text).

Can you try changing the `le_u` in interpreter/binary/decode.ml:223 to `lt_e`?

rossberg (approved):
Thanks. (But please never force-push to a branch under review, that breaks GH PR mechanics.)

SoniEx2:
oh, really? sorry, we're so used to force-pushing for other projects we contribute to, but we'll keep it in mind.

rossberg:
Well, it's a common mistake, but you shouldn't do that for any GH project. As just one example of what breaks, [here](https://github.com/WebAssembly/spec/pull/1751/files/b2e95b5d1231ad49d0b381a1af39483e5697c79c..9a09f09f6c6191706ae4a39b6c0aee905f3029ae) is what happens to commit links; reviewers lose relative diffs and diff links, comments get orphaned or vanish, etc.

PR #1753, opened 2024-06-13
Test that exporting random globals does not change memory semantics

fitzgen:
@rossberg PTAL, thanks!

### DIFF
--- a/test/core/memory.wast
+++ b/test/core/memory.wast
@@ -240,3 +240,28 @@
   "(import \"\" \"\" (memory $foo 1))"
   "(import \"\" \"\" (memory $foo 1))")
   "duplicate memory")
+
+;; Test that exporting random globals does not change a memory's semantics.
+
+(module
+  (memory (export "memory") 1 1)
+
+  ;; These should not change the behavior of memory accesses.
+  (global (export "__data_end") i32 (i32.const 10000))
+  (global (export "__stack_top") i32 (i32.const 10000))
+  (global (export "__heap_base") i32 (i32.const 10000))
+
+  (func (export "load") (param i32) (result i32)
+    (i32.load8_u (local.get 0))
+  )
+)
+
+;; None of these memory accesses should trap.
+(assert_return (invoke "load" (i32.const 0)) (i32.const 0))
+(assert_return (invoke "load" (i32.const 10000)) (i32.const 0))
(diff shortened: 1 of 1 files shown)


rossberg (commented):
Hm, I mean, this looks fine, but also like a totally arbitrary test. Is there some context that I'm missing?

fitzgen:
There are engines that exist today that fail this spec test, therefore I think it is useful to have.

rossberg (approved):
Interesting, but I'm curious how they possibly manage to fail this? Would they fail with other kinds of exports of these names as well?

fitzgen:
I believe they are pattern matching exactly these global exports, as `wasm-ld` leaves them behind after linking.

(Also, I don't have permissions to merge in this repo, so you or someone else will have to do it for me. Thanks!)

sjamesr:
@fitzgen I'm looking at WAMR, which fails this test. Can you help me understand how an engine should handle this export? Is there a reference/spec I can look at?

fitzgen:
Hello @sjamesr,

> the module loader uses these globals to resize the available memory

This behavior is not standards compliant and is in violation of WebAssembly's semantics.

> Can you help me understand how an engine should handle this export?

Engines should handle these exported globals the same way they would handle any other exported entity: by exporting them from the instance without otherwise changing the behavior of, for example, unrelated memories defined within that instance.

> Is there a reference/spec I can look at?

It is difficult choose a single link in the spec to share, because the non-standard compliant behavior is by definition not defined within the spec. Here are a couple links related to exports and memories:

* Exports at runtime: https://webassembly.github.io/spec/core/exec/runtime.html#syntax-exportinst

  Notably, this does not have any text about changing the size of memory based on the name or value of an exported global.

* Memories at runtime: https://webassembly.github.io/spec/core/exec/runtime.html#syntax-meminst

  In particular, note this bit:

  > The length of the vector always is a multiple of the WebAssembly page size, which is defined to be the constant 65536 – abbreviated 64Ki.

-------------

I encourage engines that wish to support smaller-than-64Ki memories, while simultaneously being good actors within the Wasm ecosystem and its standards community, to do the following:

* Remove all behavior that is not standards compliant.
* Implement support for the standards-track [custom-page-sizes proposal](https://github.com/WebAssembly/custom-page-sizes).
* Require the use of flags or configuration options by users to opt into the custom-page-sizes proposal until it reaches stage 4 of the standardization process.

PR #1757, opened 2024-06-18
Rename data segment expression test message

CharlieTap:
Something small that confused me whilst working on validation, the message "unknown global" or generally "unknown x" elsewhere is used for tests where a global/memory/table is missing. This test however seems to be checking that the expression of the data segment is constrained to just const instructions.

### DIFF
--- a/test/core/data.wast
+++ b/test/core/data.wast
@@ -83,11 +83,11 @@
 
 (assert_invalid
   (module (memory 1) (global i32 (i32.const 0)) (data (global.get 0) "a"))
-  "unknown global"
+  "const expression required"
 )
 (assert_invalid
   (module (memory 1) (global $g i32 (i32.const 0)) (data (global.get $g) "a"))
-  "unknown global"
+  "const expression required"
 )
 
 


rossberg:
The message is intentional. Global.get is generally allowed in constants, but only with _imported_ (and immutable) globals. The spec models that by setting it up such that non-imported globals are simply not in scope for constants. The reference interpreter implements that literally and hence produces this error message. As you can see from the CI failure, a failure to match the message will make it fail the test.

CharlieTap:
Ah I see, is the idea that referencing not imported globals is pointless because you could simply embed the const expression directly? whereas imported globals can be the result of some modules prior computation?

CharlieTap:
How would you validate that a global is not an imported given the context?

image
I imagine the list of global types is a mix of imports and non imports, but if we are just representing them by their types, theres seemingly not enough information to derive the validation 🤔

rossberg:
> Ah I see, is the idea that referencing not imported globals is pointless because you could simply embed the const expression directly? whereas imported globals can be the result of some modules prior computation?

Yes, and it also was a simple measure to prevent cyclic initialisers. Note however that this restriction has been relaxed with the advent of the GC proposal, with which later globals can now [refer to previously defined globals](https://webassembly.github.io/gc/core/valid/modules.html#xref-syntax-modules-syntax-global-mathit-global-ast).

> How would you validate that a global is not an imported given the context?

Simply by only putting the imports in the context for validating constants, see the context C' in the [module typing rule](https://webassembly.github.io/spec/core/valid/modules.html#valid-module). That's exactly why the interpreter gives the error message you observed. ;)

PR #1763, opened 2024-07-02
[spec] Small typo in validation of vector load/store_lane

raoxiaojia:
The latex version of validation for vector load/store_lane currently use strict equalities for alignment bounds instead of the intended `\leq`.


rossberg (approved):
Thanks!

PR #1771, opened 2024-07-29
[js-api] Fix bikeshed errors

Ms2ger:
This does not fix the following issue:
```
No 'dfn' refs found for 'object address'.
```


rossberg (approved):
Thanks!

PR #1772, opened 2024-07-31
add missing cases in name printing escaping

zapashcanon:
According to how names are parsed in `text/lexer.mll`, it looks like some escaping was missing in the text-format printing ?


rossberg (approved):
Indeed, thanks.

PR #1773, opened 2024-08-02
[test] Add tests checking that multiple sections aren't supported

dzfrias:
Only the start section had a test assuring its uniqueness. Cases in which there were two or more of the other sections weren't failing when implementing LibWasm for the Ladybird browser.

I'm not sure if these test cases are totally needed, but LibWasm's `Module` type used to have an interesting way to represent sections, and these cases slipped through... so feel free to close this PR if it's deemed unnecessary.

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -1144,3 +1144,117 @@
   )
   "unexpected content after last section"
 )
+
+;; Multiple function sections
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Definitely good to have these tests, thanks!

PR #1775, opened 2024-08-07
[test] Add negative test for invalid section ordering

dzfrias:
As far as I know, there are no tests that suggest the correct order of sections. Binary modules with out-of-order sections are not failing in the Ladybird WebAssembly implementation, LibWasm.

### DIFF
--- a/test/core/binary.wast
+++ b/test/core/binary.wast
@@ -1258,3 +1258,116 @@
   )
   "unexpected content after last section"
 )
+
+;; Type section out of order
+(assert_malformed
+  (module binary
+      "\00asm" "\01\00\00\00"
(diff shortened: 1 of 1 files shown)


rossberg (commented):
Thanks, definitely good to test this! Can I motivate you to add versions for other section pairs as well? It should be enough to test consecutive pairs.

dzfrias:
Ok, I added tests for all the section pairs!

rossberg (approved):
Thanks!

(FYI, please never force-push to branches under review, since that breaks GH's review work-flow. Didn't matter much for this small PR, but would for more complex ones.)

dzfrias:
Ok, good to know for next time!

Issue #1774, opened 2024-08-06
Fix CSS so the specification shows on narrow viewports without horizontal scrolling.


matj1:
The width of the pages of the specification (like [this](https://webassembly.github.io/spec/core/text/instructions.html)) is hardcoded to 1000px. That means that, if I view the webpage with split screen, it shows a horizontal scrollbar and I can't see whole lines without scrolling.

Specifically, I have the browser window at a half of a Full HD monitor, so the width of the webpage is \[960 − \[the scrollbar's width\]\] pixels, which is 945 pixels according to the inspector. So 55 pixels are hidden. I need to scroll to see them even if the content is mostly alright when it is narrower.

## Solution

- Remove `width: 1000px` from `div.document` and `div.footer` in *custom.css*.
- Also remove `width: 940px` from `div.document` and `div.footer` in *alabaster.css*.

The second thing is to allow the page to be even narrower, which works until 860 pixels, when the layout breaks both with and without this fix.

The following shows which CSS rules are supposed to be removed; these are the crossed out ones with empty checkboxes.

![Snímek obrazovky z 2024-08-06 16-25-46](https://github.com/user-attachments/assets/b6506d33-8e47-4257-a0ed-b84523238153)  
![Snímek obrazovky z 2024-08-06 16-24-48](https://github.com/user-attachments/assets/4d12c19a-80b0-4398-8b25-814424eef722)

rossberg:
Good idea. Would you be willing to contribute a PR?

matj1:
Probably. But alabaster.css is from a theme built in sphinx, so I can't change it. So I change the the mentioned cases in custom.css `width: 1000px` to `width: unset`, which should do the same.

matj1:
To which branch should I commit it? main is protected, and I don't know what other branches are for.

rossberg:
You'll probably have to fork and create a pull request.

Does the footnote size need changing, too?

matj1:
The footnote size too needs changing because that too is fixed to 1000 pixels.

matj1:
It seems that I shouldn't make a pull request according to the [contributing guidlines](https://github.com/WebAssembly/design/blob/main/Contributing.md) because I am not in W3C Community Group, and I don't bother to join it just because of this. At worst, the pull request will be rejected, but then I encourage everyone there to make this change in my stead.

rossberg:
[...]

PR #1777, opened 2024-08-09
Unset the width of documents in CSS

matj1:
This is to allow that the documents of the specification display well on a narrow viewport. On viewports 100px or more wide, it does nothing, and, on viewports less than 1000px wide, it shrinks the document to the width of the viewport.

The issue is discussed here: #1774


rossberg (approved):
Thanks!

PR #1784, opened 2024-08-18
[test] Add tests ensuring table type is grown in `table.grow`

dzfrias:
No previous tests check that `table.grow` grows the table instance's type. This was found when implementing LibWasm for the Ladybird browser.

### DIFF
--- a/test/core/imports.wast
+++ b/test/core/imports.wast
@@ -378,6 +378,26 @@
 (assert_trap (invoke "call" (i32.const 3)) "uninitialized element")
 (assert_trap (invoke "call" (i32.const 100)) "undefined element")
 
+(module $Tgt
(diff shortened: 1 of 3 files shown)


rossberg (commented):
The test looks good to me, but this should go to table_grow.wast (imports are only used as a means for observing the table's runtime type).

dzfrias:
A similar test for `memory.grow` is in `imports.wast`. Should I move both?

rossberg:
Ah, good point. Yeah, that is ancient, and worth moving as well.

dzfrias:
Ok, I just moved them to their respective files. Thank you!

rossberg (approved):
Excellent, thanks!

PR #1786, opened 2024-08-19
Add more tests for `sqrt`.

sunfishcode:
Add a sqrt test for values around ¼, where `sqrt(x) - x` is greatest.

And add a few more sqrt test values based on divergences in sqrt implementations found in the wild.


titzer (approved):
lgtm; rubber-stamped

rossberg (approved):

Issue #584, opened 2017-10-16
Memory allocation/growth failure


littledan:
I'm probably reading the specification wrong, but it seems like [allocating memory](https://webassembly.github.io/spec/exec/modules.html#alloc-mem) is specified to be infallable, but [growing memory](https://webassembly.github.io/spec/exec/instructions.html#exec-grow-memory) is specified to allow failure. Is this intentional?

If memory allocation fails during module instantiation, should the generated exception be LinkError, the browser-specific OOM error/crash, or RangeError?

If memory growth triggered by the Wasm executable code fails, should the generated JavaScript exception be RuntimeError, the browser-specific OOM error/crash, or RangeError?

(I was expecting all of these to be the browser-specific OOM error, but I see different things in the spec documents. I'm not sure if this is intentional or not.)

lukewagner:
Memory allocation during instantiation is fallible and should throw.  `WebAssembly.Memory.prototype.grow` throws a `RangeError` and IIRC that was chosen intentionally for symmetry with other JS builtins, so I think instantiation should throw a `RangeError`.  (FF gets that wrong, atm, but easy to fix.)

> If memory growth triggered by the Wasm executable code fails, should the generated JavaScript 
> exception be RuntimeError, the browser-specific OOM error/crash, or RangeError?

 `grow_memory` in executable code never throws, but is defined to return -1 (so that wasm can test and handle).

lukewagner:
Actually, I could see the argument for either `RangeError` or whatever the engine normally does for a failed GC allocation (e.g., when `new WebAssembly.Memory(x)` fails) which seems to be different in FF.

rossberg:
The fallibility of allocating memory is supposed to be covered by the [Appendix on implementation limitations](https://webassembly.github.io/spec/appendix/implementation.html). The grow_memory instruction is special because as @lukewagner mentions, it does not actually cause failure but returns regularly in that case. So that's why it has to make that case explicit (in the form of non-determinism).

littledan:
@rossberg Thanks for the pointer to that appendix. The wording there is a bit more specific than I was expecting. Under the execution heading, I see that an implementation can place limits on the number of allocated modules or the size of an allocated module. This sounds a little like there's a fixed number for each of these limits, that a user might discover by a bunch of testing (which isn't quite how running out of memory works; I imagine trying to implement that would reduce the available memory/memories). Is that the intention?

[...]

PR #1793, opened 2024-08-27
[js-api] Fix up missing preconditions on allocations

### LINKED ISSUE #1792: [js-api] Memory and table allocation pre-conditions not validated
The [mem_alloc](https://webassembly.github.io/spec/core/appendix/embedding.html#mathrm-mem-alloc-xref-exec-runtime-syntax-store-mathit-store-xref-syntax-types-syntax-memtype-mathit-memtype-xref-exec-runtime-syntax-store-mathit-store-xref-exec-runtime-syntax-memaddr-mathit-memaddr) and [table_alloc](https://webassembly.github.io/spec/core/appendix/embedding.html#mathrm-table-alloc-xref-exec-runtime-syntax-store-mathit-store-xref-syntax-types-syntax-tabletype-mathit-tabletype-xref-exec-runtime-syntax-ref-mathit-ref-xref-exec-runtime-syntax-store-mathit-store-xref-exec-runtime-syntax-tableaddr-mathit-tableaddr) operations in the wasm embedding spec have pre-conditions asserting that the memory and table types are valid. The JS API spec does not actually ensure that these pre-conditions are met, which means that it is not clear which kind of error should be thrown if the type is invalid.

For example, mem_alloc has a pre-condition that the memtype is valid, which means that the limits must be valid within range 2^16. The JS API definition for [`Memory(descriptor)`](https://webassembly.github.io/spec/js-api/index.html#dom-memory-memory) asserts that initial <= maximum, but not that initial <= 2^16 and maximum <= 2^16. What type of error should be thrown if the given memory type is invalid?

(table_alloc technically has the same problem, although EnforceRange makes it impossible to express an out-of-bounds size.)

bvisness:
Adds explicit steps to run memtype and tabletype validation and throw a RangeError on failure. This explicitly satisfies the preconditions from the wasm embedding spec, and removes the need for the `initial <= maximum` check in the JS API spec.

One other small note: allocation failures on tables are now explicitly reported. This captures implementation limits on table length, which also weren't quite handled by the spec. (This removes a TODO referencing #584.)

Fixes #1792.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -51,6 +51,8 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     url: valid/modules.html#valid-module
         text: valid
         text: WebAssembly module validation
+    text: valid memtype; url: valid/types.html#valid-memtype
+    text: valid tabletype; url: valid/types.html#valid-tabletype
     text: module grammar; url: binary/modules.html#binary-module
     text: custom section; url: binary/modules.html#custom-section
     text: customsec; url: binary/modules.html#binary-customsec
@@ -101,6 +103,7 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     text: table type; url: syntax/types.html#syntax-tabletype
     text: table address; url: exec/runtime.html#syntax-tableaddr
     text: function address; url: exec/runtime.html#syntax-funcaddr
+    text: memory type; url: syntax/types.html#syntax-memtype
     text: memory address; url: exec/runtime.html#syntax-memaddr
     text: global address; url: exec/runtime.html#syntax-globaladdr
     text: extern address; url: exec/runtime.html#syntax-externaddr
@@ -658,8 +661,8 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     The <dfn constructor for="Memory">Memory(|descriptor|)</dfn> constructor, when invoked, performs the following steps:
     1. Let |initial| be |descriptor|["initial"].
     1. If |descriptor|["maximum"] [=map/exists=], let |maximum| be |descriptor|["maximum"]; otherwise, let |maximum| be empty.
-    1. If |maximum| is not empty and |maximum| &lt; |initial|, throw a {{RangeError}} exception.
-    1. Let |memtype| be { min |initial|, max |maximum| }.
+    1. Let |memtype| be the [=memory type=] { **min** |initial|, **max** |maximum| }.
+    1. If |memtype| is not [=valid memtype|valid=], throw a {{RangeError}} exception.
     1. Let |store| be the [=surrounding agent=]'s [=associated store=].
     1. Let (|store|, |memaddr|) be [=mem_alloc=](|store|, |memtype|). If allocation fails, throw a {{RangeError}} exception.
     1. Set the [=surrounding agent=]'s [=associated store=] to |store|.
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -819,14 +822,14 @@ Each {{Table}} object has a \[[Table]] internal slot, which is a [=table address
         1. [=Throw=] a {{TypeError}} exception.
     1. Let |initial| be |descriptor|["initial"].
     1. If |descriptor|["maximum"] [=map/exists=], let |maximum| be |descriptor|["maximum"]; otherwise, let |maximum| be empty.
-    1. If |maximum| is not empty and |maximum| &lt; |initial|, throw a {{RangeError}} exception.
+    1. Let |type| be the [=table type=] { **min** |initial|, **max** |maximum| } |elementType|.
+    1. If |type| is not [=valid tabletype|valid=], throw a {{RangeError}} exception.
```
Ms2ger, line 826: Since IDL already enforces most of this, I would prefer keeping the existing check, and adding an "Assert: type is valid" line after it

document/js-api/index.bs
```diff
@@ -658,8 +661,8 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     The <dfn constructor for="Memory">Memory(|descriptor|)</dfn> constructor, when invoked, performs the following steps:
     1. Let |initial| be |descriptor|["initial"].
     1. If |descriptor|["maximum"] [=map/exists=], let |maximum| be |descriptor|["maximum"]; otherwise, let |maximum| be empty.
-    1. If |maximum| is not empty and |maximum| &lt; |initial|, throw a {{RangeError}} exception.
-    1. Let |memtype| be { min |initial|, max |maximum| }.
+    1. Let |memtype| be the [=memory type=] { **min** |initial|, **max** |maximum| }.
```
Ms2ger, line 664: Please also link min/max here

Ms2ger (changes requested):
Thanks; some small comments below

document/js-api/index.bs
rossberg: The IDL cannot enforce _semantic_ well-formedness of Wasm types, which is what's needed here.

Ms2ger: I don't understand your comment

rossberg: Not every _syntactically_ expressible table/memory (or other) type is actually legal. Hence, the Wasm semantics has to validate types themselves, pretty much the same way it validates instructions. Where the JS API allows expressing types in JS, we must hence validate them according to Wasm rules before using them.

bvisness: I'm not sure I quite understand which semantic issues @rossberg is referring to, but I guess to me it seems like it comes down to whether `[EnforceRange] unsigned long` is equivalent enough to wasm's `u32`. To me they do seem generally equivalent enough to not cause problems. After all, `n <= k`, `n <= m`, and `m <= k` should all evaluate exactly the same whether you operate semantically on EnforceRange'd JS numbers or on `u32`.

Still, to me it feels more natural to leverage the wasm spec's notion of validity here instead of duplicating the checks in the JS API. For one, wasm engines already have validation code around, and for another, calling out to the wasm spec is more "future-proof" in that updates to validation in the core spec will be implicitly reflected in the JS API spec. (This is especially on my mind as I work to finalize the spec for memory64.)

Ms2ger: More concretely, given the current state of the core spec, what possible way could there be for table type created from u32 values min and max, with min <= max, and reftype either funcref or externref, to fail to be [valid](https://webassembly.github.io/spec/core/valid/types.html#valid-tabletype)?

I could understand an argument about the robustness to potential changes in the core spec (though the assertion I suggest would cover that fine). I just don't understand your suggestion that the current rules are impossible to verify from "the outside".

As an alternative, I'd at least add a note that the `0 <= n < 2 ** 32` checks cannot fail due to the conversion in the IDL layer.

rossberg: While, for this moment, type validity still is relatively simple (here: proper ordering of limits, and checking the upper limit of memtypes is ≤2^16), that changes with function references or GC types, which are already included on the staging wasm-3.0 branch, and with whom it involves a recursive algorithm.

It's also a question of spec modularity and hygiene not to duplicate definitions that are "owned" by other parts of a spec if possible. That just risks incoherence and possible hick-ups. An assertion does _not_ magically cover mistakes in such a case — rather, an assertion that doesn't actually hold is a spec bug in itself.

Ms2ger: Alright, let's just stick with the note about the integer limits, then

PR #1780, opened 2024-08-14
Add a skeleton updated README/Overview

alexcrichton:
No actual content just yet, just placeholders for future content.

PR #1800, opened 2024-09-10
Fix grammar/typo in overview

fitzgen:


alexcrichton:
I think this may be the wrong repo? (I made the [same mistake myself](https://github.com/WebAssembly/spec/pull/1780) awhile back)

GitHub I think doesn't do a great job about sending a PR to a fork of a repo since it seems to want to by default target the original repository instead.

fitzgen:
D'oh.

PR #1801, opened 2024-09-12
Test data count without data segment

SoniEx2:
It's a little more verbose than it needs to be due to originally being minified from a wabt exploit.


SoniEx2:
oh right, CI doesn't show the errors because they're logged into a file instead

rossberg (commented):
Looks good, but there are a bunch of tests for the data count section in binary.wast already, so it would make sense to move this one next to those.

SoniEx2:
there we go :)

rossberg (approved):

PR #1829, opened 2024-10-07
[interpreter] Fix JS module invocation

### LINKED ISSUE #1828: js-api tests broken after #1796
I don't have time to investigate this further currently (and my OCaml knowledge is limited), but after #1796 we now fail most tests in the memory64 and js-promise-integration repositories, which seem to be rebased on a branch which contains that change.

The resulting JS code looks like
```
// address.wast:3                                                           
let $$1 = "\x00\x61\x73\x6d [...] \x7a";

// address.wast:3
let $1 = instance($$1);
````

Where `instance` simply does:
```
function instance(mod, imports = registry) {
  return new WebAssembly.Instance(mod, imports);
}
```

This can't work.

To reproduce, run something like:
`make -C interpreter wasm && (cd test/core && ./run.py --wasm ../../interpreter/wasm --js $HOME/v8/out/Optdebug/d8 address.wast)`

This produces an error like:
```
test/core/_output/address.js:62: TypeError: WebAssembly.Instance(): Argument 0 must be a WebAssembly.Module
  return new WebAssembly.Instance(mod, imports);
         ^
TypeError: WebAssembly.Instance(): Argument 0 must be a WebAssembly.Module
    at instance (test/core/_output/address.js:62:10)
    at test/core/_output/address.js:203:10
```

rossberg:
Fixes #1828.

### DIFF
--- a/interpreter/script/js.ml
+++ b/interpreter/script/js.ml
@@ -614,13 +614,16 @@ let rec of_result res =
   | EitherResult ress ->
     "[" ^ String.concat ", " (List.map of_result ress) ^ "]"
 
+let of_module bs =
+  "module(" ^ of_bytes bs ^ ")"
+
 let rec of_definition def =
   match def.it with
-  | Textual (m, _) -> of_bytes (Encode.encode m)
-  | Encoded (_, bs) -> of_bytes bs.it
+  | Textual (m, _) -> of_module (Encode.encode m)
+  | Encoded (_, bs) -> of_module bs.it
   | Quoted (_, s) ->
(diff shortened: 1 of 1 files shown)


backes (approved):
Thanks, this seems to fix it.

backes:
I am going ahead and merging this, so we can import the new tests in V8.

backes:
Sorry, I didn't test enough, this fix breaks other things :/ 

`assert_invalid` is now being translated into `assert_invalid(module(bytes))`, which then fails in `module` already. We should pass the raw bytes instead.

rossberg:
Ah, I shouldn't do late night fixes. Please try #1832.

PR #1832, opened 2024-10-08
[interpreter] Proper fix for JS conversion

rossberg:
This reverts #1829 and implements a simpler & correcter fix.

### DIFF
--- a/interpreter/script/js.ml
+++ b/interpreter/script/js.ml
@@ -677,9 +677,9 @@ let of_assertion env ass =
   | AssertInvalidCustom (def, _) ->
     "assert_invalid_custom(" ^ of_definition def ^ ");"
   | AssertUnlinkable (x_opt, _) ->
-    "assert_unlinkable(" ^ of_instance env x_opt ^ ");"
+    "assert_unlinkable(" ^ of_mod_opt env x_opt ^ ");"
   | AssertUninstantiable (x_opt, _) ->
-    "assert_uninstantiable(" ^ of_instance env x_opt ^ ");"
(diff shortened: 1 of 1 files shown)


backes:
This makes most tests pass, thanks!

There are still other issues, but that's probably unrelated. E.g. in the generated `type-rec.js`:
```
223 // type-rec.wast:69
224 let $$3 = module("\x00 [...] \x0b");
225 let M = $$3;
226
227 // type-rec.wast:69
228 let $3 = instance(M);
229 let M = $3;
```
(double definition of `M`)

backes (approved):

backes:
At some point someone should just run all tests in `wasm-3.0` against a JS engine and fix all those issues. I can't fix anything in the OCaml code, so I am not the right person for this. Maybe @sbc100 is already doing that as part of #1828.

rossberg:
Ah, just fixed that as well, can you try now? (Sorry I can test this on my end, as I am travelling.)

backes:
No problem, I can try and report back.

We now have a similar problem with
```
243 // type-rec.wast:80
244 assert_unlinkable(instance($$5));
```

This should be `assert_unlinkable(module($$5))`.

rossberg:
Ah, sorry, next try. :)  (It should actually just be `assert_unlinkable($$5)`, I think.)

rossberg:
Landed. Thanks for your patience. :)

backes:
Thanks for fixing this!

I created PRs to merge the new wasm-3.0 branch into memory64 and jspi (I hope this is how this works):
https://github.com/WebAssembly/memory64/pull/89
https://github.com/WebAssembly/js-promise-integration/pull/50

PR #1836, opened 2024-10-28
[interpreter] Fix subtype condition for result patterns

rossberg:
This should fix WebAssembly/memory64#95.

### DIFF
--- a/.github/workflows/ci-interpreter.yml
+++ b/.github/workflows/ci-interpreter.yml
@@ -2,11 +2,11 @@ name: CI for interpreter & tests
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, wasm-3.0 ]
     paths: [ .github/**, interpreter/**, test/** ]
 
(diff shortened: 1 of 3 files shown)


rossberg:
@backes, that is odd, I don't see that error on the wasm-3.0 branch. I amended the PR, that should address the assertion failure you're seeing. Let me know if that works.

rossberg:
Okay, there were a couple of issues with the generation of JS wrapper modules for ref.extern (and also `either` patterns). Those didn't pop up on this branch, because no wrapper modules were needed in any of the cases where ref.extern was used.

I also added a sanity check that all generated modules are valid, and activated CI for the wasm-3.0 branch, so that these problems should get caught earlier in the future.

backes (approved):
Nice, this seems to fix all issues!

backes:
I am merging this, and then creating a PR on `memory64` to merge the updated `wasm-3.0` branch.

PR #1837, opened 2024-10-29
[spec] Add elem/data member to moduleinst in prose

mohanson:
I think these two members are missing here.


rossberg (approved):
Indeed, thanks!

PR #1839, opened 2024-11-06
Tweaks and fixes to Memory64 proposal

rossberg:
In preparation for merging the proposal to the wasm-3.0 branch, this PR fixes various smaller issues I discovered when going through the diffs for the merge.

Spec:
- Removed unnecessary/mismatching lookup of table/memory type in execution prose
- Added missing result type lookup in formal rule for `table.size` and `memory.size`
- Fixed computation of -1 result value for `table.grow` and `table.size` to work for i64
- Some fixes around specification of text format for inline elements/data shorthand
- Fixed matching rules for tabletype/memtype to enforce same address type
- Added an entry for the proposal to the Appendix’ changelog

Interpreter:
- Fixed evaluation of v128 load/store instructions to work with i64
- Reworked bulk operation execution to still reduce to well-typed instructions for i32
- Added missing size check to table allocation
- Various minor refactorings and clean-ups

Tests:
- Added tests for size check in i64 table and memory type limits

@sbc100, PTAL. Feel free to upstream back to the repo as you see fit.

Also, it would be good if you could add some tests for SIMD load/stores on i64 memories, since those were broken in the interpreter.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2725,19 +2725,17 @@ Table Instructions
 
 5. Let :math:`\X{tab}` be the :ref:`table instance <syntax-tableinst>` :math:`S.\STABLES[a]`.
 
-6. Let :math:`\X{at}~\limits` be the :ref:`table type <syntax-tabletype>` :math:`\X{tab}.\TITYPE`.
+6. Assert: due to :ref:`validation <valid-table.get>`, a value of some :ref:`address type <syntax-addrtype>` :math:`\X{at}` is on the top of the stack.
 
-7. Assert: due to :ref:`validation <valid-table.get>`, a value of :ref:`value type <syntax-valtype>` :math:`\X{at}` is on the top of the stack.
+7. Pop the value :math:`\X{at}.\CONST~i` from the stack.
 
-8. Pop the value :math:`\X{at}.\CONST~i` from the stack.
-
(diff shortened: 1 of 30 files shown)


document/core/exec/instructions.rst
```diff
@@ -3459,25 +3440,23 @@ Memory Instructions
 
 5. Let :math:`\X{mem}` be the :ref:`memory instance <syntax-meminst>` :math:`S.\SMEMS[a]`.
 
-6. Let :math:`\X{at}~\limits` be the :ref:`memory type <syntax-memtype>` :math:`\X{mem}.\MITYPE`.
+6. Assert: due to :ref:`validation <valid-load-extend>`, a value of :ref:`adress type <syntax-addrtype>` :math:`\X{at}` is on the top of the stack.
```
bvisness, line 3443: Typo: `adress`

bvisness (approved):
All looks good to me!

document/core/exec/instructions.rst
rossberg: Fixed.

rossberg:
> I do think for the sake of keeping the history of this proposal in one place it might me good to land these changes in the memory64 repo before merging. Is that what you mean by "Feel free to upstream back to the repo as you see fit"? I'm happy to do that.

Yes, sounds good!

PR #1842, opened 2024-11-07
[js-api] Editorial: replace Type AO with new ECMA-262 type test convention

michaelficarra:
ECMA-262 now uses these prose forms for type tests and has removed the old `Type` AO. See https://github.com/whatwg/html/pull/10635 for more info.


Ms2ger (approved):
If that's how it has to be, I guess.

PR #1849, opened 2024-11-27
Test array.new_data and array.new_elem with dropped segments

tlively:
Also test that the instructions trap when the offset and size overflow back to zero, with and without the segments having been dropped.


rossberg (approved):
Test changes lgtm, but is the katex change supposed to be in this PR?

tlively:
Definitely not. I didn't even notice that. Fixed now.

PR #1851, opened 2024-11-30
Fix mobile margin

dy:
At the moment mobile version of site looks broken. It seems that custom.css overrides alabaster.css, which had good mobile media query.
image
This minimal change makes spec readable on mobiles.


rossberg (approved):
Thanks!

PR #1852, opened 2024-12-06
Update references to IEEE 754-2008.

sunfishcode:
Update them to refer to IEEE 754-2019. I've confirmed that the relevant section names and numbers in 754-2019 are the same.


rossberg (approved):
LGTM, thanks.

PR #1862, opened 2025-01-10
Change the definition of `tagtype` from `functype` into `deftype`

f52985:
This PR changes the definition of `tagtype`, which was previously `functype`, into `deftype`,
and also modifies (hopefully) every use site of it, accordingly.

The issue arised from the fact that the `exn` proposal and the `gc` proposal were being developed independently,
and they were not merged properly.

### DIFF
--- a/document/core/valid/matching.rst
+++ b/document/core/valid/matching.rst
@@ -580,12 +580,15 @@ A :ref:`global type <syntax-globaltype>` :math:`(\mut_1~t_1)` matches :math:`(\m
 Tag Types
 ~~~~~~~~~
 
-A :ref:`tag type <syntax-tagtype>` :math:`\tagtype_1` matches :math:`\tagtype_2` if and only if they are the same.
+A :ref:`tag type <syntax-tagtype>` :math:`\deftype_1` matches :math:`\deftype_2` if and only if the :ref:`defined type <syntax-deftype>` :math:`\deftype_1` :ref:`matches <match-deftype>` :math:`\deftype_2`, and vice versa.
 
 .. math::
    \frac{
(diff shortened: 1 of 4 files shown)


document/core/valid/types.rst
```diff
@@ -551,19 +551,25 @@ Memory Types
 Tag Types
 ~~~~~~~~~
 
-:math:`[t_1^n] \to [t_2^m]`
+:math:`\deftype`
 ...........................
```
rossberg, line 555:
```suggestion
:math:`\deftype`
................
```

document/core/valid/matching.rst
```diff
@@ -580,12 +580,15 @@ A :ref:`global type <syntax-globaltype>` :math:`(\mut_1~t_1)` matches :math:`(\m
...
 
 .. math::
    \frac{
+     C \vdashdeftypematch \deftype_1 \matchesdeftype \deftype_2
+     \qquad
+     C \vdashdeftypematch \deftype_2 \matchesdeftype \deftype_1
    }{
-     C \vdashtagtypematch \tagtype \matchestagtype \tagtype
+     C \vdashtagtypematch \deftype_1 \matchestagtype \deftype_2
    }
```
rossberg, line 594:
```suggestion
.. note::
   Although the conclusion of this rule looks identical to its premise,
   they in fact describe different relations:
   the premise invokes subtyping on defined types,
   while the conclusion defines it on tag types that happen to be expressed as defined types.

```

rossberg (approved):
Excellent, thanks a lot!

f52985:
Oh, it seems that I don't have the write access for this repository. @rossberg, could you merge this pull request?

PR #1867, opened 2025-01-28
[spec] Fixed incorrect description of fge operation

gibachan:
Corrected the explanation of `z1 >= z2` to accurately reflect its meaning.


rossberg (approved):
Thanks!

PR #1872, opened 2025-03-02
return_call +indirect with multi-param / multi-result / imported funcs

keithw:
Moving https://github.com/WebAssembly/tail-call/pull/25 here.

This PR adds tests of return_call and return_call indirect where:

    the function being called is an imported function
    the function being called takes multiple params
    the function being called has multiple results

These catch a bug in the WABT interpreter and two bugs in wasm2c (one of which was reported at https://github.com/WebAssembly/wabt/pull/2491).

### DIFF
--- a/test/core/return_call.wast
+++ b/test/core/return_call.wast
@@ -1,6 +1,8 @@
 ;; Test `return_call` operator
 
 (module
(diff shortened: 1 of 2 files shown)


rossberg (approved):
Thanks!

PR #1874, opened 2025-03-04
[spec] Fix validation rule for catch and catch all

ShinWonho:
Screenshot 2025-03-04 at 4 10 25 PM

I noticed that validation rule for `catch` and `catch_all` uses equality rather than subtype.
I have corrected it.


rossberg (approved):
Thanks!

PR #1876, opened 2025-03-05
[spec] Add noexn case for nullref typing

ShinWonho:
I noticed that `noexn` case is missing in null reference typing.
I have added that.
Screenshot 2025-03-05 at 2 33 48 PM


rossberg (approved):
Thanks!

PR #1879, opened 2025-03-07
[interpreter] Fix missing type check for `return_call_indirect`

f52985:
This PR fixes a missing type check for `return_call_indirect` instruction.

`return_call_indirect` expects a function reference, and the module should be considered invalid if the non-function reference is given (as in `call_indirect` instruction).
However, the current validation behaves incorrectly, and pass such invalid module.
This PR fixes this error, and adds the new test case:
```wast
(assert_invalid
  (module
  (type (func))
  (table 10 externref)
  (func $return-call-indirect (return_call_indirect (type 0) (i32.const 0)))
  )
  "type mismatch"
)
```

### DIFF
--- a/interpreter/valid/valid.ml
+++ b/interpreter/valid/valid.ml
@@ -543,6 +543,9 @@ let rec check_instr (c : context) (e : instr) (s : infer_result_type) : infer_in
   | ReturnCallIndirect (x, y) ->
     let TableT (at, _lim, t) = table c x in
     let FuncT (ts1, ts2) = func_type c y in
+    require (match_ref_type c.types t (Null, FuncHT)) x.at
+      ("type mismatch: instruction requires table of function type" ^
+       " but table has element type " ^ string_of_ref_type t);
     require (match_result_type c.types ts2 c.results) e.at
       ("type mismatch: current function requires result type " ^
        string_of_result_type c.results ^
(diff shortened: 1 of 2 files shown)


rossberg (approved):
Nice catch, thanks! Did the test gen find that?

f52985:
> Did the test gen find that?

Yeah, the test generation based on SpecTec found it!

---

By the way, an interesting bit of backstory: discovering this error was somewhat lucky because generating invalid modules isn’t the main goal of this generator.

The generator aims to produce valid modules as much as possible and relies on the reference interpreter's validation to filter out any unintentionally generated invalid ones. However, this particular test case slipped through because the reference interpreter mistakenly considered the module valid. As a result, the module ‘survived’ and got recorded.

Ironically, if the generator were ‘perfect’—meaning it always produced strictly valid modules according to the spec—this bug might never have been discovered!

rossberg:
Nice. Wouldn't it be possible to use a similar approach to systematically generate negative tests for module validation?

f52985:
> Wouldn't it be possible to use a similar approach to systematically generate negative tests for module validation?

Oh, you're absolutely right! In fact, that's exactly what we are ultimately aiming for in the future research.

If we can develop a ‘perfect’ generator—one that strictly produces test cases satisfying a given input validation rule (regardless of whether that rule itself is correct)—then systematically generating negative tests should be done in trivial manner, by deliberately giving the wrong validation rule as an input.

…That said, we’re not quite there yet, as this very example demonstrates.

PR #1881, opened 2025-03-07
[interpreter] Fix bounds check for array.new_data

### LINKED ISSUE #1880: [interpreter] Wrong boundary check for `array.new_data` and `array.init_data`
Interpretation of `array.new_data` and `array.init_data` uses the wrong boundary check condition for data,
resulting in an uncaught exception `Wasm.Data.Bounds`.

For example, here's the interpreter code for the instruction `array.new_data`:
```Ocaml
| ArrayNewData (x, y), Num n :: Num s :: vs' ->
  let n_64 = addr_of_num n in
  let s_64 = addr_of_num s in
  if data_oob c.frame y s n then
    vs', [Trapping (memory_error e.at Memory.Bounds) @@ e.at]
  else
    let ArrayT (FieldT (_mut, st)) = array_type c.frame.inst x in
    ...
```

On the fourth line, `data_oob c.frame y s n` checks if `s + n > |DATA|`,
but it should check if `s + n * (storage_size st) > |DATA|`.
(The multiplication of storage size is missing)

---

Here's the proof of concept input that triggers this uncaught exception:
```wast
(module
  (type $arr (array i32))
  (func (export "f") (result (ref $arr))
    (i32.const 0)
    (i32.const 1)
    (array.new_data $arr $short_data)
  )
  (data $short_data "dat")  ;; 3byte: Should Trap
  (data $long_data  "data") ;; 4byte: OK
)
(assert_trap (invoke "f") "")
```
result:
```
./wasm: uncaught exception Wasm.Data.Bounds
```

---

p.s. It seems a bit weird that the original code intended to throw a `Memory.Bounds` error, despite the fact that the boundary check is performed on data.

rossberg:
Fix #1880.

@f52985, PTAL.


rossberg:
Done, PTAL.

f52985:
Looks perfect! :)

PR #1882, opened 2025-03-10
[spec] Update allocmodule arguments

ShinWonho:
It seems that the arguments of allocmodule have not been updated.

Screenshot 2025-03-10 at 4 23 05 PM

I have corrected it.

Screenshot 2025-03-10 at 4 19 08 PM


rossberg (approved):
Thanks!

PR #1885, opened 2025-03-27
[spec] Fix array.new_default

ShinWonho:
Screenshot 2025-03-27 at 2 55 32 PM

I noticed that validation rule for `array.new_default` uses an incorrect instruction name,

Screenshot 2025-03-27 at 4 39 37 PM

and there is also a minor typo in execution rule that a parenthesis is not in the subscript.

I have corrected it.


rossberg (approved):
Thanks!

PR #1886, opened 2025-03-27
[test] align.wast: add tests of malformed memop flags in wasm-3.0

keithw:
This PR adds tests for malformed binary-format memop flags to the wasm-3.0 tests.

The main-branch `align.wast` has malformed tests (added in https://github.com/WebAssembly/spec/pull/1751), but multi-memory liberalizes the encoding so the existing tests were changed to `assert_invalid` in multi-memory and wasm-3.0. This adds back some (more-extreme) examples of malformed flags.

### DIFF
--- a/test/core/align.wast
+++ b/test/core/align.wast
@@ -962,3 +962,41 @@
   )
   "type mismatch"
 )
+
+;; memop flags >= 0x80
+(assert_malformed
+  (module binary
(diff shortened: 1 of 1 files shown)


rossberg (approved):
LGTM (not sure align.wast is the ideal place, but I'm fine with that)

PR #1887, opened 2025-03-27
[spec] Update limits type in wasm-3.0 text format spec

keithw:
In memory64, the types of _limits_ `min` and `max` were updated to u64 in the abstract syntax and the binary format specification. This updates the text format spec to match.

This is already tested by the wasm-3.0 tests, e.g. this `assert_invalid` test from `memory.wast` is an `assert_malformed` on the main branch:
```wasm
(assert_invalid
  (module (memory 0x1_0000_0000))
  "memory size"
)
```

### DIFF
--- a/document/core/text/types.rst
+++ b/document/core/text/types.rst
@@ -303,7 +303,7 @@ Address Types
 Abbreviations
 .............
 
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Thanks!

PR #1890, opened 2025-04-04
[test/interpreter] Extern convert as const expr

### LINKED ISSUE #1889: [spec/interpreter] Mismatch about constant instructions `extern.convert_any` and `any.convert_extern`
In `wasm-3.0` branch, there is a mismatch between the specification and the interpreter regarding whether `extern.convert_any` and `any.convert_extern` are constant instructions.

The specification treats these instructions as constant, but the reference interpreter's validation does not, causing the following code to be rejected:
```wat
;; const.wat
(module
  (global externref (ref.null any) (extern.convert_any))
  (global anyref (ref.null extern) (any.convert_extern))
)
```
```sh
$ wasm const.wat
const.wat:2.21-2.56: validation error: constant expression required
```

I believe the specification is the correct one (not entirely sure, though).

Additionally, I could not find a dedicated test for checking the constant-ness of instructions.
If that is missing, I think a separated `.wast` test should be added for this purpose.
(I attempted to name it `const.wast`, but that name is already used for testing `nt.const` instructions.)

rossberg:
Fix #1889.

@f52985, PTAL. Turns out we already have tests for the other const instructions, distributed over the respective test files.

### DIFF
--- a/interpreter/valid/valid.ml
+++ b/interpreter/valid/valid.ml
@@ -1012,8 +1012,9 @@ let is_const (c : context) (e : instr) =
   | Const _ | VecConst _
   | Binary (Value.I32 I32Op.(Add | Sub | Mul))
   | Binary (Value.I64 I64Op.(Add | Sub | Mul))
-  | RefNull _ | RefFunc _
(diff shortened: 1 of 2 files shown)


f52985:
Oh, I see. LGTM!

PR #1900, opened 2025-05-08
[spec] Describe backwards-compatibility and reserve 0xFF opcode

rossberg:
Triggered by WebAssembly/design#1539, this spells out the intended backwards-compatibility constraints we assume of future Wasm versions. I believe this captures what was discussed in that context.

### DIFF
--- a/document/core/appendix/changes.rst
+++ b/document/core/appendix/changes.rst
@@ -29,9 +29,10 @@ Concretely:
 2. All non-:ref:`trapping <trap>` :ref:`executions <exec>` of a valid program retain their behaviour with an equivalent set of possible :ref:`results <syntax-result>` (or a non-empty subset).
 
   .. note::
-    This allows previously illegal programs to become executable.
+    This allows previously malformed or invalid programs to become executable.
 
-    It also allows program executions that previously trapped to execute successfully.
+    It also allows program executions that previously trapped to execute successfully,
+    although the intention is to only exercise this where the possibility of such an extension has been previously noted.
 
     And it allows reducing the set of observable behaviours of a program execution,
     e.g., by reducing non-determinism.
--- a/document/core/appendix/changes.rst
+++ b/document/core/appendix/changes.rst
@@ -7,6 +7,39 @@ Change History
 Since the original release 1.0 of the WebAssembly specification, a number of proposals for extensions have been integrated.
 The following sections provide an overview of what has changed.
 
+All present and future versions of WebAssembly are intended to be *backwards-compatible* with all previous versions.
+Concretely:
+
+1. All syntactically :ref:`well-formed <binary>` and :ref:`valid <valid>` modules remain well-formed and valid with an equivalent :ref:`module type <syntax-moduletype>` (or a subtype).
+
+  .. note::
+     This allows previously malformed or invalid modules to become legal,
+     e.g., by adding new features or by relaxing typing rules.
+
+     It also allows reclassifying previously malformed modules as well-formed but invalid,
+     or vice versa.
+
+     And it allows refining the typing of :ref:`imports <syntax-import>` and :ref:`exports <syntax-export>`,
+     such that previously unlinkable modules become linkable.
+
+2. All non-:ref:`trapping <trap>` :ref:`executions <exec>` of a valid program retain their behaviour with an equivalent set of possible :ref:`results <syntax-result>` (or a non-empty subset).
+
+  .. note::
+    This allows previously illegal programs to become executable.
+
+    It also allows program executions that previously trapped to execute successfully.
+
+    And it allows reducing the set of observable behaviours of a program execution,
+    e.g., by reducing non-determinism.
+
+    In a program linking prior modules with modules using new features,
+    a prior module may encounter new behaviours,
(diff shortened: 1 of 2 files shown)


document/core/appendix/changes.rst
```diff
@@ -7,6 +7,39 @@ Change History
...
+
+     And it allows refining the typing of :ref:`imports <syntax-import>` and :ref:`exports <syntax-export>`,
+     such that previously unlinkable modules become linkable.
+
+2. All non-:ref:`trapping <trap>` :ref:`executions <exec>` of a valid program retain their behaviour with an equivalent set of possible :ref:`results <syntax-result>` (or a non-empty subset).
+
+  .. note::
+    This allows previously illegal programs to become executable.
+
+    It also allows program executions that previously trapped to execute successfully.
```
titzer, line 30: I'm not sure about this point. Have we utilized this to extend semantics to date?

One reason to avoid that is that trapping is observable in an embedding context, and both Wasm programs and embedded environments can use trapping operations as security checks; i.e. a program won't continue past a divide by zero or and out-of-bounds memory access and continue executing in an unanticipated state.

document/core/appendix/changes.rst
```diff
@@ -7,6 +7,39 @@ Change History
...
+
+    And it allows reducing the set of observable behaviours of a program execution,
+    e.g., by reducing non-determinism.
+
+    In a program linking prior modules with modules using new features,
+    a prior module may encounter new behaviours,
+    e.g., new forms of control flow or side effects when calling into a latter module.
+
+In addition, the :ref:`instruction opcode <binary-instr>` :math:`\hex{FF}` is considered reserved for custom use and will never be allocated to represent an instruction or instruction prefix.
```
titzer, line 39: Should we say "future versions of the spec will not allocate `0xFF` to a valid opcode"?

rossberg: Yes, reworded.

tlively (commented):
This all sounds right to me. Do we define subtyping on module types?

keithw:
In the added text, "All syntactically well-formed and valid modules remain well-formed and valid...", the hyperlink on "well-formed" goes to the binary chapter, but the language could be read as applying to both formats.

If it's intended that the text format will also be backwards-compatible, that would be great, but it would be a change from some post-1.0 cases (I think bulk memory) where compatibility was intentionally broken. Or if the text format isn't part of this, it could say something like "**In the binary format**, all syntactically well-formed..."?

tlively:
Do you have more details about the backward-incompatible text format change? I don't remember that.

keithw:
> Do you have more details about the backward-incompatible text format change? I don't remember that.

Sure, e.g. this was a valid module in Wasm 1.0 and became malformed after bulk-memory:

```wasm
(memory $x 0)
(data $x (i32.const 0))
(data $x (i32.const 0))
```

And, admittedly pre-1.0 (but still in a lot of online documentation), this was a valid module as late as 76d26bbc70c51729536243bbd7c64241a6d993d3 but is now malformed:

```wasm
(table 0 anyfunc)
```

tlively:
Thanks, I hadn't known about that bulk memory change. Here's the issue about it, in case anyone wants to see the original discussion: https://github.com/WebAssembly/bulk-memory-operations/issues/122. I hope that if we found a similar issue today, we would spend a little more time thinking about it 😅

I would be happy to document that we will maintain backward compatibility of the text format going forward.

rossberg:
@keithw, I clarified that this applies to both formats, despite us having been a tiny bit more loose with the text format in the past. (Though I would almost tend to define away this one away as an early bug fix. :) )

@tlively, we do not formally define subtyping on module types as such, only on individual external types. But this formulation is merely intended to give an intuition. It necessarily is a bit informal, since it is relating different language versions, and the definition of types and subtyping itself can change (and has changed) between versions.

document/core/appendix/changes.rst
sunfishcode: It seems useful to at least guarantee (if we can make such guarantees) that at least some traps will never be lifted, at the very least for the `unreachable` instruction.

Would it be feasible to include this sense of "cases where we previously documented that certain trapping cases might have legal behaviour in the future" in the text here, and limit the guarantee to that?

rossberg: Added a qualification along these lines.

document/core/appendix/changes.rst
```diff
@@ -7,6 +7,43 @@ Change History
...
+     such that previously unlinkable modules become linkable.
+
+     Historically, minor breaking changes to the *text format* have been allowed
+     that turned previously possible valid modules invalid,
+     as long as they were unlikely to occur in practice.
+
+2. All non-:ref:`trapping <trap>` :ref:`executions <exec>` of a valid program retain their behaviour with an equivalent set of possible :ref:`results <syntax-result>` (or a non-empty subset).
+
+  .. note::
+    This allows previously illegal programs to become executable.
```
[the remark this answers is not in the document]
rossberg: Malformed or invalid. Expanded.

rossberg:
Merging for now, please open issue in case other concerns should come up.

PR #1903, opened 2025-05-09
[interpreter] Support wrapping fancy function types in JS translation

### LINKED ISSUE #1902: [interpreter] Support JS test wrappers for calls with non-flat function types
Here is a test case written in `wast` format:

```wast
;; sample.wast
(module
  (type $t (sub (func (result f32))))
  (func (export "f") (type $t)
    (f32.const 0)
  )
)
(assert_return (invoke "f") (f32.const 0))
```

This defines a function with a non-final float (`f32`) as its return type.

When this `wast` file is converted into a JavaScript test case using the `./wasm` binary (built from the `wasm-3.0` branch) with the following command:

```
./wasm sample.wast -o sample.js
```

and then executed using a JavaScript engine such as V8:

```
./d8 sample.js
```

it results in the following unexpected `LinkError`:

```sh
LinkError: WebAssembly.Instance(): Import #0 "module" "f": imported function does not match the expected type
  return new WebAssembly.Instance(mod, imports);
         ^
LinkError: WebAssembly.Instance(): Import #0 "module" "f": imported function does not match the expected type
    at instance (sample.js:65:10)
    at sample.js:209:16
    at run (sample.js:82:3)
    at sample.js:209:1
```

I'm not sure whether the issue lies in the generated JavaScript test case or the behavior of the browser engines.
However, since the same error occurs across all major engines (V8, SpiderMonkey, and JavaScriptCore),
it seems likely that the issue originates from the JS code generator (harness).

I'm not very familiar with this area—could someone take a look?

rossberg:
@f52985, PTAL. Fixes #1902.

### DIFF
--- a/interpreter/script/js.ml
+++ b/interpreter/script/js.ml
@@ -276,6 +276,92 @@ let lookup_export (env : env) x_opt name at =
       string_of_name name ^ "\" within module isntance"))
 
 
+(* Transitively unsubstitute deftype into list of unrolled recursive types *)
+
+let rec statify_list f rts = function
+  | [] -> rts, []
+  | x::xs ->
+    let rts', x' = f rts x in
+    let rts'', xs' = statify_list f rts' xs in
+    rts'', x'::xs'
+
+let rec statify_heap_type rts = function
+  | DefHT dt ->
+    let rts', i = statify_def_type rts dt in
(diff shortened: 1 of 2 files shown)


f52985:
LGTM!

PR #1906, opened 2025-05-13
Wasm 3.0+relaxed.fix

rossberg:
@dtig, this implements the relaxation for relaxed_trunk we have converged on, PTAL.

I also removed all core tests that became obsolete/incorrect with this change. Unfortunately, it turns out that there are none left for relaxed_trunc, which is a bit concerning.

In fact, I then had a look at the tests for the other Relaxed SIMD instructions, and almost all of them only test a few edge cases. That is, with the exception of 5 assertions about `relaxed_dot`, we seem to have no tests for the “regular” behaviour of these instructions at all! That’s not good. Could somebody work on creating a suitable suite of basic tests? ;)


rossberg:
@dtig, friendly ping.

PR #1933, opened 2025-06-22
[spec] Add missing 'addrtype' in memory type validation

taichimaeda:
Found a potential typo!

Before:
Screenshot 2025-06-22 at 04 07 30

After:
Screenshot 2025-06-22 at 04 08 27


rossberg (approved):
Thanks!

PR #1940, opened 2025-06-24
[interpreter/spec] Consistent nullability for br_on_non_null

### LINKED ISSUE #1938: [spec] Disagreement between the textual description and the formule rule for br_on_non_null's validation
The validation rule for [`br_on_non_null`](https://webassembly.github.io/spec/versions/core/WebAssembly-3.0-draft.pdf#subsubsection*.195) states in the textual format that the last type of the result type corresponding to the given label should be `ref null? ht`. However, in the formal rule, it can't be null and is forced to be `ref ht`.

None of the test I saw used it with a label of type `ref null? ht` (it would be useless actually).

The reference validator seems to accept `ref null ht`, as the [test done](https://github.com/WebAssembly/spec/blob/7c48a0b01884d028432519d08f8fa2ef70db4c85/interpreter/valid/valid.ml#L470) is a matching and `ref ht <= ref null ht`.

I can do a PR where the formal rule is modified if that can help.

rossberg:
Fix #1938.

### DIFF
--- a/document/core/valid/instructions.rst
+++ b/document/core/valid/instructions.rst
@@ -2317,7 +2317,7 @@ Control Instructions
 
 .. math::
    \frac{
-     C.\CLABELS[l] = [t^\ast~(\REF~\X{ht})]
+     C.\CLABELS[l] = [t^\ast~(\REF~\NULL^?~\X{ht})]
    }{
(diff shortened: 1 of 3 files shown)


4y8:
Seems perfect to me.

PR #1941, opened 2025-06-25
[spec] Fixed typos regarding blocktypes

4y8:
The link "instruction type" in the rule for validation of `loop` pointed to "function type". There was a subscript `x*` for the instruction type of `loop`. However, the validation rule for block types specifies that the indices are always empty and they are omitted for `if`, `block` and `try_table`. The rule for `try_table` validated a blocktype to a function type instead of an instruction type.


rossberg (approved):
Thanks!

PR #1948, opened 2025-07-16
Updated the instruction index for memory64

### LINKED ISSUE #1947: [spec] Incoherent type for memory.grow
In the [validation rules](https://wasm-dsl.github.io/spectec/core/valid/instructions.html#valid-memory-grow), `memory.grow` has type `at -> at` where `at` is an [address type](https://wasm-dsl.github.io/spectec/core/syntax/types.html#syntax-addrtype) which can be either `i32` or `i64`. In the [index of instructions](https://wasm-dsl.github.io/spectec/core/appendix/index-instructions.html), it has type `i32 -> i32`.

I can make a PR if that helps.

4y8:
The instruction index didn't take into account memory64 and still put `i32` in place of address types. Should fix #1947


rossberg (commented):
Thanks! Can you adjust the table instructions, too?

4y8:
Oh the toolchain I am currently developping doesn't support tables yet so I didn't check that. I'll fix that now

4y8:
> Can you adjust the table instructions, too?

It should be good now

rossberg (approved):
Thanks you!

PR #1955, opened 2025-08-05
[simd] Add additional test input for i32x4.dot_i16x8_s

titzer:
Previously, input lanes were all identical, which hid bugs in implements, such as https://github.com/titzer/wizard-engine/pull/443


titzer:
@tlively An easy one, seems like @rossberg is on vacation.

tlively (approved):

PR #1974, opened 2025-09-19
Fix prose rendering of NotOp ContextKindE

Alan-Liang:
*not* the first non-value entry of the stack is a label →
the first non-value entry of the stack is *not* a label


rossberg:
@f52985, can you review this?

f52985:
Yes, looks good to me! :+1:

rossberg (approved):

PR #1982, opened 2025-09-23
[js-api] Fix `ToValueType` error case

Liedtke:
Any invalid / unsupported string should run into this, so this shouldn't be an assertion but an error?
I assume we don't need new tests as this should already be covered by https://github.com/WebAssembly/spec/blob/main/test/js-api/global/constructor.any.js#L82?

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1179,7 +1179,7 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     1. If |s| equals "v128", return [=v128=].
     1. If |s| equals "anyfunc", return [=funcref=].
     1. If |s| equals "externref", return [=externref=].
-    1. Assert: This step is not reached.
+    1. Throw a {{TypeError}} exception.
 </div>
 
 <div algorithm>


Ms2ger:
Can you explain in which cases you think this can be reached? Note that the WebIDL layer ensures that a value of type `ValueType` is always one of the strings in the enum.

Liedtke:
> Note that the WebIDL layer ensures that a value of type ValueType is always one of the strings in the enum.

Thanks for pointing that out, I wasn't aware of that.
https://wasm-dsl.github.io/spectec/js-api/#idl-index defines the `ValueType` enum which the [`WebAssembly.Global`](https://wasm-dsl.github.io/spectec/js-api/#globals) uses. 
Do I understand correctly, that any invalid value not specified by the IDL is always reported as a `TypeError` and that is defined once for WebIDL and then implicitly relied on by the wasm-js spec?

Ms2ger:
Indeed, when converting a javascript value to an enumeration type (https://webidl.spec.whatwg.org/#js-enumeration), any invalid values will cause an exception.

(Btw, you can use https://webassembly.github.io/spec/js-api again.)

Liedtke:
> Indeed, when converting a javascript value to an enumeration type (https://webidl.spec.whatwg.org/#js-enumeration), any invalid values will cause an exception.

Closing this PR as I was just missing context about how to read the spec, thank you for helping out!

> (Btw, you can use https://webassembly.github.io/spec/js-api again.)

I updated my bookmark, thanks for the hint. :)

Issue #1980, opened 2025-09-22
[js-api] Are Wasm-gc abstract heap types included in the JS API of e.g. WebAssembly.Global?


Liedtke:
The WebAssembly [JS-API spec](https://webassembly.github.io/spec/js-api/#globals) mentions the following types for globals:
```
enum ValueType {
  "i32",
  "i64",
  "f32",
  "f64",
  "v128",
  "externref",
  "anyfunc",
};
```
However, both V8 and SpiderMonkey support multiple more values like `"anyref"` or `"i31ref"`.
After some code-search in V8, I found [my commit](https://chromium-review.googlesource.com/c/v8/v8/+/3779911) which added it in 2022 when `anyref` was introduced for wasm-gc as a separate type hierarchy vs. the existing `externref`.
In later changes we added more of these new types like `i31ref` probably assuming some kind of consistency to the already present `anyref`.

It seems that JavaScriptCore doesn't accept these values.

The [wasm-gc JS API document](https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP-JS.md) does not list any extensions to these existing APIs.
So from a spec perspective, it looks like this is a bug in V8 and SpiderMonkey?
Was it a conscious decision to not add support for the new abstract heap types to the existing JS APIs?

tlively:
I don't think this was a conscious decision, and I do think it would make sense to add the other abstract heap types. Supporting non-nullable references and references to defined heap types would require much more design work, of course, but I don't think that's a good reason to hold off on making the obvious extensions to the current scheme.

Liedtke:
Notably I didn't find `"externref"` as a new value in the JS-API documented anywhere in the [typed function references proposal](https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md).

The proposals also don't seem to link a js-api spec with their proposed changes?
So while we probably shouldn't have added `anyref`, I'm wondering whether there was any documentation that advised us to add `externref` in the first place? 
I also couldn't find a [spec test](https://github.com/WebAssembly/spec/blob/main/test/js-api/global/constructor.any.js) for an externref-global defined via the JS API.

A few more notes:
1) A similar issue applies to tables where `new WebAssembly.Table({element: 'anyref', initial: 4, maximum: 4})` will just pass in V8 as well and produce a table of anyref. And yes, for the same assumed consistency, also tables of `i31ref` and other abstract heap types are possible.
[...]

PR #1983, opened 2025-09-23
[js-api] Add test case for externref global

Liedtke:
We should add some basic coverage for externref globals.
(We should extend this if https://github.com/WebAssembly/spec/issues/1980 results in adding more allowed types.)


Ms2ger (changes requested):

Liedtke:
Can this be merged?

Ms2ger (approved):
Thanks

PR #2009, opened 2025-10-09
[spec] GC objects no-op prevent extensions

### LINKED ISSUE #2008: gc-exotic-objects `[[PreventExtensions]]` should return `true`.
GC object are already non-extensible (`[[IsExtensible]]` returns `true`). Preventing extensions on a non-extensible object should be an no-op.

By returning `false` to `[[PreventExtensions]]`, any caller to `Object.preventExtensions()` would get a thrown `TypeError` even though the object is already non-extensible, which is surprising (even though it is not technically a violation of the object invariants).

To avoid this surprising behavior, WASM Garbage Collected Objects should return `true` for `[[PreventExtensions]]`.

mhofman:
Fixes https://github.com/WebAssembly/spec/issues/2008

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1600,7 +1600,7 @@ The internal methods of an [=Exported GC Object=] use the following implementati
 <div algorithm>
   The <dfn>\[[PreventExtensions]] internal method of an Exported GC Object</dfn> <var ignore>O</var> takes no arguments and returns a boolean. It performs the following steps when called:
 
-    1. Return false.
+    1. Return true.
 </div>
 
 <div algorithm>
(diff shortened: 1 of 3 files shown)


WebReflection:
LGTM but if we could add:

```js
test(() => {
  const struct = functions.makeStruct();
  const array = functions.makeArray();
  assert_true((() => { for (const _ in struct); return true; })());
  assert_true((() => { for (const _ in array); return true; })());
}, "for...in");
```

then we'd be more covered around the current state, imho.

tlively:
This change to intentionally-designed behavior should be discussed with the CG first. Let's keep further discussion on #2008.

PR #2015, opened 2025-10-15
[spectec] Remove interpreter test message

### LINKED ISSUE #1999: [spectec] Interpreter test expectations interfere with test suite additions
With SpecTec on the main repo, the test suite can now be updated by other than SpecTec devs. Unfortunately, each time that happens, it breaks the SpecTec interpreter's test expectations as currently set up, and it requires spurious changes like [this](https://github.com/WebAssembly/spec/commit/254730acb7e771517099a91e5c6e3f59865780ec) to fix after the fact.

@f52985, can we tweak the interpreter output such that the addition of new successful tests creates no diff? For example, just output the summary when it is not 100%.

f52985:
Fixes #1999 by removing interpreter test messages for success rate.


f52985:
Makes sense, fixed accordingly.

PR #2024, opened 2025-10-31
[test] Correct catch clause in type mismatch test

g-plane:
As the function's export name says, it should be `catch_ref` instead of `catch` in `try_table`, otherwise it will be similar to the previous `"catch"` test.

Here is after changed to `catch_ref`:

image


rossberg (approved):
Thanks!

PR #2031, opened 2025-11-20
[spectec] Implement IterPr in IL evaluator

rossberg:
@zilinc, please take a look. Unfortunately, there is no proper way of testing this in isolation.

@conrad-watt, FYI.


zilinc:
Thanks a ton @rossberg . I should be able to test it with the rest of my interpreter.

PR #2036, opened 2025-12-02
[spectec] Make definition of state access nicer

rossberg:
@f52985, in an attempt to make (the rendering of) these helpers a bit more readable, I introduced a trivial indirection through two auxiliary functions. Unfortunately, that completely breaks the interpreter, and I don't understand why. Can you please have a look?


rossberg:
@f52985, I just noticed that this older PR is still awaiting your attention as well. :)

f52985:
Oops, I didn’t even realize this PR existed. 😅

The issue is that the new functions `$sof` and `$store` are translated into bodies containing only a single instruction: `Return.`

This happens because of an assumption in the translator based on the specification:

Whenever a function returns `store`, it is assumed to be used in a *store-passing style*, for example:

```
-- if s1 = f(s)
-- if s2 = g(s1)
-- if s3 = h(s2)
...
```

In this pattern, each function call is isolated, and the return value of each call is bound to a single variable.

Based on this assumption, the IL2AL translator removes explicit store passing and uses the variable `s` as if it were the single variable used to access the global store.

However, the new functions `$sof` and `$store` do not follow this assumption. Instead, they are used in an *inline* style, where the function call appears as a subexpression within another expression, and removing the return statement brings the wrong result.

Because of this, I believe these functions require special handling at some point—either through a `hint` in the specification, during translation, or during AL interpretation.

My choice was to add a hardcoded case in the interpreter that treats `$store` specially, forcing it to always return the global store.

PR #2050, opened 2026-01-12
Update LICENSE for spectec and proposals directories

### LINKED ISSUE #2049: Lacks of license in some directories
The following subdirectories does not have a `LICENSE` file:
```
- spectec/
- proposals/
```

rossberg:
Fix #2049.

Are these adequate choices? WDYT @tlively, @dschuff, @conrad-watt?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -6,5 +6,5 @@ document/    - W3C Software and Document Notice and License
 spectec/     - Apache License 2.0
 interpreter/ - Apache License 2.0
 test/        - Apache License 2.0
-proposals/   - Creative Commons Attribution 4.0 International License
-papers/      - Creative Commons Attribution 4.0 International License
+proposals/   - Creative Commons Universal 1.0 International License (CC0)
+papers/      - Creative Commons Attribution 4.0 International License (CC BY 4.0)


tlively:
LGTM, but maybe let's run this by proposal champions at the CG meeting tomorrow.

eqrion:
The CCA 4.0 license (which is added to proposals/) is not compatible with the MPL. Practically speaking this would mean we'll need to be careful not to embed any text from a proposal overview or examples into Firefox. We can do that if we have to, but it would be nice to pick another license if possible.

For standards, our [preferred license is a dual CC0+OWFa license](https://wiki.mozilla.org/Standards/license). Other licenses are acceptable to us too, if one is proposed I can ask about it.

I am not a lawyer, the above is just relaying advice I was given internally.

rossberg:
Sounds fine to me, though tbh I've got no clue how the OWFa thing works in this context, its [actual text](https://www.openwebfoundation.org/the-agreements/the-owf-1-0-agreements-granted-claims/owfa-1-0) seems to require signing?

eqrion:
I'm not sure how the OWFa thing works either. It looks like it's just for patent policy, and we are already working within the W3C patent policy, so I don't think we'd need it. The wiki page I linked to says that CC0 is approved (for Mozilla) for use by itself without OWFa, so I'd guess that'd be the cleanest.

tlively:
Another option would be to apply the same W3C Software and Document Notice and License that we apply to the document/ directory. But CC0 is fine with me, too.

rossberg:
Okay, I switched to CC0 for the proposals directory.

PR #2054, opened 2026-01-16
[js-api] Fix result types in string builtins

tlively:
The string builtins that return strings should return non-nullable `(ref extern)` rather than nullable `externref`. This change reflects current implementation reality in both Chrome and FireFox.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1948,7 +1948,7 @@ The <dfn abstract-op lt="CharCodeAt">CharCodeAt(|string|, |index|)</dfn> abstrac
 
 <h4 id="js-string-cast">cast</h4>
 
-The |funcType| of this builtin is `(rec (type (func (param externref) (result externref)))).0`.
+The |funcType| of this builtin is `(rec (type (func (param externref) (result (ref extern))))).0`.
 
 <div algorithm="js-string-cast">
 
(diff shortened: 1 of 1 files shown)


jakobkummerow (approved):
LGTM.
FWIW, `(ref extern)` is what the proposal's [Overview](https://github.com/WebAssembly/js-string-builtins/blob/main/proposals/js-string-builtins/Overview.md#wasmjs-string-fromcharcode) has always said. Dunno why the final spec ended up diverging from that; I'm guessing that wasn't intentional.

tlively:
cc @rmahdav

sideshowbarker:
> Seems fine. I'd like @sideshowbarker's review of the echidna failures before merging, though

Our charter has expired, so we no longer have perms to (auto)publish to TR space. I’ve sent a request for an extension of the charter. If/when that gets OKed, the echidna failures should go away.

sideshowbarker:
> > Seems fine. I'd like @sideshowbarker's review of the echidna failures before merging, though
> 
> Our charter has expired, so we no longer have perms to (auto)publish to TR space. I’ve sent a request for an extension of the charter. If/when that gets OKed, the echidna failures should go away.

No more echidna failures. We got our charter extension — so it’s back again to working now. All green 🍀

Ms2ger:
Thanks!

PR #2055, opened 2026-01-17
[js-api] Test validation of string builtin types

tlively:
Validation should fail if string builtin imports do not have the expected types. Add a test for this validation.


tlively:
I validated locally (by applying this diff to WPT) that FireFox passes this new test. Chrome does not, however, because V8 currently checks the builtin params and results individually instead of checking the entire import type. It seems worth fixing this (and testing it) to avoid compatibility hazards.

Ms2ger (approved):

tlively:
cc @rmahdav

PR #2056, opened 2026-01-20
[spectec] Clean up IL semantics and elaborator

rossberg:
This PR cleans up and tightens SpecTec IL scoping and validation and revamps the frontend elaborator accordingly.

The relevant tweaks to the IL are as follows (see also the diff on il/ast.ml):

- Binders are renamed to quantifiers and share the same AST type with parameters.
- Dependent tuple types `(x1:t1,...,xN:tN)`now bind just variables, not expressions. Furthermore, these variables are properly locally scoped. That implies that they no longer show up in outer binder lists. They scope over field types to the right as expected.
- As a special case, when a tuple type occurs as the parameter of a variant or record constructor, its names also scope over any premises of that arm. For example, in `syntax t = C(a : t, b : u(a)) -- if $p(a, b)`, the variables `a` and `b` scope over the premise.
- Consequently, it is no longer necessary that the binders list of variant and record constructors scopes over their parameter type and can instead be treated solely as an existential quantifier for the free variables in the premises. As such, it moves to the right of the parameter and can depend on it, as in `syntax t = C(a : t, b : u(a)) {c : v(b)} -- if $f(a, b) = c`.
- Iteration variables (both `i` and `j` in `e^(i<n){j <- k}` are now properly locally scoped (this was a bit ambiguous before). That implies that they also no longer show up in outer binder lists.

The net effect of these changes is that all scoping now is much more standard and there no longer is any duplication between "binders" and parameters. As a concrete example, this e.g. is how the elaboration of the type `name` simplifies:

```diff
 syntax name =
-  | `%`{`char*` : char*}(char*{char <- `char*`} : char*)
+  | `%`(`char*` : char*)
     -- if (|$utf8(char*{char <- `char*`})| < (2 ^ 32))
```
There are only very few cases left where quantifiers on variants actually still show up (for example, the `SUBNORM` constructor for floats). Note that quantifiers remain heavily used for function definitions and their argument patterns.

- In addition to the above, the internal representation of mixfix operators changes from a list of atoms to a tree of atoms, in order to preserve knowledge about precedences (needed in the elaborator). Backends that do not construct/deconstruct mixops should be unaffected.

Currently there still is a test failure on the AL interpreter.

### DIFF
--- a/.github/workflows/ci-interpreter.yml
+++ b/.github/workflows/ci-interpreter.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           ocaml-compiler: 5.04.x
       - name: Setup OCaml tools
-        run: opam install --yes ocamlfind.1.9.5 js_of_ocaml.4.0.0 js_of_ocaml-ppx.4.0.0
+        run: opam install --yes ocamlfind js_of_ocaml js_of_ocaml-ppx
       - name: Build interpreter
         run: cd interpreter && opam exec make
       # Neither V8 nor SpiderMonkey can currently handle all 3.0 tests, so we disable checking JS translation for now.


f52985:
Fixed the issues on AL interpreter.

The issues were:

**1. FUNC comptype representation**
For the `comptype` `FUNC rt1 -> rt2`, the AL `CaseV` representation is now `->(rt1, rt2)` instead of `FUNC(rt1, rt2)`, since the `Mixop.head` of the mixop returns "->". Previously, some components relied on the hardcoded string "FUNC", (e.g., `backend-interpreter/construct.ml`). All uses are now standardized to use "->" as the constructor name for this `CaseV`.

**2. IterE with indexed ListN**
The interpreter backend previously assumed that if `IterE` uses `ListN` with an index variable, the bound variables (`xes`) must also include that index variable. Since this assumption no longer holds, the translation and the interpretation logic were fixed to correctly handle such case.

rossberg:
@f52985, thanks a lot for looking into this! I'll land this PR then.

I have two follow-up questions/requests:

1. One of the errors I saw previously seems related to your first issue:
   ```
    $subst_comptype: on expr `(FUNC t_1*{t_1 <- t_1*} -> t_2*{t_2 <- t_2*})` invalid assignment: cannot be an assignment target for (FUNC [] []) (interpreting  Let (FUNC t_1*{t_1 <- t_1*} -> t_2*{t_2 <- t_2*}) = comptype at ../../../_specification/wasm-3.0/1.2-syntax.types.spectec:394.21-396.40)
   ```
   However, this looks as if lhs and rhs have the same constructor. Is the difference hiding behind the pretty-printing? If so, can we tweak it to make such discrepancies visible?

2. Can we change the AL test runner to abort a test file after the first failure? Having to wade through dozens of pages of follow-up "option is none" exceptions to find the actual error becomes a bit tedious after a while. :)

f52985:
1. I believe that the previous error messages containing the constructor name `FUNC` was not because of pretty-printing, but because they were from the manually constructed `CaseV`s, and indeed contained the contructor name `FUNC`. Since all of the hardcoded names are now fixed to the normalized name `FUNC`, there should not be such error message now.

2. Yes, now the AL interpreter would abort on the first test failure.

PR #2063, opened 2026-01-30
[spec] add missing rules for HANDLER

zilinc:


rossberg (commented):
Looks good, but I just noticed another problem with these rules: they assume the frame is unchanged, which of course makes them incompatible with local.set.

While here, do you mind fixing this by adding the state's `f` component to each rule, with `f'` on the righthand-side?

zilinc:
Reordered the rules and squashed the commits.

rossberg:
Thanks! There are some merge conflicts with the test expectations. I think you need to merge main and rerun tests.

(Aside: No need to squash, that happens anyway when merging the PR. To the contrary, please never force-push to a branch under review, as that breaks some of GH's review functionality.)

zilinc:
> Thanks! There are some merge conflicts with the test expectations. I think you need to merge main and rerun tests.
> 
> (Aside: No need to squash, that happens anyway when merging the PR. To the contrary, please never force-push to a branch under review, as that breaks some of GH's review functionality.)

Okay I've done that. Sorry about messing up the workflow. Hope this time it's the right way.

rossberg (approved):

rossberg:
Thanks! No worries, it's a super common mistake, and unfortunately, Github does not offer any means for preventing it, despite popular request. In general, it's almost always better to use git merge instead of rebase.

PR #2064, opened 2026-01-30
Create specification/wasm-latest

tlively:
To give proposal repos a consistent location to make changes to the
SpecTec source and to simplify merging those proposals back into the
spec repo once they are completed. The new directory is a copy of the
wasm-3.0 directory in its current state, created via a sequence of git
mv and merge operations to preserve the blame history in both
directories.

Also add a new CI script ensuring that the contents of wasm-latest match
the contents of the highest versioned wasm-X.Y directory. This will
ensure that bug fixes are applied in both directories.


rossberg:
~~Oh, and I think we also want to change which directory is used for the document build. That should always be latest, to make sense on forks.~~

Oops, sorry, only saw the follow-up now. :)

tlively:
> * It would be great if the diff check is also invoked locally, e.g., as part of the `check` target in the specification dir's Makefile. The shell script could live there as well.

But we wouldn't want to run this on forks, right? Since a local development could have both WebAssembly/spec and proposal repos as upstreams, I don't think there's a good way to automatically detect whether this check should run or not locally.

tlively:
Well the CI script works :) Now I have to figure out how to pull the latest changes to wasm-3.0 into this cursed merge commit structure...

tlively:
Looks like the fancy moves and merges to preserve history didn't survive the PR process :( We could force push this branch to main, but it might not be worth it. Wdyt?

rossberg:
No real surprise, I suppose... Yeah, please push it manually. Especially since there are other PRs in flight that touch the spec, so we already need the history now, I think.

PR #2065, opened 2026-01-30
Use wasm-latest instead of wasm-3.0 in tests and docs

tlively:
This will reduce the number of changes necessary when we cut new
releases in the future.

### DIFF
--- a/specification/Makefile
+++ b/specification/Makefile
@@ -31,7 +31,7 @@ check-3.0: $(EXE)
 check-latest: $(EXE)
 	$(EXE) wasm-latest/*
 	@echo Wasm latest spec checked OK.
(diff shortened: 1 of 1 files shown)


specification/Makefile
```diff
@@ -28,25 +28,30 @@ check-3.0: $(EXE)
 	@echo Wasm 3.0 spec checked OK.
 	@echo =========================
 
-check-all: check-1.0 check-2.0 check-3.0
+check-latest: $(EXE)
+	$(EXE) wasm-latest/*
+	@echo Wasm latest spec checked OK.
+	@echo =========================
```
rossberg: Nit: adjust underlines (here and below).

PR #2067, opened 2026-02-01
[spec] Restore explicit `reftype` in Element section

### LINKED ISSUE #2066: Backwards-compatibility question regarding dropped `reftype` in Element section
Hello! I'm currently implementing support for v3 in my [decoder](https://github.com/agis/wadec) and ran into a failure while implementing the changes for the `Element` section.

v2 defined one of the possible variants of the `elem` item in the [`Element` section](https://www.w3.org/TR/wasm-core-2/#element-section%E2%91%A0)):

```
elem ::= ...
         | 6:u32 𝑥:tableidx 𝑒:expr et : reftype el *:vec(expr) ⇒ {type 𝑒𝑡, init el *, mode active {table 𝑥, offset 𝑒}}
         ...
```

If I understand correctly, as per the v2 spec, an explicit `reftype` is expected. Essentially after `e` either `0x70` or `0x6F` is expected.

However, v3 changed the above definition to omit the `reftype`:

```
elem ::= ...
         | 6:u32 𝑥:tableidx 𝑒o:expr 𝑒*:list(expr) ⇒ elem (ref null func) 𝑒* (active 𝑥 𝑒o)
         ...
```

This means that now, a module that was previously valid in v2 becomes invalid in v3, given that after `𝑒o`, the byte `0x70` (or `0x6F`) would be interpreted as the `e` list's length.

Is this not a backwards-incompatible change? I'm probably missing something and it's not, given that both `wasm-tools` and `wasm-validate` accept the above module. But nevertheless I wanted to understand what actually a conformant decoder has to do in this case.

In case it helps, the test that failed in my suite uses the following module: https://github.com/agis/wadec/blob/85d9f6dc1035978263303f4e5e81b0bef25a6072/tests/fixtures/element_section_all_alts.wasm.

Thanks in advance!

agis:
Version 2 of the specification parsed an explicit `reftype` in the `elem` item, inside the Element section:

    elem ::= ...
         | 6:u32 𝑥:tableidx 𝑒:expr et:reftype el *:vec(expr) ⇒ {type 𝑒𝑡, init el *, mode active {table 𝑥, offset 𝑒}}
         ...

The `reftype` item was removed in version 3:

    elem ::= ...
         | 6:u32 𝑥:tableidx 𝑒o:expr 𝑒*:list(expr) ⇒ elem (ref null func) 𝑒* (active 𝑥 𝑒o)
         ...

However, this is a breaking change, given that v3-conformant decoders would treat modules that had this `reftype` as invalid.

Fixes #2066

### DIFF
--- a/specification/wasm-3.0/5.4-binary.modules.spectec
+++ b/specification/wasm-3.0/5.4-binary.modules.spectec
@@ -110,8 +110,8 @@ grammar Belem : elem =
       ELEM (REF NULL FUNC) e* (ACTIVE 0 e_O)
   | 5:Bu32 rt:Breftype e*:Blist(Bexpr) =>
       ELEM rt e* PASSIVE
-  | 6:Bu32 x:Btableidx e_O:Bexpr e*:Blist(Bexpr) =>
-      ELEM (REF NULL FUNC) e* (ACTIVE x e_O)
+  | 6:Bu32 x:Btableidx e_O:Bexpr rt:Breftype e*:Blist(Bexpr) =>
+      ELEM rt e* (ACTIVE x e_O)
   | 7:Bu32 rt:Breftype e*:Blist(Bexpr) =>
       ELEM rt e* DECLARE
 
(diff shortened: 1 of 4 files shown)


agis:
> Yes, that was a bug. Thanks for the fix!
> 
> To land this, you will probably have to update the SpecTec test expectations as well (in `spectec/`, run `make clean` and then `make test; make testpromote` until there's no diff anymore).

Not sure, but it should be good now.

PR #2068, opened 2026-02-02
Ensure wasm-latest is correct in CI

tlively:
In the upstream spec repo, the contents of wasm-latest must always be
the same as the highest numbered wasm-X.Y spec directory. Add a CI
script to enforce this.


tlively:
Split out from #2064 so that only the minimal change had to be pushed manually.

PR #2070, opened 2026-02-03
Use wasm-latest instead of wasm-3.0 in tests and docs

tlively:
This will reduce the number of changes necessary when we cut new releases in the future.


tlively:
This is a copy of #2065 after some branch shenanigans caused that one to be closed.

PR #2071, opened 2026-02-04
[spec] Remove note restricting `memidx <= 0` in text format

zsakowitz:
We could alternatively replace the "no memidx>0 are allowed" note with "it used to be the case that memidx>0 were not allowed in the text format", so that anybody who visits is made aware of the change. Not sure if there's a preference for one or the other in the general spec.


rossberg (approved):
Thanks!

PR #2081, opened 2026-02-12
[test] Remove unused 'cont-inner' function from loop.wast

Sainan:


rossberg (commented):
Not sure what exactly this was supposed to test, but it may have been something about validation. It probably isn't called because it doesn't actually terminate. Either way, I see no harm in keeping it?

PR #2082, opened 2026-02-12
[test] Check that backwards branches trim the stack

### LINKED ISSUE #2080: Tests don't check that backwards branches implicitly trim the stack
Kinda self-reporting here that I have a broken implementation, but it seems loop.wast, br.wast, br_if.wast, & labels.wast pass regardless of this:
```wat
(module
  (func (export "test") (param) (result i32)
    (local $i i32)
    i32.const 1
    (loop $loop (param) (result)
      (local.set $i (i32.add (local.get $i) (i32.const 1)))
      i32.const 2
      (br_if $loop (i32.lt_s (local.get $i) (i32.const 10)))
      drop
      )
    )
  )
```
The expected return value of this function is `1`, but an incorrect implementation that does not implement the implicit dropping/trimming of the stack would produce `2`.

Sainan:
Wasn't really sure where to put it or what to call it, but I hope this is fine.

Closes #2080

### DIFF
--- a/test/core/loop.wast
+++ b/test/core/loop.wast
@@ -425,6 +425,17 @@
     )
     (drop) (drop) (drop)
   )
+
+  (func (export "br-implicit-drop") (result i32)
+    (local $i i32)
+    (i32.const 1)
+    (loop $loop
+      (local.set $i (i32.add (local.get $i) (i32.const 1)))
+      (i32.const 2)
+      (br_if $loop (i32.lt_s (local.get $i) (i32.const 10)))
+      (drop)
+    )
+  )
 )
 
 (assert_return (invoke "empty"))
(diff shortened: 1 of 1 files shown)


tlively:
test/core/br_if.wast would probably be a better location since this is part of the behavior of br_if more than loops.

Sainan:
Alright, moved it there.

Sainan:
I did try writing this using `br` initially but I kept producing non-terminating code. I do suspect `br_if` is unavoidable for this test.

rossberg:
You'll need a separate branch for terminating the loop. It's probably not worht it, but then I think it makes more sense to keep the test in loop.wast (sorry for the churn :) ).

Sainan:
Okay, pointed the branch at a1b91d44d8ee234afae1600aa23e1faa957f3b66 again ☺️

PR #2084, opened 2026-02-13
[js-api] Use [AllowResizable] AllowSharedBufferSource for bytecode

### LINKED ISSUE #1907: Should JS-API compile methods accept shared buffers?
The JS-API for compiling modules takes a `BufferSource` for accepting bytecode [1]. This does not allow shared array buffers or shared array buffer views [2]. Firefox until Fx138 actually supported shared array buffers/views, but I removed this to be compliant with the spec here [3]. This also made some optimizations easier as you can assume there are no racing writes to the buffer.

We have since gotten bug reports that while Chrome doesn't support shared array buffers (which is why I thought it was a safe change to make in Firefox), it does support views of shared array buffers. Safari supports boths shared array buffers and view of shared array buffers.

[1] https://webassembly.github.io/spec/js-api/index.html#modules
[2] https://webidl.spec.whatwg.org/#BufferSource
[3] https://bugzilla.mozilla.org/show_bug.cgi?id=1931407

eqrion:
Fixes #1907.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -372,11 +372,11 @@ dictionary WebAssemblyCompileOptions {
 
 [Exposed=*]
 namespace WebAssembly {
-    boolean validate(BufferSource bytes, optional WebAssemblyCompileOptions options = {});
-    Promise&lt;Module> compile(BufferSource bytes, optional WebAssemblyCompileOptions options = {});
+    boolean validate([AllowResizable] AllowSharedBufferSource bytes, optional WebAssemblyCompileOptions options = {});
+    Promise&lt;Module> compile([AllowResizable] AllowSharedBufferSource bytes, optional WebAssemblyCompileOptions options = {});
 
     Promise&lt;WebAssemblyInstantiatedSource> instantiate(
-        BufferSource bytes, optional object importObject, optional WebAssemblyCompileOptions options = {});
+        [AllowResizable] AllowSharedBufferSource bytes, optional object importObject, optional WebAssemblyCompileOptions options = {});
 
     Promise&lt;Instance> instantiate(
         Module moduleObject, optional object importObject);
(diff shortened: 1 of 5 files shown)


eqrion:
This should have updated WPTs for this. I always forget this, but is it canonical to land WPTs here or in the upstream WPT repository?

Ms2ger:
> This should have updated WPTs for this. I always forget this, but is it canonical to land WPTs here or in the upstream WPT repository?

Both, I'm afraid. I intended to have something (semi-)automatic, but never found the time to set it up. (It's complicated by the proposals being in separate repositories.)

eqrion:
I've now added some basic tests. I realized my original commit had some largish generated build files in it, so I removed that and force pushed to clear it from the git history.

eqrion:
review-ping @Ms2ger. Don't want to start approving my own PRs quite yet :)

eqrion:
@jakobkummerow Just double checking, are you okay with accepting resizable array buffers?
@kmiller68 FYI. I believe you already support SAB. Are you okay with supporting RAB/GSAB here?

jakobkummerow:
Yeah, no objections.

eqrion:
I'm going to go ahead an merge this then. Thanks!

saschanaz:
Creating SharedArrayBuffer requires extra header to the page, right? Otherwise the add tests will probably fail by not being able to create SAB?

And the tests are [not available in WPT repo yet](https://wpt.fyi/results/wasm/jsapi/constructor/compile.any.html?label=master&label=experimental&aligned), do any of you know when it syncs?

eqrion:
> Creating SharedArrayBuffer requires extra header to the page, right? Otherwise the add tests will probably fail by not being able to create SAB?
> 

I believe you can create a SAB even without any headers set, but you can't postMessage it and it's not exposed on the global directly. These tests might need a tweak to get working, I did run them in the jsshell harness in SpiderMonkey and so I might have missed this by not running the normal WPT harness.

> And the tests are [not available in WPT repo yet](https://wpt.fyi/results/wasm/jsapi/constructor/compile.any.html?label=master&label=experimental&aligned), do any of you know when it syncs?

There is no automatic sync for the JS-API and I haven't manually synced this over yet. https://github.com/WebAssembly/spec/issues/2130 is a discussion about the best way to add an auto sync feature, and I hope to work on that soon.

PR #2089, opened 2026-02-19
Rework JS API spec implementation limits

### LINKED ISSUE #1863: [js-api] Ambiguous and inconsistent handling of implementation limits

bvisness:
Resolves #1863 by reworking how implementation limits are organized, and drawing a clear distinction between compile-time and runtime limits.

I have currently made the decision to throw a RangeError when instantiating a module that trips a runtime resource limit, e.g. instantiating a module with a table whose initial size exceeds the implementation limit. This was already reflected in some tests and mirrors what happens if constructing an oversize resource from JS, e.g. new WebAssembly.Table.

I've left a couple inline comments to highlight the impact this spec will have on engines.

A couple other things to note:
- There was no defined implementation limit for the maximum number of element segments, despite this being enforced in the tests.
- The tests for limits were rather thin and have been fleshed out. Memory and table tests are grouped into the `memory` and `table` folders and memory64 limits are only checked in test files with `-memory64` in the name, following what seems to be an existing pattern. However, I honestly have no idea where we're supposed to have tests for this stuff, since the WPT tests in this repo and in the main WPT repo are wildly out of sync, so please tell me if I should organize things differently.
- There was a random Promise.resolve() in the limits tests that seems like it would just skip the actually important part of the tests. I removed it and things seem all right, but I can't fathom why it was there in the first place, and I fear some test behavior might change.

**Important:** Currently _no_ major engines pass all of these tests. SpiderMonkey throws RuntimeError where I think we should now throw RangeError, V8 rejects tables at compile time when it should reject them at instantiation time, etc. This suite of updates attempts to resolve those discrepancies, but it's less clear than usual which behavior we want.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -100,6 +100,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
         text: String.prototype.substring; url: sec-string.prototype.substring
         text: Array; url: sec-array-exotic-objects
         text: BigInt; url: sec-ecmascript-language-types-bigint-type
+        text: safe integer; url: #safe-integer
 urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: dfn
     text: embedding interface; url: appending/embedding.html
     text: scope; url: intro/introduction.html#scope
@@ -399,6 +400,7 @@ Note:
   To <dfn>compile a WebAssembly module</dfn> from source bytes |bytes|, perform the following steps:
     1. Let |module| be [=module_decode=](|bytes|). If |module| is [=error=], return [=error=].
     1. If [=module_validate=](|module|) is [=error=], return [=error=].
+    1. If any <a href="#limits-compile-time">compile-time limits</a> are exceeded, return [=error=].
     1. Return |module|.
 </div>
 
@@ -608,6 +610,7 @@ The verification of WebAssembly type requirements is deferred to the
     1. Let |result| be [=module_instantiate=](|store|, |module|, |imports|).
     1. If |result| is [=error=], throw an appropriate exception type:
         * A {{LinkError}} exception for most cases which occur during linking.
+        * If a <a href="#limits-runtime">runtime implementation limit</a> is exceeded, throw a {{RangeError}}.
         * If the error came when running the start function, throw a {{RuntimeError}} for most errors which occur from WebAssembly, or the error object propagated from inner ECMAScript code.
         * Another error type if appropriate, for example an out-of-memory exception, as documented in <a href="#errors">the WebAssembly error mapping</a>.
     1. Let (|store|, |instance|) be |result|.
@@ -879,6 +882,8 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     1. If |descriptor|["maximum"] [=map/exists=], let |maximum| be [=?=] [=AddressValueToU64=](|descriptor|["maximum"], |addrtype|); otherwise, let |maximum| be empty.
     1. Let |memtype| be [=memory type=] |addrtype| { **min** |initial|, **max** |maximum| }.
     1. If |memtype| is not [=valid memtype|valid=], throw a {{RangeError}} exception.
+    1. If |maximum| is not empty:
+        1. If |addrtype| is "i64" and |maximum| exceeds the [=memory type size limit=], throw a {{RangeError}} exception.
     1. Let |store| be the [=surrounding agent=]'s [=associated store=].
     1. Let (|store|, |memaddr|) be [=mem_alloc=](|store|, |memtype|). If allocation fails, throw a {{RangeError}} exception.
     1. Set the [=surrounding agent=]'s [=associated store=] to |store|.
@@ -908,6 +913,8 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     1. Let |store| be the [=surrounding agent=]'s [=associated store=].
     1. Let |ret| be the [=mem_size=](|store|, |memaddr|).
     1. Let |store| be [=mem_grow=](|store|, |memaddr|, |delta|).
+
+        Note: This should check the <a href="#limits-runtime">runtime implementation limits</a>.
     1. If |store| is [=error=], throw a {{RangeError}} exception.
     1. Set the [=surrounding agent=]'s [=associated store=] to |store|.
     1. [=Refresh the memory buffer=] of |memaddr|.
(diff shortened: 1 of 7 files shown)


rossberg (commented):
This all looks good to me. Have you compared by any chance whether the new limits introduced here are also listed in the core spec's Appendix? I wouldn't be surprised if that also has some gaps regarding newer features.

bvisness:
I don't think I've introduced any new limits. The core spec already allows syntactic limits on everything here (including both the number and size of elem segments), and execution limits on the size of tables and memories. GC things are well-covered by the core spec too. If it looks like I added any limits here, it's either because it was missing from the JS API spec only, or just part of the general refactoring I did to the section.

eqrion (approved):
Sorry for the delay. This LGTM.

document/js-api/index.bs
```diff
@@ -908,6 +913,8 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     1. Let |store| be the [=surrounding agent=]'s [=associated store=].
     1. Let |ret| be the [=mem_size=](|store|, |memaddr|).
     1. Let |store| be [=mem_grow=](|store|, |memaddr|, |delta|).
+
+        Note: This should check the <a href="#limits-runtime">runtime implementation limits</a>.
```
Ms2ger, line 917: "should" is an RCF2119 keyword, please don't use it in notes.

document/js-api/index.bs
```diff
@@ -608,6 +610,7 @@ The verification of WebAssembly type requirements is deferred to the
     1. Let |result| be [=module_instantiate=](|store|, |module|, |imports|).
     1. If |result| is [=error=], throw an appropriate exception type:
         * A {{LinkError}} exception for most cases which occur during linking.
+        * If a <a href="#limits-runtime">runtime implementation limit</a> is exceeded, throw a {{RangeError}}.
```
Ms2ger, line 613: This whole step is a mess, but that predates your change. Ideally core would tell us what happened rather than making us hand-wave like this.

document/js-api/index.bs
```diff
@@ -2211,47 +2217,53 @@ Note: ECMAScript doesn't specify any sort of behavior on out-of-memory condition
 
 <h2 id="limits">Implementation-defined Limits</h2>
 
-The WebAssembly core specification allows an implementation to define limits on the syntactic structure of the module.
+The WebAssembly core specification allows an implementation to define limits on the syntactic structure of a module and on runtime resources.
```
Ms2ger, line 2220: I couldn't find where it says that. Do you happen to have a pointer?

document/js-api/index.bs
```diff
@@ -879,6 +882,8 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
     1. If |descriptor|["maximum"] [=map/exists=], let |maximum| be [=?=] [=AddressValueToU64=](|descriptor|["maximum"], |addrtype|); otherwise, let |maximum| be empty.
     1. Let |memtype| be [=memory type=] |addrtype| { **min** |initial|, **max** |maximum| }.
     1. If |memtype| is not [=valid memtype|valid=], throw a {{RangeError}} exception.
+    1. If |maximum| is not empty:
+        1. If |addrtype| is "i64" and |maximum| exceeds the [=memory type size limit=], throw a {{RangeError}} exception.
```
Ms2ger, line 886: Why is this here? Is it guaranteed there's no other callers of mem_alloc? If there are others, should this restriction not also be applied there?

Ms2ger (commented):
(Partial review, ran out of time.)

document/js-api/index.bs
bvisness: That would be the Execution section of the "implementation limitations" section here: https://webassembly.github.io/spec/core/appendix/implementation.html

document/js-api/index.bs
bvisness: What would be the conventional way of writing guidance like this? My intent is to call the reader's attention to the place where runtime implementation limits are relevant.

document/js-api/index.bs
bvisness: This is so long ago now that I don't quite remember why this is structured this way. It seems to me that it could be covered by or folded into the `{{RangeError}}` case on the `mem_alloc` step below.

PR #2092, opened 2026-02-24
Add wasm at the tripple ticks

micttyoid:


rossberg:
Note that the proposals directory is an archive of historical documents, so we do not normally update them.

micttyoid:
Oops, sorry! Let's close.

PR #2095, opened 2026-02-25
Use elemexpr in the text format instead of just expr

bvisness:
The text format spec for elem segments states that the grammar for elemlist is a ref type followed by list(expr). But this is obviously wrong because it leaves elemexpr completely unused, disallowing the (item (foo)) syntax.

Currently a draft PR because I can't get it to build locally for inscrutable reasons.


rossberg:
I tried locally and ran into a "too few arguments" type error from SpecTec. Is that what you're referring to? That turned out to be some mishandling of syntactic sugar for arguments when using an identifier declared only later. Should be fixed on trunk now.

bvisness:
Yes, that's the same bug I got. Glad to know it's already fixed.

bvisness:
Sorry for force pushing :) It should be good to go now, hopefully.

PR #2097, opened 2026-02-27
[test] Improve array.init_elem tests

tlively:
Test the identities of the initialized elements. This would have caught
a bug in the binaryen interpreter where the first element of the range
was copied `n` times. Also add a test that initializing from the same
element segment multiple times writes the same references each time
(rather than re-evaluating the initializer expressions, for example).
This catches another bug in the binaryen interpreter.

See https://github.com/WebAssembly/binaryen/pull/8349.

### DIFF
--- a/test/core/gc/array_init_elem.wast
+++ b/test/core/gc/array_init_elem.wast
@@ -42,23 +42,30 @@
 )
 
 (module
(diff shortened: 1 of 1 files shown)


tlively:
cc @stevenfontanella

PR #2100, opened 2026-03-06
fix(test): correct expected test invocation

marxin:
In both tested tail-call invocations, the exception is thrown, but caught. That's why we can't use `assert_exception`.


marxin:
Apparently, it's a documented/discussed `spec` corner case: https://github.com/WebAssembly/exception-handling/issues/249.

rossberg:
Right, the semantics of return_call is to first unwind the current call frame and then call the function. That's what makes it a tail call. It doesn't matter if the current call frame had a handler before, that is gone by the time the next function is called. If a producer wants exceptions from the callee to be caught, it should not generate a tail call.

marxin:
Gotcha, thanks for clarification!

PR #2101, opened 2026-03-08
[spec] Fixing memarg argument types

raoxiaojia:
The argument types of `n` and `m` are [switched ](https://github.com/WebAssembly/spec/blob/c07e5c9a80fb2ee593648c5957ccd2f0ad6f40fc/specification/wasm-3.0/6.3-text.instructions.spectec#L177) in the spec DSL construct for `memarg`:
```
grammar Tmemarg_(N) : memarg =
  | n:Toffset m:Talign_(N) => {ALIGN n, OFFSET m}
```
Looking at the binary format, it seems that it is intended for `n` to be the alignment and `m` to be the offset, so the above should be
```
grammar Tmemarg_(N) : memarg =
  | n:Talign_(N) m:Toffset => {ALIGN n, OFFSET m}
```
instead.

### DIFF
--- a/specification/wasm-latest/6.3-text.instructions.spectec
+++ b/specification/wasm-latest/6.3-text.instructions.spectec
@@ -175,7 +175,7 @@ grammar Tplaininstr_(I)/elem : instr = ...
 ;; Memory instructions
 
 grammar Tmemarg_(N) : memarg =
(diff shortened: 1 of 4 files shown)


raoxiaojia:
Think it's all fixed. Prose doesn't seem to be affected as it's accessing `memarg` by field names directly.

PR #2107, opened 2026-03-10
[spec] Cumulative fixes for text format and binary format specification

raoxiaojia:
This PR fixes several errors in the text format and binary format spec. I include a description for each in the order of their appearances in the DSL sources.

1. At the definition of [`5.4-Belemkind`](https://github.com/WebAssembly/spec/blob/fa8ea418ffe69976c5bad90b47ab38c8e5c23857/specification/wasm-3.0/5.4-binary.modules.spectec#L97): current spec has
```
grammar Belemkind : reftype hint(desc "element kind") =
  | 0x00 => REF NULL FUNC
```
Which parses the only allowed elemkind `0x00` to a nullable funcref. This disagrees with the [reference interpreter](https://github.com/WebAssembly/spec/blob/fa8ea418ffe69976c5bad90b47ab38c8e5c23857/interpreter/binary/decode.ml#L1154), which parses it to a non-nullable one:
```
let elem_kind s =
  match byte s with
  | 0x00 -> (NoNull, FuncHT)
  | _ -> error s (pos s - 1) "malformed element kind"
```
I think the reference interpreter is correct here, because in [`elem.wast`](https://github.com/WebAssembly/spec/blob/fa8ea418ffe69976c5bad90b47ab38c8e5c23857/test/core/elem.wast#L87), there's this test case:
```wat
(module
  (func $f)
  (func $g)

  (table $t 10 (ref func) (ref.func $f))
  (elem (i32.const 3) $g)
)
```
which has a table containing non-nullable funcref, and then try to initialise it with some `elem`. The matching relation for `reftype` states that non-nullable funcref matches nullable ones but not the other way round, so the `funcref`s from `elem` must not be nullable. Let me know if I read anything wrong and it's the reference interpreter that needs to get fixed though.

2. A minor typo at [6.2-abbreviation for `Tresult`](https://github.com/WebAssembly/spec/blob/fa8ea418ffe69976c5bad90b47ab38c8e5c23857/specification/wasm-3.0/6.2-text.types.spectec#L94).

3. The text format spec for composite types in 6.2 currently states [the following](https://github.com/WebAssembly/spec/blob/fa8ea418ffe69976c5bad90b47ab38c8e5c23857/specification/wasm-3.0/6.2-text.types.spectec#L96) for their associated identifier contexts:
```
grammar Tcomptype_(I) : (comptype, idctxt) =
  | "(" "struct" (ft,id?)*:Tlist(Tfield_(I)) ")" => (STRUCT ft*, {FIELDS (id?)*})
  | "(" "array" ft:Tfieldtype_(I) ")" => (ARRAY ft, {})
  | "(" "func" (t_1,id?)*:Tlist(Tparam_(I)) t_2*:Tlist(Tresult_(I)) ")" => (FUNC t_1* -> t_2*, {})
```
So only structure types contribute to the `field` component of the context. This means that the running identifier context's `field` component will be a list of length equal to the number of structure types it has seen (instead of the number of all component types) due to context concatenation being defined as listwise-concat by each component. But the text format resolution of e.g. `struct.get x i` lookup from both the `TYPES` and `FIELDS` components of the context at index `x`. This means if a structure type definition occurs after some other composite types, the `FIELDS` lookup will be out of bound for any `struct.get` that refers to it. A concrete example is the following:
```wat
(module
  (type $ft (func))
  (type $st (struct (field $x i32)))
  (func (export "get") (param $s (ref $st)) (result i32)
    (struct.get $st $x (local.get $s))
  )
)
```
(note that wabt doesn't seem to support `struct.get` yet so this doesn't run successfully in e.g. wat2wasm, but the reference interpreter runs this fine).

There are several ways to fix this, my proposed fix simply returns `{FIELDS ( () )}` (a singleton list containing an `eps`) to align the indices of `FIELDS` with `TYPES`, which seems to match the intention of the spec when it tried to return just `{}`.

4. This is a continuation of the fix to `memarg` text format which I didn't fix properly -- I've later realised that the current `Talign` definition has some issues as well. This is the [current version](https://github.com/WebAssembly/spec/blob/fa8ea418ffe69976c5bad90b47ab38c8e5c23857/specification/wasm-3.0/6.3-text.instructions.spectec#L184-L186):
```
grammar Talign_(N) : u64 =
  | "align=" m:Tu64 => m  -- if $(m = 2^n)
  | eps => N
```
But the AST stores the alignment as exponents (`n`) instead of the raw value (`m=2^n`), so this should return `n` instead of `m`. Also the alignment parameter has type `u32` in the AST instead of `u64`. I fixed both of these, and also made a cosmetic change of the binder name of `Toffset` from `n` to `m` to match that of `Tmemarg`.

5. Several minor typos at [6.3-abbreviation for `v128.storenn_lane`](https://github.com/WebAssembly/spec/blob/fa8ea418ffe69976c5bad90b47ab38c8e5c23857/specification/wasm-3.0/6.3-text.instructions.spectec#L287).

### DIFF
--- a/spectec/test-latex/TEST.md
+++ b/spectec/test-latex/TEST.md
@@ -12633,9 +12633,9 @@ $$
 & & | & \mbox{‘\texttt{{(}}’}~~\mbox{‘\texttt{param}’}~~{{{\mathtt{valtype}}}_{I}^\ast}~~\mbox{‘\texttt{{)}}’} & \quad\equiv\quad{} & {(\mbox{‘\texttt{{(}}’}~~\mbox{‘\texttt{param}’}~~{{\mathtt{valtype}}}_{I}~~\mbox{‘\texttt{{)}}’})^\ast} \\
 & {{\mathtt{result}}}_{I} & ::= & \mbox{‘\texttt{{(}}’}~~\mbox{‘\texttt{result}’}~~t{:}{{\mathtt{valtype}}}_{I}~~\mbox{‘\texttt{{)}}’} & \quad\Rightarrow\quad{} & t \\
 & & | & \mbox{‘\texttt{{(}}’}~~\mbox{‘\texttt{result}’}~~{{{\mathtt{valtype}}}_{I}^\ast}~~\mbox{‘\texttt{{)}}’} & \quad\equiv\quad{} & {(\mbox{‘\texttt{{(}}’}~~\mbox{‘\texttt{result}’}~~{{\mathtt{valtype}}}_{I}~~\mbox{‘\texttt{{)}}’})^\ast} \\
-& {{\mathtt{comptype}}}_{I} & ::= & \mbox{‘\texttt{{(}}’}~~\mbox{‘\texttt{struct}’}~~{({\mathit{ft}}, {{\mathit{id}}^?})^\ast}{:}{\mathtt{list}}({{\mathtt{field}}}_{I})~~\mbox{‘\texttt{{)}}’} & \quad\Rightarrow\quad{} & (\mathsf{struct}~{{\mathit{ft}}^\ast}, \{ \mathsf{fields}~{({{\mathit{id}}^?})^\ast} \}) \\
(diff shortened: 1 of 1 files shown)


raoxiaojia:
Thanks for the review! Yes I agree with both comments.

PR #2108, opened 2026-03-11
[spec] Add missing side condition on growtable/memory

rossberg:
This adds back missing side conditions to check against the max table/memory size allowed per address type.

@f52985, I'm again seeing interpreter failures that I don't understand. The PR just adds simple arithmetic side conditions, but I get type mismatches for them. Can you please have a look?


f52985:
Fixed. It was simply missing a case for the optionally itered (`?`) boolean premise.

PR #2114, opened 2026-03-14
[spec] fix _binary-mut anchor to point to the place it's defined

### LINKED ISSUE #2113: `_binary-mut` anchor points to the wrong place
`_binary-mut` is anchored to the global types section, but it's actually defined in the comptype section. The generated documentation has anchor links that take you to the global types section, where it's not actually defined.

Taywee:
The comptype section actually defines mut, but clicking a mut link takes you to the global section.

closes #2113


rossberg (approved):
Thanks!

Issue #2034, opened 2025-11-26
[js-api] Ambiguity in `WebAssembly.instantiate(bytes)` if `WebAssembly.Module` is a thenable


backes:
I believe the current JS-API spec underspecifies the behavior of `WebAssembly.instantiate(bytes)` when `WebAssembly.Module` has been modified to include a `then` method (making the module a "thenable").

Reproducer:
```
const bytes = new Uint8Array([0,97,115,109,1,0,0,0]);  // Minimal Wasm bytes.
WebAssembly.Module.prototype.then = resolve => resolve(17);
WebAssembly.instantiate(bytes).then(v => print(`Resolved to ${v} / ${v.instance}`), e => print(`Rejected: ${e}`));
```

(to be run in a `d8`, `jsc`, or `sm` shell; on the Web replace `print` by `console.log`).

**The issue:**
According to the [JS-API spec](https://webassembly.github.io/spec/js-api/#dom-webassembly-instantiate), `instantiate(bytes)` performs the following steps:
1. Asynchronously compile the bytes into a WebAssembly module, let `promiseOfModule` be the result.
2. [Instantiate](https://webassembly.github.io/spec/js-api/#instantiate-a-promise-of-a-module) that `promiseOfModule`.

The second step creates a new promise which reacts to the first promise.

At the end of the first step (when compilation finishes), the internal promise is [`resolve`d](https://webidl.spec.whatwg.org/#resolve) with the `module` object. According to the [ECMAScript spec for promise resolution](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise-resolve-functions), if the resolution value (the Module) has a `then` method, that method is invoked to determine the final resolution.

In the reproducer above, the user-provided `then` resolves the promise to `17`. Consequently, the next step, [`instantiate-a-promise-of-a-module`](https://webassembly.github.io/spec/js-api/#instantiate-a-promise-of-a-module) receives `17` instead of a `WebAssembly.Module`.

Currently, the spec does not define what should happen if the promise resolves to a value that is not a `WebAssembly.Module`.

**Current Implementation Behavior:**
- **SpiderMonkey & JavaScriptCore**: Seem to ignore the `then` property on the prototype, resolving with the Module object regardless.
[...]

PR #2118, opened 2026-03-17
[js-api] Handle thenable Module in instantiate(bytes)

backes:
Add an explicit check in "instantiate a promise of a module" to ensure the fulfilled value of the compilation promise implements the Module interface. This prevents a modified WebAssembly.Module.prototype.then from subverting the instantiation process with a non-Module value.

Also add regression tests in test/js-api/constructor/instantiate.any.js.

Fixes https://github.com/WebAssembly/spec/issues/2034


eqrion:
This fix looks correct to me, but I added a comment on the original issue for more discussion.

PR #2120, opened 2026-03-23
Fix subst_externtype matching on incorrect argument types

raoxiaojia:
The substitution function for `externtype` is incorrectly using the substitution for `deftype` for function externs instead of the intended `typeuse`, missing the other two cases of `typeuse` (most importantly `typeidx`). 

The [reference interpreter](https://github.com/WebAssembly/spec/blob/228e9109e2f6e7a23bec92f590eb8dc83a7963e3/interpreter/syntax/types.ml#L232) is correct by using `subst_typeuse`.

The AL interpreter seems to be unaffected because iirc only `deftype`s exist during actual runtime, and `subst_typeuse` simply dispatches deftypes using `subst_deftype`, so this type error is hidden by implicit coercions (either in IL or elaboration), taking the substitution function as deliberately only defined on func externs that are `deftype`s. Is there something we could do in any of the checks to produce a warning for this kind of scenario (which would have made it obvious)?

For the other part of the tests, module validation should have been affected if generated from the spectec DSL, but that is out of the scope of AL as it focuses on generating the interpreter only and delegates validation etc to the reference interpreter, which is correct on this. Latex backend doesn't seem to print the full substitution definition either. As a result, the prose seems to be the only one that makes this change observable.

There also seems to be some test mismatches under `spectec/test-interpreter` (I think this is the one for AL?), but those are pre-existing in the current main (probably after #2093 looking at the errors).


rossberg:
You'll need to sync wasm-latest. (Tip: `make sync-diff` in the specification dir usually does the trick.)

raoxiaojia:
My local `dune test` is somehow caching builds a lot for some reason -- this should fix all the test expectations.

PR #2123, opened 2026-03-25
[test][js-api] WebAssembly.Exception.length should be 2

Liedtke:
Based on: https://webassembly.github.io/spec/js-api/#exceptions
```
[LegacyNamespace=WebAssembly, Exposed=(Window,Worker,Worklet)]
interface Exception {
  constructor(Tag exceptionTag, sequence<any> payload, optional ExceptionOptions options = {});
  any getArg(Tag exceptionTag, [EnforceRange] unsigned long index);
  boolean is(Tag exceptionTag);
  readonly attribute (DOMString or undefined) stack;
};
```

At least our WPT tests are unhappy about V8 returning 1 here instead of 2 and according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length#description) it seems like this number should include all parameters before the first one with a default value?
(I'm not very familiar with JS semantics and WebIDL definitions.)

@Ms2ger: Could you please take a look?

### DIFF
--- a/test/js-api/exception/constructor.tentative.any.js
+++ b/test/js-api/exception/constructor.tentative.any.js
@@ -10,7 +10,7 @@ test(() => {
 }, "name");
 
 test(() => {
-  assert_function_length(WebAssembly.Exception, 1, "WebAssembly.Exception");
+  assert_function_length(WebAssembly.Exception, 2, "WebAssembly.Exception");
 }, "length");
 
 test(() => {


eqrion:
It looks like we need to 'Set length to the length of the shortest argument list of the entries in S.'  where S is the effective overload set [1] [2]. This gets passed to create builtin function which calls SetFunctionLength.

I don't see in the WebIDL spec where we determine the 'length of an argument list' for a constructor, and if an optional argument counts or not. It seems reasonable to exclude it, but just want to confirm first.

[1] https://webidl.spec.whatwg.org/#interface-object
[2] https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-createbuiltinfunction

eqrion:
> I don't see in the WebIDL spec where we determine the 'length of an argument list' for a constructor, and if an optional argument counts or not. It seems reasonable to exclude it, but just want to confirm first.

This is done by computing the effective overload set [1] operation. A constructor with an optional parameter appears as two overloads, one with the parameter and one without. The length is then the minimum which doesn't contain the optional parameter.

This LGTM then.

[1] https://webidl.spec.whatwg.org/#compute-the-effective-overload-set

PR #2125, opened 2026-03-27
[spec] Define soundness rules in SpecTec

### LINKED ISSUE #2141: [spec] Extended subtype validity relation missing validity on supertype
The spec defines the following [generalised version](https://github.com/WebAssembly/spec/blob/f9c743aeccadabbcdf7b05508b95d753d9fa1bcd/specification/wasm-3.0/2.1-validation.types.spectec#L161) for validity of `subtype` that allow the supertype to be a general `typeuse` instead of just a type index `_IDX` (this corresponds to the extension of `typeuse` from just an index to also include `rec n` and `deftype`).:
```
rule Subtype_ok2:
  C |- SUB FINAL? typeuse* compttype : OK x i
  -- if |typeuse*| <= 1
  -- (if $before(typeuse, x, i))*
  -- (if $unrollht(C, typeuse) = SUB typeuse'* comptype')*
  ----
  -- Comptype_ok: C |- comptype : OK
  -- (Comptype_sub: C |- comptype <: comptype')*
```
But the rule is missing a check that the `typeuse` itself needs to be valid (`(Typeuse_ok: C |- typeuse : OK)*`). Note that the original version of `subtype_ok` does not require such a check, since only a type index `_IDX` is allowed. 

This only affects the metatheoretic soundness proofs, since the extended supertype definition cannot appear in real modules, and the above definition is only involved in the soundness proofs. I noticed that this definition is rendered in the appendix despite being formulated in 2.1, so maybe it can be fixed as a part of #2125 .

The specific soundness proof that got me stuck and raise this issue is a bit involved -- I include a description of the issue below.

<details> 
<summary>Long text of explanation </summary>

I've crafted a mechanised interpreter and have been trying to proving interpreter correctness against the opsem. Several instructions (e.g. ref.test, ref.cast, br_on_cast) require the matching relation to be dynamically checked at runtime and return different results depending on whether the matching holds. The interpreter calls a computable matching oracle (following the design of the reference interpreter's match.ml), and correctness of the "else" branch requires oracle completeness: if the declarative relation Heaptype_sub C ht1 ht2 holds, then the oracle returns true.

The completeness proof is by structural induction on Heaptype_sub. The critical case is Heaptype_sub/trans:
```
rule Heaptype_sub/trans:
  C |- ht_1 <: ht_2
  -- Heaptype_ok: C |- ht' : OK
  -- Heaptype_sub: C |- ht_1 <: ht'
  -- Heaptype_sub: C |- ht' <: ht_2
```
The inductive hypotheses give oracle results for ht_1 <: ht' and ht' <: ht_2, but combining them requires proving oracle transitivity -- which in turn requires Heaptype_ok on both endpoints (instead of just the middle heaptype) -- otherwise there are counterexamples with ill-formed heaptypes where the oracle is not transitive while the declarative rules are transitive (see below). Nevertheless, I don't consider this to be a bug for the declarative rules of heaptype matching, since it only manifests for ill-typed heaptypes at the endpoints. However, this forces the completeness theorem to carry Heaptype_ok as a precondition on both endpoints.

Now, Deftype_sub/super says:
```
rule Deftype_sub/super:
  C |- deftype_1 <: deftype_2
  -- if $unrolldt(deftype_1) = SUB final? typeuse* ct
  -- Heaptype_sub: C |- typeuse*[i] <: deftype_2
```
To prove oracle completeness for this case, it is needed to invoke the completeness result on the `Heaptype_sub` premise, which requires `Heaptype_ok C (.typeuse (typeuse*[i]))`. Since `typeuse*[i]` comes from `$unrolldt(deftype_1)`, this can only be from `Deftype_ok C deftype_1`. This is the case for `_IDX` supertypes (validity from context lookup) and `REC` supertypes (validity from the same `Rectype_ok`). But for inline `deftype` supertypes, `Subtype_ok2` only checks `$unrollht` (unrollability) without checking `Typeuse_ok` (validity), so `Deftype_ok` does not guarantee `Heaptype_ok` for the supertype. Compounded with the above requirement of heaptype_ok in the completeness proof, a counterexample can be crafted by plugging in the above ill-typed transitivity example into the supertype of a deftype.

Note that there's no concrete module-level counterexample, because the module format does not support generalised supertypes. The following is an encoding of the counterexample directly in the OCaml AST.
```ocaml
let debug_inline_supertype_witness () =
    let open Types in
    let open Match in

    let show_bool label b =
      Printf.printf "%s = %b\n" label b
    in

    let dt_func : deftype =
      DefT (RecT [
        SubT (NoFinal, [], FuncT ([], []))
      ], 0l)
    in

    let dt_bad : deftype =
      DefT (RecT [
        SubT (NoFinal, [Def dt_func], StructT [])
      ], 0l)
    in

    let dt_outer : deftype =
      DefT (RecT [
        SubT (NoFinal, [Def dt_bad], StructT [])
      ], 0l)
    in

    let c : Match.context = [] in

    let rt_func : reftype =
      (Null, UseHT (Def dt_func))
    in

    let rt_outer : reftype =
      (Null, UseHT (Def dt_outer))
    in

    let rt_abs_func : reftype =
      (Null, FuncHT)
    in

    let bad_sub_func =
      match_deftype c dt_bad dt_func
    in

    let outer_sub_func =
      match_deftype c dt_outer dt_func
    in

    let func_sub_abs =
      match_reftype c rt_func rt_abs_func
    in

    let outer_sub_abs =
      match_reftype c rt_outer rt_abs_func
    in

    Printf.printf "dt_func  = %s\n" (string_of_deftype dt_func);
    Printf.printf "dt_bad   = %s\n" (string_of_deftype dt_bad);
    Printf.printf "dt_outer = %s\n\n" (string_of_deftype dt_outer);

    show_bool "match_deftype [] dt_outer dt_func" outer_sub_func;
    show_bool "match_reftype [] (Null, UseHT (Def dt_func)) (Null, FuncHT)" func_sub_abs;
    show_bool "match_reftype [] (Null, UseHT (Def dt_outer)) (Null, FuncHT)" outer_sub_abs;

    print_endline "";
```
The first two matchings are true while the last one is false, and `dt_outer/dt_func` are both valid deftypes given the definition in the current appendix (as subtype validity does not check supertype validity).

</details>

rossberg:
Formulate extra rules from Type Soundness Appendix in SpecTec. Also closes #2141 and #2142.

Fix SpecTec sideconditions pass to not generate vacuous iterated premises.

@f52985, the prose backend crashes failing to find one of the added relations. That may have to do with the preceding "Yet" warning from `translate_rulepr`, which IIUC means that something is unimplemented in the backend. Can you please have a look?
```
../../specification/wasm-latest/7.1-soundness.configurations.spectec:211.62-211.76: translate_rulepr: Yet `(fv_1, s, fv_2)`
Untranslated relation Instr_ok2: `%;%|-%:%`(store, context, instr, instrtype)
Untranslated relation Instrs_ok2: `%;%|-%:%`(store, context, instr*, instrtype)
Untranslated relation Expr_ok2: `%;%|-%:%`(store, context, expr, resulttype)
../../spectec/spectec: uncaught exception Failure("Unknown relation id: NotImmReachable")

Backtrace:
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Backend_prose__Gen.prem_to_instrs in file "src/backend-prose/gen.ml", line 288, characters 16-58
Called from Backend_prose__Gen.prem_to_instrs in file "src/backend-prose/gen.ml", line 327, characters 42-61
Called from Stdlib__List.concat_map in file "list.ml", line 288, characters 32-37
Called from Stdlib__List.prepend_concat_map in file "list.ml", line 292, characters 20-46
Called from Stdlib__List.map in file "list.ml", line 85, characters 15-19
Called from Backend_prose__Gen.prose_of_rules in file "src/backend-prose/gen.ml", lines 432-434, characters 4-48
Called from Backend_prose__Gen.proses_of_rel in file "src/backend-prose/gen.ml", line 451, characters 23-70
Called from Stdlib__List.concat_map in file "list.ml", line 288, characters 32-37
Called from Stdlib__List.prepend_concat_map in file "list.ml", line 292, characters 20-46
Called from Backend_prose__Gen.prose_of_rels in file "src/backend-prose/gen.ml" (inlined), line 579, characters 20-48
Called from Backend_prose__Gen.gen_validation_prose in file "src/backend-prose/gen.ml" (inlined), line 583, characters 2-39
Called from Backend_prose__Gen.gen_prose in file "src/backend-prose/gen.ml", line 648, characters 25-48
Called from Dune__exe__Main in file "src/exe-spectec/main.ml", line 306, characters 18-54
```


raoxiaojia:
A comment on the soundness appendix -- is there any plan of adding a definition for the validity of the typing context C itself (which probably just says that all types in C.Types are valid)? Unlike pre-3.0, the type system is quite rich currently, and C.Types can contain ill-formed types in itself (e.g. invalid deftypes/heaptypes etc). The validity of C is required to prove that intermediate matchings for deftypes/heaptypes are correct.

rossberg:
@raoxiaojia, good point. Yes, at this point we need that, too. I'll look into it.

rossberg:
@raoxiaojia, I just pushed a commit that adds a wf-rule for contexts. Can you please check if it makes sense to you?

raoxiaojia:
@rossberg Looks mostly fine, except this line which I'm unsure:
```
-- (Subtype_ok2: {TYPES dt^n, RECS st^m} |- st : OK n i)^(i<m)
```
I would have thought `n` is the starting offset of the rec group, which needs to increase per element in the group, so should it be `OK (n+i) i`? 

Upd: I guess using `OK n i` is saying that the `REC` group cannot access any types in the same group using `_IDX`, which might be intended, since all subtypes should already be rolled (IDX replaced with REC) at the point of validating the group (which populates the C.REC entries). In comparison, the current statement of `rectype_ok2` allows this shifted offset per element: 
```
rule Rectype_ok2/empty:
  C |- REC eps : OK x i

rule Rectype_ok2/cons:
  C |- REC (subtype_1 subtype*) : OK x i
  -- Subtype_ok2: C |- subtype_1 : OK x i
  -- Rectype_ok2: C |- REC subtype* : OK $(x+1) $(i+1)
```
but that seems to be also fine because `rectype_ok2` may need to handle pre-rolled types.

rossberg:
You're right, thanks for spotting! Fixed.

raoxiaojia:
@rossberg @zilinc mentioned that the fix for `rectype_ok` was not synced to 3.0 yet (only on latest) -- just checking that the soundness section is intended to be part of the 3.0 spec as well?

zilinc:
> @rossberg @zilinc mentioned that the fix for `rectype_ok` was not synced to 3.0 yet (only on latest) -- just checking that the soundness section is intended to be part of the 3.0 spec as well?

At least I can see 2.1 in wasm-3.0 and latest are done differently.

rossberg:
Ah, sorry, propagated now.

f52985:
> the prose backend crashes failing to find one of the added relations. That may have to do with the preceding "Yet" warning from translate_rulepr, which IIUC means that something is unimplemented in the backend. Can you please have a look?

Yes, the prose generator was somewhat overfitted to specific relation forms. I have fixed it so that it can robustly handle the new relation formats. The rendered prose is still somewhat naive and may occasionally require a prose hint, but the generator should no longer crash in most cases.

rossberg:
Hi @f52985, thanks for the hotfix! That works, but when I renamed `ImmReachable` to `ImmutReachable`, it crashed again. I was able to work around that by adjusting `valid_helper_relations`. Two questions for further improvements, though:

- Can we make it not crash, but output an error message, ideally with advice on how to fix the error?
- Would it be possible to get rid of the hard-coded relations list in favour of either using hints, or even better, compute the list automatically by taking the dependencies of relations using `|-`?

f52985:
> * Can we make it not crash, but output an error message, ideally with advice on how to fix the error?
> * Would it be possible to get rid of the hard-coded relations list in favour of either using hints, or even better, compute the list automatically by taking the dependencies of relations using `|-`?

Done. Now it automatically detects the validation rules, based on the dependencies, removing all the hardcodings.

rossberg:
@f52985, oh excellent, thanks for the super-fast granting of my wishes! Since this PR already landed, can you please create a new one with that change?

f52985:
Sure, I opened the new PR: #2147

PR #2133, opened 2026-04-03
Fix typo about $(T)+ syntax

masak:
The typo looks like the result of a copy+paste, and then forgot to update the copy.


rossberg (approved):
Thanks!

PR #2138, opened 2026-04-07
[spectec] Fix `$fnat` show hint

### LINKED ISSUE #2135: [spec] Typo in `select` example of instructions validation
In the section about the validation of instructions (https://webassembly.github.io/spec/core/valid/instructions.html#instructions), there is a note that contains an example using instruction `select` with `f64` consts. The example is rendered as:

> `(f64.const +64) (f64.const +64) (f64.const +64) (select)`

but it should contain 1, 2 and 3 instead of the three +64, as it is suggested by the previous example (which uses `i32`) and by the fact that the source code contains instructions `$fnat(64, 1)`, `$fnat(64, 2)` and `$fnat(64, 3)`. Maybe the problem is in the rendering of these instructions.

Moreover, I think that the third parameter should still be an `i32` instead of an `f64`, as the type of the `select` instruction is `t t i32 -> t`.

alessiofaieta:
Fix show hint for `$fnat` and update tests.

Fix `select` example in the section about instructions validation in the spec.

Closes #2135.


rossberg (approved):
Thanks!

alessiofaieta:
Thank you for your help!

PR #2139, opened 2026-04-09
[spec] Fix typo causing bad rendering of instruction in a note

alessiofaieta:
The instruction was rendered as `const i32 2` instead of `i32.const 2` because `instr` was missing.


rossberg (approved):
Thanks!

Issue #2144, opened 2026-04-20
Constant expression section says "val" instead of "instr"


bvisness:
The validation section on [constant expressions](https://webassembly.github.io/spec/core/valid/instructions.html#constant-expressions) is structured like so:

```
instr* is constant if:
  - For all instr in instr*:
    - instr is constant.
val is constant if:
  - <list of all the constant instructions>
```

Where did `val` come from? It links to [#syntax-val](https://webassembly.github.io/spec/core/exec/runtime.html#syntax-val), which is obviously wrong. Unfortunately, since this prose is generated by SpecTec, I have been unable to track down the source of this bug on my own. (Too in the weeds for me to work through right now, unfortunately.)

rossberg:
Yes, that looks like a bug in prose generation. `val` is a subset of `instr`. I guess there is some respective naming heuristics in the prose generator that goes wrong here. @f52985, can you please take a look?

bvisness:
Seems this was indeed fixed in #2161.

PR #2145, opened 2026-04-20
Fix cross-references to constant expressions

bvisness:
Links to constant expressions in the spec were incorrectly linking to the t.const instruction, which is irrelevant.

This is the considerably more straightforward counterpart to #2144 :)


rossberg (approved):
Thanks!

Issue #2142, opened 2026-04-19
[spec] Rectype_ok rules contain a loophole that allows oob `REC` typeuses in nested deftypes


raoxiaojia:
There seems to be an issue in the `Rectype_ok` [rules ](https://github.com/WebAssembly/spec/blob/f9c743aeccadabbcdf7b05508b95d753d9fa1bcd/specification/wasm-3.0/2.1-validation.types.spectec#L171) that allow oob `REC` typeuses to be valid. The relevant rules are the following:
```
rule Rectype_ok/empty:
  C |- REC eps : OK(x)

rule Rectype_ok/cons:
  C |- REC (subtype_1 subtype*) : OK(x)
  -- Subtype_ok: C |- subtype_1 : OK(x)
  -- Rectype_ok: C |- REC subtype* : OK($(x+1))

rule Rectype_ok/_rec2:
  C |- REC subtype* : OK(x)
  -- Rectype_ok2: C, RECS subtype* |- REC subtype* : OK x 0

rule Rectype_ok2/empty:
  C |- REC eps : OK x i

rule Rectype_ok2/cons:
  C |- REC (subtype_1 subtype*) : OK x i
  -- Subtype_ok2: C |- subtype_1 : OK x i
  -- Rectype_ok2: C |- REC subtype* : OK $(x+1) $(i+1)
```
Note that `Rectype_ok/cons` allows the subtype to be validated directly using whatever `C.RECS` was. Now consider the following example, with an outer 2-member `rec` group `sts = REC [st_a, st_b]` containing the following:
```
  st_a = SUB [] [] (STRUCT [ref (deftype inner_dt)])
  st_b = SUB [] [] (STRUCT [])
```
where the inner deftype is the 0th projection of a length-1 `rec` group which contains an oob reference:
```
  inner_dt = DEF (REC [inner_st]) 0
  inner_st = SUB [] [] (STRUCT [ref (REC 1)])
```
I claim that the outer rec group `sts` is valid under an empty context `C` via the following chain of sufficient conditions (I include the explicit *type_ok relations instead of using the C |- notations to avoid ambiguity):
```
rectype_ok C sts x
<= (by (rectype_ok/rec2)) rectype_ok2 (C' := {C with RECS := sts}) (REC sts) x 0
<= (by (rectype_ok2/cons)) subtype_ok2 C' st_a x 0 && rectype_ok2 C' [st_b] (x+1) 1
```
[...]

PR #2148, opened 2026-04-22
[spec] Slight simplification to Rectype_ok2 judgement

rossberg:
Remove the [redundant](https://github.com/WebAssembly/spec/issues/2142#issuecomment-4294543560) typeidx on the Rectype/Subtype_ok2 judgements.

@raoxiaojia, PTAL.

@zilinc, FYI.


raoxiaojia:
The simplification looks correct to me.

I'm curious on the choice of type indices there for `rectype_ok/ok2`, which uses `typeidx` (bounded by 2^32-1) and `nat` respectively. Is this a deliberate choice to limit the number of subtypes in a module to `2^32-1`? Note that the module definition itself doesn't have such a bound: it just says that a modules's types is a general sequence of `rectype*` instead of using the bounded sequence definition in spectec (`list(rectype)`). In fact, even the latter would not work, because that is only setting a bound on the number of rec *groups* but not the individual subtypes themselves, whereas using `typeidx` as the index in `rectype_ok` basically makes all types that appear at a position `>=2^32` inaccessible (which seems intended).

raoxiaojia (approved):

rossberg:
Yes, `idx` and `list` are used over `nat` and `*` where the domain or length is constrained in the AST, usually reflecting the binary format.

You are totally right about the abstract syntax of modules not doing that! That's an oversight, will fix. (FWIW, the binary format indeed applies the limit to the number of type _groups_. So you can define more than 2^32 types in a module, making the upper ones inaccessible by Wasm syntax. Doesn't really do any harm, though.)

raoxiaojia:
Oops, I think there's some test expects that were not updated (frontend or sth).

rossberg:
Yeah, on it, but I'm on the train and the connection is horrible.

PR #2149, opened 2026-04-22
[spec] Use `fieldidx` instead of `u32`

alessiofaieta:
Use `fieldidx` instead of `u32` in the syntax and binary definitions of instructions `struct.get` and `struct.set`.


alessiofaieta:
Since I added the definition of `fieldidx` for the binary format, I was wondering if some new tests are required.

rossberg (approved):

PR #2154, opened 2026-04-24
[spectec] Fix prose title of if instruction

### LINKED ISSUE #2151: [spec] Missing `else` from `if` instruction in title
In section "Execution > Instructions > Control Instructions", the title for the `if` instruction ([here](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-if-mathit-bt-xref-syntax-instructions-syntax-instr-mathit-instr-1-ast-xref-syntax-instructions-syntax-instr-mathit-instr-2-ast)) is rendered as `if bt instr_1* instr_2*`, so an `else` is missing between `instr_1*` and `instr_2*`.

f52985:
Fixes #2151.

Previously, the `RuleA` algorithm used `atom` to render the title, which caused loss of information. It now uses `mixop` instead, preserving the full information and allowing the `else` atom to be rendered correctly.

### DIFF
--- a/spectec/src/backend-prose/render.ml
+++ b/spectec/src/backend-prose/render.ml
@@ -1483,20 +1483,22 @@ and render_instrs env algoname depth instrs =
 
 (* Prose *)
 
-let render_atom_title env name params =
-  (* TODO a workaround, for algorithms named label or name
-     that are defined as LABEL_ or FRAME_ in the dsl *)
-  let name' =
-    match name.it with
-    | Atom.Atom "label" -> Atom.Atom "LABEL_"
-    | Atom.Atom "frame" -> Atom.Atom "FRAME_"
-    | Atom.Atom s -> Atom.Atom (String.uppercase_ascii s)
-    | _ -> name.it
+let render_mixop_title env mixop params =
+  let head = Mixop.head mixop in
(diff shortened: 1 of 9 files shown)


spectec/src/backend-prose/render.ml
```diff
@@ -1483,20 +1483,22 @@ and render_instrs env algoname depth instrs =
...
-    | Atom.Atom s -> Atom.Atom (String.uppercase_ascii s)
-    | _ -> name.it
+let render_mixop_title env mixop params =
+  let head = Mixop.head mixop in
+  (* HARDCODE: Tweak for these instructinos *)
+  let mixop =
+    match head with
+    | Some atom ->
+      (match atom.it with
+      | Atom ("LABEL_" | "FRAME_" | "HANDLER_") -> Mixop.Atom atom
```
rossberg, line 1493: Out of curiosity: why is this special case needed?

rossberg (approved):
Thanks!

spectec/src/backend-prose/render.ml
f52985: Oh, that's because the algorithms (`RuleA`) for these context instructions do not have parameters, and the `tails` of the mixop should be dropped to match the arity.

PR #2156, opened 2026-05-05
[spec] Add missing meta-theory rules for recursive type indices

rossberg:
@raoxiaojia, PTAL.


raoxiaojia:
lgtm. I was unaware that spectec supports dropping implicitly default argument in a middle position as well (for empty typeuse* there) before this.

I thought about the interaction between heaptype unrolling and these new rules as well -- but I can't think of anything that breaks.

raoxiaojia (approved):

PR #2161, opened 2026-05-08
[spectec] Fix prose of the rule instr_const

f52985:
This PR fixes the prose for the `instr_const` rule, where the variable `val` was incorrectly used instead of `instr`, resolving #2144.

The root cause was that the rule unification logic only considered the first two rules (`CONST` and `VCONST`) and incorrectly assumed the unified variable type should be `val`. It now considers all cases and correctly generates the unified variable with type `instr`.

### DIFF
--- a/spectec/src/il2al/unify.ml
+++ b/spectec/src/il2al/unify.ml
@@ -115,7 +115,7 @@ let rec overlap env e1 e2 = if eq_exp e1 e2 then e1 else
   match e1.it, e2.it with
     (* Already unified *)
     | VarE _, _ when is_unified_exp e1 ->
-      e1
+      generalize_unified_wasm_val env e1 e2
     | IterE ({ it = VarE id; _} as e, i), _ when is_unified_id id.it ->
       let t = overlap_typ env e1.note e2.note in
       { e1 with it = IterE (e, i); note = t }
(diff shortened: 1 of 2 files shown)


spectec/src/il2al/unify.ml
```diff
@@ -202,6 +202,13 @@ let rec overlap env e1 e2 = if eq_exp e1 e2 then e1 else
       in
       { e1 with it; note = ty }
 
+(* HARDCODE: Prevent falsely unifying mixture of instr and val into `__unify:val` *)
+and generalize_unified_wasm_val env u e =
+  if Il2al_util.is_val u && not (Il2al_util.is_val e) then (
+    let new_var = VarE ("instr" $ no_region) $$ no_region % e.note in
```
rossberg: Would it be possible to avoid hard-coding the variable name? Could it invoke gen_new_unified without a prefix here right away?

(Btw, looking at gen_new_unified just now, I noticed that
```
  let var =
    match prefix with
    | Some prefix -> introduce_fresh_variable ~prefix env.frees ty
    | _ -> introduce_fresh_variable env.frees ty in
```
can be simplified to
```
  let var = introduce_fresh_variable ?prefix env.frees ty
```
)

f52985: Makes sense. Fixed it accordingly.

rossberg (approved):

spectec/src/il2al/unify.ml
Ahamed009: 1.

PR #2162, opened 2026-05-08
[spec] fix validation rules

zilinc:
* `Subtype_ok` rule too restrictive in the supertypes' syntax.
* `Module_ok` rule doesn't include references from `start?` and `export*`.


rossberg (approved):
Thanks!

PR #2167, opened 2026-05-13
[interpreter] Fix deftype closure invariant in validator

### LINKED ISSUE #2166: [interpreter] Bug in subtype matching for types with equal closures after deep `IDX` substitution
The reference interpreter's matching/validation algorithm seems to reject certain subtyping relations between types that have the same closure -- therefore valid under `deftype_sub/refl` -- but contains `IDX` references at depth >= 2. 

Consider this example:
```wat
(module
  (type $a (sub (struct)))
  (type $b (sub (struct)))
  (type $c (sub $a (struct)))
  (type $d (sub $b (struct)))
  (type $e (sub $c (struct)))
  (type $f (sub $d (struct)))
  (type $g (sub (struct (field (ref $e)))))
  (type $h (sub $g (struct (field (ref $f)))))
)
```
Or in AST:
```
dt_a = dt_b = [(SUB [] struct)].0
dt_c = [(SUB [IDX 0] struct)].0
dt_d = [(SUB [IDX 1] struct)].0
dt_e = [(SUB [IDX 2] struct)].0
dt_f = [(SUB [IDX 3] struct)].0
dt_g = [(SUB [] (struct (ref (IDX 4))))].0
dt_h = [(SUB [IDX 6] (struct (ref (IDX 5))))].0

C := { TYPES := [dt_a, dt_b, ..., dt_h] }
```
`dt_h` declares `dt_g` as a supertype, which triggers a subtyping check of their component types, which eventually comes down to a check of whether `IDX 5 <: IDX 4` as heaptypes. That should be true from the declarative rules, as their closures are equal. However, the reference interpreter rejects this module, claiming that `sub type 7 does not match super type 6`. 

I think this is because `match_deftype` only performs a shallow substitution against `C.types`, which is only correct wrt `clos_deftypes` if the substitution table is already IDX-free. The reference interpreter only rolls up the intra-group `IDX` to `rec`: https://github.com/WebAssembly/spec/blob/aed61c2b5efb8240e8d4343bf5db5409b7509874/interpreter/valid/valid.ml#L178-L185
But it doesn't substitute the cross-group IDX references, which then leak through the roll and remain in the context returned.

The most plausible fix seems to be adding a substitution after the roll at `check_rectype` -- so something like this:
```ocaml
let check_rectype (c : context) (rt : rectype) at : context =
  let RecT sts = rt in
  let x = Lib.List32.length c.types in
  let rolled = roll_deftypes x rt in
  let closed = List.map (subst_deftype (subst_of c.types)) rolled in
  let c' = {c with types = c.types @ closed} in
  List.iter (fun st -> check_subtype c' st at) sts;
  Lib.List32.iteri
    (fun i st -> check_subtype_sub c' st (Int32.add x i) at) sts;
  c'
```
This will maintain an invariant that `C.types` is IDX-free, which is quite a strong invariant, and will limit the usefulness of the matching functions (they are only correct wrt IDX-free C.types); but it seems to be required if `match_deftype` wants to use a shallow substitution only. This is relatively cheap as the cost is one-off during validation. The other option is to change `match_deftype` to use `clos_deftypes` instead of the current shallow substitution, but that is quite inefficient as the function can be called at runtime (and also due to mutual recursion with match_heaptype).

For curiosity, I've also tested wasmtime's interpreter, which correctly accepts the above module by canonicalizing all types at validation. My original matching function was pre-closing `C.types` before doing the check and hence avoided this; I found out about this when I was trying to drop the pre-closing phase.

rossberg:
Fix #2166.

@raoxiaojia, PTAL.

### DIFF
--- a/interpreter/valid/valid.ml
+++ b/interpreter/valid/valid.ml
@@ -178,12 +178,15 @@ let check_subtype_sub (c : context) (sut : subtype) x at =
 let check_rectype (c : context) (rt : rectype) at : context =
   let RecT sts = rt in
   let x = Lib.List32.length c.types in
-  let c' = {c with types = c.types @ roll_deftypes x rt} in
+  let dt =
+    try List.map (subst_deftype (subst_of c.types)) (roll_deftypes x rt)
+    with UnknownIndex x -> [type_ c (x @@ at)]  (* force error *)
+  in
+  let c' = {c with types = c.types @ dt} in
   List.iter (fun st -> check_subtype c' st at) sts;
   Lib.List32.iteri
     (fun i st -> check_subtype_sub c' st (Int32.add x i) at) sts;
   c'
-
 
 let check_tagtype (c : context) (tt : tagtype) at =
   let TagT ut = tt in
--- a/interpreter/valid/valid.ml
+++ b/interpreter/valid/valid.ml
@@ -178,11 +178,11 @@ let check_subtype_sub (c : context) (sut : subtype) x at =
 let check_rectype (c : context) (rt : rectype) at : context =
   let RecT sts = rt in
   let x = Lib.List32.length c.types in
-  let dt =
+  let dts =
     try List.map (subst_deftype (subst_of c.types)) (roll_deftypes x rt)
     with UnknownIndex x -> [type_ c (x @@ at)]  (* force error *)
   in
-  let c' = {c with types = c.types @ dt} in
+  let c' = {c with types = c.types @ dts} in
   List.iter (fun st -> check_subtype c' st at) sts;
   Lib.List32.iteri
     (fun i st -> check_subtype_sub c' st (Int32.add x i) at) sts;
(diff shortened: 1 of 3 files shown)


interpreter/valid/valid.ml
```diff
@@ -178,13 +178,16 @@ let check_subtype_sub (c : context) (sut : subtype) x at =
 let check_rectype (c : context) (rt : rectype) at : context =
   let RecT sts = rt in
   let x = Lib.List32.length c.types in
-  let c' = {c with types = c.types @ roll_deftypes x rt} in
+  let dt =
```
raoxiaojia: Technically `dts` to match the style of plurality in other places, but it doesn't matter much.

raoxiaojia (approved):

raoxiaojia:
lgtm!

interpreter/valid/valid.ml
rossberg: Indeed, fixed.

PR #2169, opened 2026-05-19
[spectec] IL semantics: be sufficiently eager to not drop failures

rossberg:
@zilinc, PTAL.


zilinc:
Thanks for the fix. LGTM. 
In [rule Step_exp/UPD-ROOT](https://github.com/WebAssembly/spec/blob/14fe8be7085c07dcb814d7f2b9afa092eb1408bd/spectec/doc/semantics/il/5-reduction.spectec#L381), do we need to restrict on `e_1`?

rossberg:
Yes, fixed!

PR #2170, opened 2026-05-21
[js-api] Migrate js-string builtin set to js/string

guybedford:
This incorporates the naming change as agreed on in the 19/05/2026 CG meeting to update JS builtins to use the `wasm:js/` prefix, by updating the string builtin name.

For backwards compatibility, the old `js-string` for `wasm:js-string` builtin remains specified and tested, to allow a guaranteed deprecation path.

With regards to guidelines about the `wasm:js/` convention, at the very least setting the precedent here seems to be the first step, suggestions with respect to the second welcome further too.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1910,7 +1910,7 @@ To <dfn>validate an import for builtins</dfn> with |import|, enabled builtins |b
 
 <h3 id="builtins-js-string">String Builtins</h3>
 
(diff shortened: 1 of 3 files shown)


eqrion (approved):
We probably need to wait for https://github.com/WebAssembly/design/issues/1584 to get consensus. My recommendation from that is to use `wasm:js/String` (uppercased string to match the ECMA global property) instead of `wasm:js/string`. Other than that, this LGTM.

PR #2017, opened 2025-10-19
[test][harness] Add table64 to the JS spectest definitions

cqundefine:
It's already present in spec interpreter and the lack of it causes some WPT tests to fail.

PR #2172, opened 2026-05-22
[test][harness] Fix table64 in the JS spectest definitions

Ms2ger:
Followup from #2017. The values need to be BigInts in the i64 case.


Ms2ger:
CC @cqundefine

eqrion (approved):

PR #2173, opened 2026-05-25
Js builtin naming

daxpedda:
Following https://github.com/WebAssembly/design/issues/1584 this PR renames `wasm:text-encoding` to `wasm:js/TextEncoder`.

This is not entirely resolved yet:
- Its currently unclear if we want this proposal to introduce two namespaces or one. Despite the discussion in https://github.com/WebAssembly/design/issues/1584, this proposal doesn't actually introduce `wasm:text-decoding`.
- Maybe we should go with the name of the actual JS spec: `wasm:js/Encoding`.

Resolves #4.


daxpedda:
So sorry, filed to the wrong repository.

PR #2174, opened 2026-05-26
[spec] Compact subtyping check in branch rules (WIP)

rossberg:
Remove redundant subtyping premise  in cast execution rule, as discussed in #2171.

@f52985, @raoxiaojia, this breaks the implementation of Ref_ok in the AL interpreter, so I could use help from either of you to adapt that.

Issue #2171, opened 2026-05-21
[spectec] Reftype subtyping premises in opsem


raoxiaojia:
I noticed that the reference instructions allow an additional `reftype_sub` at their premises, e.g.: https://github.com/WebAssembly/spec/blob/d34a973e70ab127476b5d8591c6f558be289b929/specification/wasm-3.0/4.3-execution.instructions.spectec#L149-153. But `Ref_ok` already has a `sub` rule that allows the same subtyping. 

It seems that the idea is to making subtyping the responsibility of `Ref_ok`, as I see that other uses of `Ref_ok` such as `elem/tableinst_ok` no longer has a separate subtyping step. There are also inlined TODO comments suggesting the version without the rules, which seems fine to me. 

I wonder if the extra subtyping premise in the opsem rules can be dropped -- unless there's a pass in spectec that requires them to be explicitly present in the rules?

rossberg:
Yeah, IIRC, some of the animation currently depends on it. Is that right, @zilinc, @f52985?

zilinc:
> Yeah, IIRC, some of the animation currently depends on it. Is that right, [@zilinc](https://github.com/zilinc), [@f52985](https://github.com/f52985)?

I learnt it from you, and my understanding is that some part of AL depends on it. IL animation doesn't need it though. In fact, it's easier without it. Currently, depending on which way it's written (a single `Ref_ok` vs `Ref_ok` + `Reftype_sub`), it flips the mode of the `Ref_ok` rule, which is a bit annoying.

f52985:
Yes, AL depends on it.
Here is the AL for the corresponding rule:
```
...
4. Let rt be $Ref_ok(ref).
5. If rt matches $inst_reftype(f.MODULE, rt_2), then:
  a. ...
6. Else:
  a. ...
...
```

Backend-interpreter manually defines `$Ref_ok` as a function that returns the most specific type for the given ref, so  it requires the explicit variable that will store the `return` of the function.

I think it might be possible to refactor this behavior by making `$Ref_ok` as a predicate function that requires a ref and a reftype, and returns a boolean value to be directly used in the `if` instruction.

raoxiaojia:
> Backend-interpreter manually defines $Ref_ok as a function that returns the most specific type for the given ref, so it requires the explicit variable that will store the return of the function.
> 
[...]

PR #2175, opened 2026-05-26
[spectec] AL's Ref_ok fix for redundant side-condition removal

raoxiaojia:
Refactored ref_ok for #2171 and #2174 . Also updated test expects that were not fixed yet in #2174.

@rossberg I don't have write access so created a new PR here based on #2174, as the standalone ref_ok fix doesn't work without the spec rule change due to argument number mismatch. It's fine if you prefer to cherry-pick 7ab34d5 to your other branch in case you have some work in progress there.


rossberg (approved):
Looks good, thanks!

PR #2176, opened 2026-05-26
Improve test coverage around data.drop

bvisness:
There were very few tests around data.drop, which can be strange to handle in engines due to bounds check behavior. In particular, this adds a test that would have caught a minor bounds check bug in SpiderMonkey.


rossberg (approved):
That's good, thanks!

If you feel motivated, you could add the analogous tests for elem.drop. :)

bvisness:
Adding analogous tests for tables / element segments is a bit wacky because commit a21b0df95 ignored the big DO NOT EDIT at the top of table_init.wast. I had to rework the generator slightly to introduce that test there (which is fine; it's a good test).

As a warning, though, I have another bigger patch coming, because if you step back and look at a lot of what is being added in this patch, you will notice that it utterly fails to test table64, because almost nothing in that file actually touches table64 at all for some reason!! So this PR adds the correct shape of tests for "consistency" but that will be remedied in a much bigger and uglier follow-up.

rossberg:
Cool, thanks!

PR #2177, opened 2026-05-26
Actually test table64 and memory64 in *64.wast files

bvisness:
Incredibly, there were many cases in our memory64 tests where we simply did not test memory64 or table64. These were likely copy-paste errors, but magnified due to the errors being in test generators.

This is the follow-up patch mentioned in #2176. Please enjoy.


test/meta/generate_table_init.js
```diff
@@ -3,6 +3,9 @@
 
 print_origin("generate_table_init.js");
 
+const addrtype = INDEX_TYPE;
```
sbc100, line 6: Why do we need the additional `addrtype` here?   Can we just use `INDEX_TYPE` throughout?

sbc100 (commented):
Oops, was this an oversight when we (I) first created the tests?  or did this happen when we split the memory64 tests into their own subdirectory?

test/meta/generate_table_init.js
bvisness: There's no good reason except that this is copying from analogous memory tests, which had `memtype` and `decltype` (which basically differ only by a leading space). For the new table ones I opted to just use the name `addrtype` for a small amount of extra spec consistency. But it's true that a new variable is not strictly necessary here.

Ahamed009: 1.

PR #2178, opened 2026-05-27
[spec] Externaddr ok guard

raoxiaojia:
Externaddr_ok/sub needs a similar RHS validity guard as `Ref_ok/sub` (which we forgot in resolving #2171 ).


rossberg (approved):
Ah, yes!

PR #2180, opened 2026-06-04
[spec] Typing rule fix for `handler`

raoxiaojia:
The effect handler admin instruction currently allows its body to have non-empty consumed types, unlike `Label`/`Frame` which correctly assert that the consumed types must be empty. 

This causes (incorrectly) valid expressions with no consumed types to get stuck in execution, e.g. `[i32.const 1; Handler {catch*} [drop] ]`.


rossberg (approved):
Ah, thanks!

PR #2181, opened 2026-06-04
[spectec] Handle shadowing in IL semantics

rossberg:
@zilinc, addresses https://github.com/Wasm-DSL/spectec/issues/236 by using a lookup function that works with shadowing.


zilinc:
Thanks. LGTM.

PR #2183, opened 2026-06-08
[spec] Tightened resulttype validity guard context for `frame` blocks

### LINKED ISSUE #2182: [spec] Missing/over-lenient context for resulttype validity for admin `frame` block
In the typing rule of `frame`, the current two different versions of it (formulated slightly differently in the [spectec source](https://github.com/WebAssembly/spec/blob/7a366e1531e6aab089c9a07ae5608909e187bb13/specification/wasm-3.0/7.1-soundness.configurations.spectec#L23) and spec's soundness appendix) are both slightly too lenient (copied below).
```
rule Instr_ok2/frame:
  s; C |- FRAME_ n `{f} instr* : eps -> t^n
  -- Frame_ok: s |- f : C'
  -- Expr_ok2: s; C' |- instr* : t^n
```

Image

The spectec version doesn't check the validity of the result type `t^n`. Note that this does not outright allow ill-formed resulttype `t^n`, because each individual `instr_ok` does carry a guard on the validity of the produced type. 

The actual issue is in the lifting of the body type to the `frame` block itself, which used to be fine in pre-3.0 but is too lenient now, since `t^n` can carry context-aware indices such as `IDX` that can mean different things inside the body's context (C') and the outer context (C). This allows a closed struct reference to be re-described as an `IDX` type under the inner context, smuggled across the frame, and then re-interpreted as a different type under the outer context.

The prose version does the validity check, but it checks `t^n` under `C`, which does not prevent the above issue. Adding a validity check for `t^n` under the inner context `C'` doesn't help either, as the same smuggling goes through.

imo the fix should be to require `resulttype_ok: {} |- t^n : OK` under an empty context, which forces the body to type itself with a closed type. This seems correct design-wise as well, as a function's result type should be context-agnostic when used by other entities.

(I didn't create a PR as I didn't actually find where to fix the prose only.)

raoxiaojia:
Closes #2182 .

I also realised that I didn't fix the soundness prose in #2180 (didn't know it was not automatically populated back then), so I fixed in a separate commit here.

### DIFF
--- a/specification/wasm-3.0/7.1-soundness.configurations.spectec
+++ b/specification/wasm-3.0/7.1-soundness.configurations.spectec
@@ -24,7 +24,7 @@ rule Instr_ok2/frame:
   s; C |- FRAME_ n `{f} instr* : eps -> t^n
   -- Frame_ok: s |- f : C'
   -- Expr_ok2: s; C' |- instr* : t^n
-  -- Resulttype_ok: {} |- t^n: OK
+  -- Resulttype_ok: {} |- t^n : OK
 
 rule Instr_ok2/handler:
   s; C |- HANDLER_ n `{catch*} instr* : eps -> t*
--- a/specification/wasm-3.0/7.1-soundness.configurations.spectec
+++ b/specification/wasm-3.0/7.1-soundness.configurations.spectec
@@ -24,6 +24,7 @@ rule Instr_ok2/frame:
   s; C |- FRAME_ n `{f} instr* : eps -> t^n
   -- Frame_ok: s |- f : C'
   -- Expr_ok2: s; C' |- instr* : t^n
+  -- Resulttype_ok: {} |- t^n: OK
 
 rule Instr_ok2/handler:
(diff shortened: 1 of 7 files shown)


specification/wasm-3.0/7.1-soundness.configurations.spectec
```diff
@@ -24,6 +24,7 @@ rule Instr_ok2/frame:
   s; C |- FRAME_ n `{f} instr* : eps -> t^n
   -- Frame_ok: s |- f : C'
   -- Expr_ok2: s; C' |- instr* : t^n
+  -- Resulttype_ok: {} |- t^n: OK
```
rossberg:
```suggestion
  -- Resulttype_ok: {} |- t^n : OK
```

rossberg (approved):
Thanks. LGTM modulo nit.

specification/wasm-3.0/7.1-soundness.configurations.spectec
raoxiaojia: Thanks, fixed.

PR #2186, opened 2026-06-10
[spec] Fixed typo in prose of `is_packtype` showing the negation of the intended premise

raoxiaojia:
`is_packtype` is currently rendered to display the equality between the original type and its unpacked version, instead of the intended inequality.


rossberg (approved):
Thanks.

PR #2188, opened 2026-06-15
[spectec] Add generalised typing rules for call_ref soundness

### LINKED ISSUE #2187: [spec] Missing admin lifting of `call_ref` and `return_call_ref`
`call_ref`/`return_call_ref` (3.0's `invoke`/`return_inovke`) seem to be missing a lift to `instr_ok2` that works for general typeuses (in particular, deftypes) instead of just `idx`. This is a soundness gap only, since deftypes can only appear as their arguments during execution (via funcinst lookups), and the binary format only contains encodings for `idx` arguments.

The minimally sufficient fix is to add a rule for those 2 instructions carrying deftypes, since `rec n` genuinely cannot appear in the current opsem rules (I think). So something like this should work:

```
rule Instr_ok2/call_ref:
  s; C |- CALL_REF dt : t_1* (REF NULL dt) -> t_2*
  -- Expand: dt ~~ FUNC t_1* -> t_2*

rule Instr_ok2/return_call_ref:
  s; C |- RETURN_CALL_REF dt : t_3* t_1* (REF NULL dt) -> t_4*
  -- Expand: dt ~~ FUNC t_1* -> t_2*
  -- if C.RETURN = (t'_2*)
  -- Resulttype_sub: C |- t_2* <: t'_2*
  -- Instrtype_ok: C |- t_3* -> t_4* : OK
```
(I noticed that `$expanddt` was removed at some point, and all expansions now use the `~~` syntax instead).

Generalising it for general typeuses is also possible, but there seems to be no need for that.

rossberg:
Fixes #2187.  @raoxiaojia, PTAL.

### DIFF
--- a/specification/wasm-3.0/7.1-soundness.configurations.spectec
+++ b/specification/wasm-3.0/7.1-soundness.configurations.spectec
@@ -13,11 +13,13 @@ rule Instr_ok2/plain:
 
 rule Instr_ok2/call_ref:
   s; C |- CALL_REF yy : t_1* (REF NULL yy) -> t_2*
+  -- Typeuse_ok: C |- yy : OK
   -- Expand_use: yy ~~_C FUNC t_1* -> t_2*
 
 ;; TODO(3, rossberg): enable t_2* <: C.RETURN
 rule Instr_ok2/return_call_ref:
   s; C |- RETURN_CALL_REF yy : t_3* t_1* (REF NULL yy) -> t_4*
+  -- Typeuse_ok: C |- yy : OK
   -- Expand_use: yy ~~_C FUNC t_1* -> t_2*
   -- if C.RETURN = (t'_2*)
   -- Resulttype_sub: C |- t_2* <: t'_2*
--- a/specification/wasm-3.0/7.1-soundness.configurations.spectec
+++ b/specification/wasm-3.0/7.1-soundness.configurations.spectec
@@ -11,6 +11,18 @@ rule Instr_ok2/plain:
   s; C |- instr : t_1* ->_(x*) t_2*
   -- Instr_ok: C |- instr : t_1* ->_(x*) t_2*
 
+rule Instr_ok2/call_ref:
+  s; C |- CALL_REF yy : t_1* (REF NULL yy) -> t_2*
+  -- Expand_use: yy ~~_C FUNC t_1* -> t_2*
+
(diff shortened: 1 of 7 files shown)


raoxiaojia:
Thanks! In fact, I realised later yesterday that we might need a validity guard on the typeuse argument, else `call_ref` can be valid while carrying an invalid inline deftype. This has some annoying consequences (most notably instrs_ok2 no longer implies instrtype_ok, even under context_ok). Putting the validity guard in is certainly fine, but I'm studying if progress still holds without the validity guard (so we don't add useless guards, which seems to be the design spirit of the rest of the spec).

The original `IDX`-only rule was fine because it sourced the type signature from the context, so it's the same kind of issue as #2141; but for this instance not putting the guard in might be fine, because `call_ref`'s signature doesn't seem to matter for execution except for the arity.

I'll post again when I have a conclusive answer either way.

(also, this fixes #2187 instead of #2178)

rossberg:
Ah yes, good point, we definitely want to guard it, otherwise it violates the assumed preconditions of some of the premise judgements.

raoxiaojia:
> Ah yes, good point, we definitely want to guard it, otherwise it violates the assumed preconditions of some of the premise judgements.

That would be a welcomed addition for the proofs!

A note for the spectec proof backend on this: this allows two possible derivations of `instr_ok2` for `call_ref`/`return_call_ref`, which makes the derivation tree non-unique (which is not a problem since they agree).

raoxiaojia (approved):

rossberg:
Added WF premises, PTAL.

Yes, I'm aware that this allows two derivations. But as you saw, it shouldn't matter. (Previously I considered generalising the existing Instr_ok rules, but then "soundness-only" stuff would leak into the main rules.)

Maybe it's cleaner to introduce proper administrative instructions.

raoxiaojia:
> Added WF premises, PTAL.
> 
> Yes, I'm aware that this allows two derivations. But as you saw, it shouldn't matter. (Previously I considered generalising the existing Instr_ok rules, but then "soundness-only" stuff would leak into the main rules.)
> 
> Maybe it's cleaner to introduce proper administrative instructions.

lgtm!

Yes this is really because this new `invoke` now has a binary format opcode, but the current additions in this PR work correctly regardless. Perhaps this can be considered the next time when there's a clean-up of the rules and `instr/admin`.

PR #2190, opened 2026-06-17
Rename `(offset foo)` text production to `offsetexpr` (fixing broken links)

bvisness:
The text production for `(offset foo)` in data and elem segments was simply called "offset", which clashed with the identically-named `offset=foo` production used for memargs. Links in the data and elem segments sections therefore went to the wrong place. This has been resolved by renaming `(offset foo)` to "offsetexpr", symmetric with "elemexpr".


rossberg (approved):
That makes sense, thanks!

PR #2195, opened 2026-06-23
[spec] Add valtype validity guard to `Local_ok`

raoxiaojia:
The two rules for `Local_ok` is missing a validity guard for the type it quotes:
https://github.com/WebAssembly/spec/blob/869387d57e9a6f6fd65554a7593bd7e4e921787d/specification/wasm-3.0/2.4-validation.modules.spectec#L46-L53

It is used by `Func_ok` downstream later in the file, but it only delegates to `Local_ok` without any guard there either.

This seems to be missing since the guard is present for all other rules in that region, and there are also tests in the test suite that target these guards: https://github.com/WebAssembly/spec/blob/869387d57e9a6f6fd65554a7593bd7e4e921787d/test/core/ref.wast#L59-L63

(The reference interpreter correctly reject these invalid modules.)

### DIFF
--- a/specification/wasm-3.0/2.4-validation.modules.spectec
+++ b/specification/wasm-3.0/2.4-validation.modules.spectec
@@ -45,10 +45,12 @@ rule Table_ok:
 
 rule Local_ok/set:
   C |- LOCAL t : SET t
+  -- Valtype_ok: C |- t : OK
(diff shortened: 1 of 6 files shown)


rossberg (approved):
Indeed, thanks!

rossberg:
There still is some out-of-date test expect.

PR #2197, opened 2026-06-23
[spec] Fix instr sequence typing rules to actually be composable

### LINKED ISSUE #2194: [spec] instrs_ok/seq's principality requirement is slightly over-restrictive
There seems to be an issue in the structural rules of `instrs_ok` that actually prevents `instrs_ok/seq` to be applied because it requires the principal typing form (i.e. `instr_ok`) for the head instruction: https://github.com/WebAssembly/spec/blob/869387d57e9a6f6fd65554a7593bd7e4e921787d/specification/wasm-3.0/2.3-validation.instructions.spectec#L613-L633

For example, `[nop; drop]` should be typeable under `[t] ->_(locs) []` (locs is irrelevant here). However, `instr_ok` only permits `nop` to be typed under `[] -> []`, and `instrs_ok/seq` strictly requires sticking the principal type of `nop` with the rest, which makes the derivation impossible. `instrs_ok/frame/sub` can't be applied preemptively as a setup step either since none of them is applicable/useful.

Note that adding an `instrtype_sub` to `instrs_ok/seq` will not resolve this issue, since `instrtype_sub` only allow subtyping instead of framing (like `instrs_ok/frame` does). imo the correct fix needs 2 changes: change `instr_ok instr_1` to `instrs_ok [instr_1]` in `instrs_ok/seq`, and then separately add a `instrs_ok/lift` that simply lifts `instr_ok` to `instrs_ok` of the same signature -- or equivalently, compress these into `instrs_ok/seq`, which will then need to handle both frame and subtyping. This permits the principal type of `[instr_1]` to be pre-weakened (to [t] -> [t]) to fit the rest of the stack.

This fix might need to be backported to the 1.0/2.0 spec as well.

rossberg:
Fixes #2194.

@raoxiaojia, PTAL.

### DIFF
--- a/specification/wasm-1.0/6-typing.spectec
+++ b/specification/wasm-1.0/6-typing.spectec
@@ -136,7 +136,7 @@ rule Instrs_ok/empty:
 
 rule Instrs_ok/instr:
   C |- instr : t_1* -> t_2*
-  -- Instr_ok: C |- instr_1 : t_1* -> t_2*
+  -- Instr_ok: C |- instr : t_1* -> t_2*
 
 rule Instrs_ok/seq:
   C |- instr_1* instr_2* : t_1* -> t_3*
--- a/specification/wasm-1.0/6-typing.spectec
+++ b/specification/wasm-1.0/6-typing.spectec
@@ -134,9 +134,13 @@ rule Expr_ok:
 rule Instrs_ok/empty:
   C |- eps : eps -> eps
 
-rule Instrs_ok/seq:
-  C |- instr_1 instr_2* : t_1* -> t_3*
+rule Instrs_ok/instr:
+  C |- instr : t_1* -> t_2*
   -- Instr_ok: C |- instr_1 : t_1* -> t_2*
+
+rule Instrs_ok/seq:
+  C |- instr_1* instr_2* : t_1* -> t_3*
+  -- Instrs_ok: C |- instr_1* : t_1* -> t_2*
   -- Instrs_ok: C |- instr_2* : t_2* -> t_3*
 
 rule Instrs_ok/frame:
--- a/document/core/valid/instructions.rst
+++ b/document/core/valid/instructions.rst
@@ -1358,22 +1358,28 @@ $${rule-prose: Instrs_ok}
 
 $${rule: Instrs_ok/empty}
 
+$${rule: Instrs_ok/instr}
+
 $${rule: Instrs_ok/seq}
 
-$${rule: {Instrs_ok/sub Instrs_ok/frame}}
+$${rule: Instrs_ok/sub}
 
 .. note::
-   In combination with the previous rule,
-   subsumption allows to compose instructions whose types would not directly fit otherwise.
+   This *subsumption rule* allows to weaken the type of an instruction sequence to a supertype,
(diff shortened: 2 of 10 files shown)


specification/wasm-1.0/6-typing.spectec
```diff
@@ -134,9 +134,13 @@ rule Expr_ok:
 rule Instrs_ok/empty:
   C |- eps : eps -> eps
 
+rule Instrs_ok/instr:
+  C |- instr : t_1* -> t_2*
+  -- Instr_ok: C |- instr : t_1* -> t_2*
```
raoxiaojia, line 139: Oh, I am amazed that this can be correctly parsed by the backends, as at first glance it really looks like a self-referential rule; but actually it says `instr_ok C instr : t1* -> ts2*` constructs `instrs_ok C [instr]`.

specification/wasm-1.0/6-typing.spectec
```diff
@@ -134,9 +134,13 @@ rule Expr_ok:
...
+
 rule Instrs_ok/seq:
-  C |- instr_1 instr_2* : t_1* -> t_3*
-  -- Instr_ok: C |- instr_1 : t_1* -> t_2*
+  C |- instr_1* instr_2* : t_1* -> t_3*
+  -- Instrs_ok: C |- instr_1* : t_1* -> t_2*
```
raoxiaojia, line 143: Hmm, this formulation will allow derivations of arbitrary lengths now, since either instruction list can be empty. I thought it would be better to preserve the original cons form of the rule and leave the general list append as a lemma (not too difficult to prove). Or did I overlook anything here?

Same for the 2.0/3.0 changes below.

raoxiaojia:
Just a quick check regarding `instrs_ok/seq` -- is it deliberate that we want to allow arbitrary list concatenations (instead of the original 'cons' style of forward composition in the previous versions)?

specification/wasm-1.0/6-typing.spectec
rossberg: Well, the frontend does the resolution. There is nothing specific to do in the after elaboration.

rossberg (commented):
Yes, it's deliberate. It's more symmetric and a better separation of concerns this way. In particular, I prefer if there is only one rule that injects from Instr_ok into Instrs_ok.

Arbitrarily long derivations are already possible without this change, since you can always throw in another redundant application of the subsumption (with it = it') or frame (with t* = eps) rule. Likewise, there generally are multiple choices where you can invoke these rules when they're necessary, so a unique shortest derivation also already doesn't exist. So I don't think this formulation loses anything relevant.

raoxiaojia:
Thanks for the clarification! ~~In fact, that actually makes the proofs easier, but I just want to make sure that I'm not granted a simplification by accident~~. the other parts already look good to me as well!

Upd: actually, it's probably about equal, because forward composition is free but inversion is slightly harder (and actually a bit more than slightly harder for 3.0 due to instruction types carrying the additional local flags).

specification/wasm-1.0/6-typing.spectec
raoxiaojia: Then I'll praise the frontend to have handled this cleanly!

PR #2198, opened 2026-06-23
Merge CSP into main

fgmccabe:
This (should) reflect changes needed to incorporate phase 4 csp.

This should probably NOT just be merged, as the change seems to include notes that might not normally be part of the spec.


document/web-api/index.bs
tlively, line 1: Are the deletions here expected? It seems it would be better to edit the "Security and Privacy Considerations" section rather than remove it.

tlively (commented):
Are there expected to be any tests here?

README.md
```diff
@@ -5,6 +5,16 @@
 [![CI for specs](https://github.com/WebAssembly/spec/actions/workflows/ci-spec.yml/badge.svg)](https://github.com/WebAssembly/spec/actions/workflows/ci-spec.yml)
 [![CI for interpreter & tests](https://github.com/WebAssembly/spec/actions/workflows/ci-interpreter.yml/badge.svg)](https://github.com/WebAssembly/spec/actions/workflows/ci-interpreter.yml)
 
+# Content Security Policy Proposal for WebAssembly
```
rossberg: You'll want to undo the changes to the README.

document/web-api/index.bs
eqrion: Yes we should keep this.

document/js-api/index.bs
```diff
@@ -2204,6 +2207,18 @@ Note: ECMAScript doesn't specify any sort of behavior on out-of-memory condition
...
+<h3 id="host-ensure-can-compile-wasm-bytes">HostEnsureCanCompileWasmBytes()</h3>
+
+<dfn abstract-op>HostEnsureCanCompileWasmBytes()</dfn> is an implementation-defined abstract operation
+that allows host environments to block certain WebAssembly functions which
+compile bytes into WebAssembly code.
+
+An implementation of HostEnsureCanCompileWasmBytes may complete normally or
+abruptly. Any abrupt completions will be propagated to its callers. The default
+implementation of HostEnsureCanCompileWasmBytes is to unconditionally return an
+empty normal completion.
```
eqrion, line 2221: The concrete instantiation of this abstract operation requires a realm to be passed in [1]. I think we need to have that be an operand and explicitly provide it. We can probably just do `Let |realm| be the [=current Realm=]` in the right places.

Also CSP calls it "EnsureCSPDoesNotBlockWasmByteCompilation" not HostEnsureCanCompileWasmBytes. Not sure if that's an issue.

[1] https://w3c.github.io/webappsec-csp/#can-compile-wasm-bytes

document/js-api/index.bs
```diff
@@ -393,6 +393,8 @@ Note:
 
 <div algorithm>
   To <dfn>compile a WebAssembly module</dfn> from source bytes |bytes|, perform the following steps:
+    1. Perform [$HostEnsureCanCompileWasmBytes()$].
```
eqrion: This will happen when you call WebAssembly.validate. Is that intentional? It's also called by "asynchronously compile a WebAssembly module" in the "in parallel" section. My understanding is that's not allowed because it implies there could be racy accesses to the realm to perform the CSP check. 

I think we should remove this check here and move it to the JS-API entry points that need it. So the WebAssembly.Module constructor, WebAssembly.compile, WebAssembly.instantiate, and WebAssembly.instantiateStreaming.

eqrion:
```suggestion
    1. Perform [=?=] [$HostEnsureCanCompileWasmBytes()$].
```

I believe we can use the ReturnIfAbrupt macro here. We use it elsewhere in the spec.

eqrion (changes requested):
I'd also be interested to know if there are tests for this already in WPT.

document/js-api/index.bs
fgmccabe: If you look slightly above that link, it states that:
"WebAssembly defines the [HostEnsureCanCompileWasmBytes()](https://webassembly.github.io/content-security-policy/js-api/#host-ensure-can-compile-wasm-bytes) abstract operation which allows the host environment to block the compilation of WebAssembly sources into executable code."

The operation defined is the implementation of that abstract operation.

document/js-api/index.bs
fgmccabe: This seems like a non-trivial change. Personally, I think that it is ok to have it invoked during validation. There does not seem much to be gained by delaying the check.

fgmccabe: Seems ok to me.

document/js-api/index.bs
```diff
@@ -2204,6 +2207,18 @@ Note: ECMAScript doesn't specify any sort of behavior on out-of-memory condition
     See [Issue 879](https://github.com/WebAssembly/spec/issues/879) for further discussion.
 </div>
 
+<h2>Implementation-defined Operations</h2>
+
+<h3 id="host-ensure-can-compile-wasm-bytes">HostEnsureCanCompileWasmBytes()</h3>
+
+<dfn abstract-op>HostEnsureCanCompileWasmBytes()</dfn> is an implementation-defined abstract operation
```
Ms2ger, line 2214: I think this should be a host-defined AO, not an implementation-defined one.

document/js-api/index.bs
```diff
@@ -2204,6 +2207,18 @@ Note: ECMAScript doesn't specify any sort of behavior on out-of-memory condition
...
+that allows host environments to block certain WebAssembly functions which
+compile bytes into WebAssembly code.
+
+An implementation of HostEnsureCanCompileWasmBytes may complete normally or
+abruptly. Any abrupt completions will be propagated to its callers. The default
+implementation of HostEnsureCanCompileWasmBytes is to unconditionally return an
+empty normal completion.
```
Ms2ger, line 2222: Please add a non-normative note pointing to the CSP spec here

document/js-api/index.bs
```diff
@@ -652,6 +654,7 @@ The verification of WebAssembly type requirements is deferred to the
 <div algorithm>
   The <dfn method for="WebAssembly">instantiate(|bytes|, |importObject|, |options|)</dfn> method, when invoked, performs the following steps:
     1. Let |stableBytes| be a [=get a copy of the buffer source|copy of the bytes held by the buffer=] |bytes|.
+    1. Perform [$HostEnsureCanCompileWasmBytes()$]
```
Ms2ger, line 657:
```suggestion
    1. Perform [=?=] [$HostEnsureCanCompileWasmBytes()$]
```

document/js-api/index.bs
```diff
@@ -393,6 +393,8 @@ Note:
 
 <div algorithm>
   To <dfn>compile a WebAssembly module</dfn> from source bytes |bytes|, perform the following steps:
+    1. Perform [=?=] [$HostEnsureCanCompileWasmBytes()$].
```
Ms2ger, line 396: This algorithm is used from a parallel section in <https://webassembly.github.io/spec/js-api/#asynchronously-compile-a-webassembly-module>, so can't throw or access the realm. The check needs to happen synchronously before going "in parallel".

Ms2ger (changes requested):

fgmccabe:
There is a potential gap in the specification around the lack of a gate on instantiating modules that had been previously compiled. Closing this gap will be discussed in an upcoming meeting of the CG.

document/js-api/index.bs
fgmccabe: The solution is to handle the thrown exception in asynchronously-compile-a-webassembly-module. The reason is that that too is gated by HostEnsureCanCompileWasmBytes.

However, there is another issue that needs to be resolved: validate invokes the same operation and that is not supposed to be gated.

PR #2199, opened 2026-06-23
PR for JSPI

fgmccabe:
PR that represents merging of JSPI into main spec.


document/js-api/index.bs
```diff
@@ -913,20 +796,16 @@ which can be simultaneously referenced by multiple {{Instance}} objects. Each
...
-Immediately after a WebAssembly [=memory.grow=] |x| instruction executes, perform the following steps:
+Immediately after a WebAssembly [=memory.grow=] instruction executes, perform the following steps:
 
 <div algorithm="memory.grow">
-    1. If the top of the stack is not [=i32.const=] (−1) and the top of the stack is not [=i64.const=] (−1),
+    1. If the top of the stack is not [=i32.const=] (−1),
         1. Let |frame| be the [=current frame=].
-        1. Assert: due to validation, |frame|.[=frame/module=].[=moduleinst/memaddrs=][|x|] exists.
-        1. Let |memaddr| be the memory address |frame|.[=frame/module=].[=moduleinst/memaddrs=][|x|].
+        1. Assert: Due to validation, |frame|.[=frame/module=].[=moduleinst/memaddrs=][0] exists.
```
fgmccabe:
```suggestion
        1. Assert: due to validation, |frame|.[=frame/module=].[=moduleinst/memaddrs=][0] exists.
```

document/core/util/macros.def
```diff
@@ -3,13 +3,13 @@
...
+.. _WasmDraft: https://webassembly.github.io/js-promise-integration/core/
 
-.. |WasmIssues| replace:: |issuelink|
-.. _WasmIssues: |issuelink|
+.. |WasmIssues| replace:: https://github.com/webassembly/js-promise-integration/issues/
+.. _WasmIssues: https://github.com/webassembly/js-promise-integration/issues/
 
-.. |IEEE754| replace:: IEEE 754
+.. |IEEE754| replace:: IEEE 754-2019
 .. _IEEE754: https://ieeexplore.ieee.org/document/8766229
```
eqrion, line 13: I think we want to revert these changes.

document/core/index.rst
```diff
@@ -3,7 +3,7 @@ WebAssembly Specification
 
 .. only:: html
 
-   | Release |release|
```
eqrion, line 6: Revert this change.

README.md
```diff
@@ -1,10 +1,18 @@
-![Spectec Status](https://github.com/WebAssembly/spec/actions/workflows/ci-spectec.yml/badge.svg)
-![Interpreter Status](https://github.com/WebAssembly/spec/actions/workflows/ci-interpreter.yml/badge.svg)
-![Spec doc Status](https://github.com/WebAssembly/spec/actions/workflows/ci-spec.yml/badge.svg)
+[![Build Status](https://travis-ci.org/WebAssembly/js-promise-integration.svg?branch=main)](https://travis-ci.org/WebAssembly/js-promise-integration)
```
eqrion: Revert this change too.

document/js-api/index.bs
eqrion, line 1333: This looks like an accidental removal?

document/js-api/index.bs
```diff
@@ -141,6 +110,8 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     text: func_alloc; url: appendix/embedding.html#embed-func-alloc
     text: func_type; url: appendix/embedding.html#embed-func-type
     text: func_invoke; url: appendix/embedding.html#embed-func-invoke
+    text: evaluation_suspend; url: appendix/embedding.html#embed-evaluation-suspend
+    text: evaluation_resume; url: appendix/embedding.html#embed-evaluation-resume
```
eqrion: These are broken links. Did you intend to use them?

document/js-api/index.bs
```diff
@@ -399,126 +354,76 @@ Note:
...
+        1. If [=Type=](|o|) is not Object, throw a {{TypeError}} exception.
         1. Let |v| be [=?=] [$Get$](|o|, |componentName|).
-        1. If |externtype| is of the form [=external-type/func=] |functype|,
-            1. If [$IsCallable$](|v|) is false, throw a {{LinkError}} exception.
-            1. If |v| has a \[[FunctionAddress]] internal slot, and therefore is an [=Exported Function=],
-                1. Let |funcaddr| be the value of |v|'s \[[FunctionAddress]] internal slot.
-            1. Otherwise,
-                1. [=Create a host function=] from |v| and |functype|, and let |funcaddr| be the result.
-                1. Let |index| be the number of external functions in |imports|. This value |index| is known as the <dfn>index of the host function</dfn> |funcaddr|.
+        1. If |externtype| is of the form [=func=] |functype|,
```
eqrion:
```suggestion
        1. If |externtype| is of the form [=external-type/func=] |functype|,
```

document/js-api/index.bs
```diff
@@ -1350,58 +1241,63 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not
...
-        1. Let |result| be the result of [=run a host function|running a host function=] from |func|, |functype|, and |arguments|.
+        1. Let [|parameters|] → [|resultTypes|] be |functype|.
+        1. Let |jsArguments| be the result of [=coerce WebAssembly arguments=](|parameters|,|arguments|).
+        1. Let |result| be the result of Completion([$Call$](|func|, undefined, |jsArguments|)).
         1. [=Clean up after running a callback=] with |stored settings|.
         1. [=Clean up after running script=] with |relevant settings|.
         1. Assert: |result|.\[[Type]] is <emu-const>throw</emu-const> or <emu-const>normal</emu-const>.
         1. Let |store| be the [=surrounding agent=]'s [=associated store=].
+        1. If |result|.\[[Type]] is <emu-const>normal</emu-const>, then:
+            1. Return the result of performing [=coerce a JavaScript return=] on |resultTypes| and |ret|.
```
eqrion: |ret| is not defined. Did you mean |result|.[[Type]]? Also not sure it matters, but you're passing the args in a different order than defined in the op.

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+        1. [=Clean up after running script=] with |relevant settings|.
+        1. Assert: |ret|.\[[Type]] is <emu-const>throw</emu-const> or <emu-const>normal</emu-const>.
+        1. If |ret|.\[[Type]] is <emu-const>throw</emu-const>, then:
+            1. Let |type|, |payload| and |opaqueData| be the result of [=coerce a JavaScript exception|coercing the JavaScript exception=] |ret|.\[[Value]].
+            1. [=WebAssembly/Throw=] with |type|, |payload| and |opaqueData|.
+        1. Otherwise, if [=list/size=] of |ret| is 1 and [$IsPromise$](|ret|.\[[Value]][0]):
+            1. Let |promise| be |ret|.\[[Value]][0].
+            1. Set the entry for |async_context| in |map| to [=paused=].
+            1. Let |awaitResult| be the result of performing [$Completion$]([$Await$](|promise|)).
+            1. Note: We only invoke [$Await$] if the call to |func| has returned a Promise object.
```
eqrion: Didn't we change this to unconditionally do Promise.resolve?

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+            1. Note: This will suspend both this algorithm, and the WebAssembly function being invoked by the [=evaluate a Promising function=] algorithm. On return, |ret| will be either a normal completion or a throw completion.
+            1. If the entry for |async_context| in |map| is not [=paused=] then:
+                1. Perform [=throw a JavaScript exception=] with a {{SuspendError}}.
+            1. Otherwise, set the entry to [=active=].
+            1. If |awaitResult|.\[[Type]] is <emu-const>throw</emu-const>, then:
+               1. Let |type|, |payload| and |opaqueData| be the result of [=coerce a JavaScript exception|coercing the JavaScript exception=] |ret|.\[[Value]].
```
eqrion: I think we want to throw the |awaitResult|.[[Value]] not the |ret|.[[Value]]. ret is the original promise the wrapped function returned. The note on L1496 has the same typo.

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+               1. [=WebAssembly/Throw=] with |type|, |payload| and |opaqueData|.
```
eqrion: Where is WebAssembly/Throw defined? I see we have a `throw a JavaScript exception` operation which boils down to:

```
Execute the WebAssembly instructions ([=ref.exn=] |address|) ([=throw_ref=]).
```

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+In addition, the {{promising}} function takes as argument a WebAssembly function and returns a JavaScript function that returns a Promise that is resolved when the WebAssembly function completes.
+
+Each [=agent=] maintains a [=Execution Context Status map=], mapping from [=execution context|execution contexts=] to a status symbol. The purpose of this map is to ensure that applications do not try to suspend JavaScript frames and also to ensure that calls to imports marked with a {{Suspending}} marker object are properly balanced by corresponding uses of {{WebAssembly.promising}}.
+
+If present, a status can be one of two <dfn>stack status</dfn> values:
+
+   * <dfn>active</dfn>. This signals that the associated execution context is actively executing and has the potential to be [=paused=].
+   * <dfn>paused</dfn>. This signals that the associated execution is not currently involved in computation and has been paused.
+
+If an execution context is not present in the status mapping, then it may not be [=Pause|paused=] or [=Enter|reentered=].
```
eqrion:
```suggestion
If an execution context is not present in the status mapping, then it may not be [=Pause|paused=] or [=Active|active=].
```

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+        1. Let |relevant settings| be |realm|'s [=realm/settings object=].
+        1. Let |async_context| be the [=surrounding agent=]'s [=running execution context=].
+        1. Let |map| be the [=surrounding agent=]'s associated [=Execution Context Status map=].
+        1. If the entry for |async_context| in |map| is not [=active=], then:
+            1. Perform [=throw a JavaScript exception=] with a {{SuspendError}} exception.
+        1. [=Prepare to run script=] with |relevant settings|.
+        1. [=Prepare to run a callback=] with |stored settings|.
+        1. Let [|parameters|] → [|resultTypes|] be |functype|.
+        1. Let |jsArguments| be the result of [=coerce WebAssembly arguments=](|parameters|,|arguments|).
+        1. Let |ret| be [$Completion$]([$Call$](|func|, undefined, |jsArguments|)).
```
eqrion, line 1630: Elsewhere we do just `Completion(..` without the $. Which is correct?

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
-    1. If |n| &lt; 0 or |n| &gt; 2<sup>64</sup> &minus; 1, [=throw=] a {{TypeError}}.
+A {{Suspending}} marker object represents a JavaScript function whose calls via WebAssembly imports should be *suspended* when they return a Promise object.
+Each {{Suspending}} marker object has a \[[WrappedFunction]] internal slot which holds a JavaScript function.
 
-        Note: This operation is designed to emulate [=[EnforceRange]=].
-    1. Return [=ℝ=](|n|) as a WebAssembly [=u64=].
-1. Assert: This step is not reached.
+In addition, the {{promising}} function takes as argument a WebAssembly function and returns a JavaScript function that returns a Promise that is resolved when the WebAssembly function completes.
+
+Each [=agent=] maintains a [=Execution Context Status map=], mapping from [=execution context|execution contexts=] to a status symbol. The purpose of this map is to ensure that applications do not try to suspend JavaScript frames and also to ensure that calls to imports marked with a {{Suspending}} marker object are properly balanced by corresponding uses of {{WebAssembly.promising}}.
```
eqrion, line 1534:
```suggestion
Each [=agent=] maintains a [=Execution Context Status map=], mapping from [=execution context|execution contexts=] to a status symbol. The purpose of this map is to ensure that applications do not try to suspend JavaScript frames and also to ensure that calls to imports marked with a {{Suspending}} marker object are properly balanced by corresponding uses of {{promising}}.
```

I think you need to do the short form to get the auto-linking to work.

document/js-api/index.bs
```diff
@@ -24,54 +23,41 @@ Date: now
...
         text: RangeError; url: sec-native-error-types-used-in-this-standard-rangeerror
+    type: dfn;
+        text: AbstractClosure; url: sec-abstract-closure
+        text: CreateBuiltinFunction; url: sec-createbuiltinfunction
+        text: PromiseCapabilityRecord; url: sec-promisecapability-records
+        text: EvaluateCall; url: sec-evaluatecall
+        text: ExecutionContext; url: sec-execution-contexts
+        text: IsPromise; url: sec-ispromise
+        text: PerformPromiseThen; url: sec-performpromisethen
+        text: Execution Context Stack; url: execution-context-stack
```
eqrion, line 52: PerformPromiseThen and EvaluateCall are not used. You link to "execution context" in a couple places, but I'm not sure that will match to the ExecutionContext (no space) you use here.

document/js-api/index.bs
```diff
@@ -1318,14 +1195,16 @@ Note: [=call an Exported Function|Calling an Exported Function=] executes in the
 Note: Exported Functions do not have a \[[Construct]] method and thus it is not possible to call one with the `new` operator.
 
 <div algorithm>
-  To <dfn>run a host function</dfn> from the JavaScript object |func|, type |functype|, and [=list=] of [=WebAssembly values=] |arguments|, perform the following steps:
-
-    1. Let [|parameters|] → [|results|] be |functype|.
-    1. If any type in |parameters| or |results| [=matches/valtype|matches=] [=v128=] or [=exnref=], throw a {{TypeError}}.
+  To <dfn>coerce WebAssembly arguments</dfn> from a [=list=] of |parameterTypes| and a [=list=] of JavaScript arguments |arguments|, perform the following steps
+    1. If |parameterTypes| contain [=v128=], throw a {{TypeError}}.
```
eqrion: Previously we would check both param and result types for v128 and throw TypeError before actually doing any coercions. This relaxes that. We should keep the upfront check for v128.

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+    1. Let |ec| be the currently executing [=execution context=], i.e., the [=execution context=] that is at the top of the [=surrounding agent=]'s current [=execution context stack=].
+    1. Assert: |map| does not contain any entry for |ec|.
+    1. Add an entry mapping |ec| to [=active=] in |map|.
+    1. Let (|store|, |result|) be the result of [=func_invoke=](|store|, |funcaddr|, |args|).
+    1. Assert: If control reaches here, we have done waiting for suspended imports.
+    1. If the entry for |ec| in |map| is not [=active=] then throw a WebAssembly {{SuspendError}} exception. Otherwise, remove the entry for |ec| from [=map=].
+    1. Set the [=surrounding agent=]'s [=associated store=] to |store|.
+    1. If |result| is [=error=], throw a WebAssembly {{RuntimeError}} exception, unless otherwise indicated by <a href="#errors">the WebAssembly error mapping</a>.
+    1. Otherwise, if |result| is of the form [=throw=] exnaddr,
+        1. [=Reject=] |promise| with |result|.
```
eqrion: |result| here is a `throw exnaddr` abstract value, and we're rejecting a promise with it. I think we need to convert the exnaddr into a JS value first.

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+            1. Note: We only invoke [$Await$] if the call to |func| has returned a Promise object.
+            1. Note: This will suspend both this algorithm, and the WebAssembly function being invoked by the [=evaluate a Promising function=] algorithm. On return, |ret| will be either a normal completion or a throw completion.
+            1. If the entry for |async_context| in |map| is not [=paused=] then:
+                1. Perform [=throw a JavaScript exception=] with a {{SuspendError}}.
+            1. Otherwise, set the entry to [=active=].
+            1. If |awaitResult|.\[[Type]] is <emu-const>throw</emu-const>, then:
+               1. Let |type|, |payload| and |opaqueData| be the result of [=coerce a JavaScript exception|coercing the JavaScript exception=] |ret|.\[[Value]].
+               1. [=WebAssembly/Throw=] with |type|, |payload| and |opaqueData|.
+            1. Otherwise, return the result of performing [=coerce a JavaScript return=] on |resultTypes| and |awaitResult|.
+        1. Otherwise, return the result of performing [=coerce a JavaScript return=] on |resultTypes| and |ret|.
```
eqrion: Both of these pass args in the reverse order the op expects.

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+    constructor(Function jsFun);
+};
+</pre>
+
+<div algorithm>
+    The  <dfn method for="WebAssembly">promising(|wasmFunc|)</dfn> function, when invoked, performs the following steps:
+    1. If [$IsCallable$](|wasmFunc|) is false, throw a {{TypeError}}.
+    1. If |wasmFunc| does not have a \[[FunctionAddress]] internal slot, throw a {{TypeError}}.
+    1. Let |builder| be a new [=AbstractClosure=] with no parameters that captures |wasmFunc| and performs [=run a Promising function=] when called.
+    1. Returns the result of [=CreateBuiltinFunction=](|builder|,1,"", « »).
```
eqrion:
```suggestion
    1. Return the result of [=CreateBuiltinFunction=](|builder|,1,"", « »).
```

document/js-api/index.bs
```diff
@@ -1480,35 +1376,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+    1. Return UNUSED.
+</div>
+
+<div algorithm>
+    The <dfn constructor for="Suspending">Suspending(|jsFun|)</dfn> constructor, when invoked, performs the following steps:
+    1. If [$IsCallable$](|jsFun|) is false, throw a {{TypeError}}.
+    1. Let |suspendingProto| be \[[%WebAssembly.Suspending.prototype%]].
+    1. Let |susp| be the result of [$OrdinaryObjectCreate$](|suspendingProto|,«\[[WrappedFunction]]»).
+    1. Set |susp|.\[[WrappedFunction]] to |jsFun|.
+    1. Return |susp|.
```
eqrion, line 1611: WebIDL will automatically create an object for us that is wired up with the correct prototype (and other slots such as the internal platform object brand). The other WebIDL constructors in this spec just initialize the slots on the `this` object.

document/js-api/index.bs
```diff
@@ -1276,40 +1149,44 @@ This slot holds a [=function address=] relative to the [=surrounding agent=]'s [
...
-  To <dfn>call an Exported Function</dfn> with [=function address=] |funcaddr| and a [=list=] of JavaScript arguments |argValues|, perform the following steps:
-
-    1. Let |store| be the [=surrounding agent=]'s [=associated store=].
-    1. Let |functype| be [=func_type=](|store|, |funcaddr|).
+  To <dfn>coerce JavaScript arguments</dfn> from a |functype| and a [=list=] of JavaScript arguments |argValues|, perform the following steps
     1. Let [|parameters|] → [|results|] be |functype|.
-    1. If any type in |parameters| or |results| [=matches/valtype|matches=] [=v128=] or [=exnref=], throw a {{TypeError}}.
-
+    1. If |parameters| or |results| contain [=v128=] or [=exnref=], throw a {{TypeError}}.
         Note: the above error is thrown each time the \[[Call]] method is invoked.
```
eqrion, line 1286: Did you mean to delete this?

eqrion (changes requested):
This is a tricky one, and I'm still learning the spec interactions for WHATWG/TC39 so forgive me if I messed some of these comments up. I'd like to take another look at this.

document/core/util/macros.def
fgmccabe: I did the pagelink fixes. But, what about the IEEE reference?

rossberg: Please revert. This changes the inline text produced and would result in awkward wording like "integers and IEEE 754-2019 numbers". In the contexts where this macro is used, the version is irrelevant.

document/js-api/index.bs
fgmccabe: got it.

document/js-api/index.bs
fgmccabe: I believe that [$Completion$](...) is correct because it (should) sets up a hyperlink to the ES262 notion of Completion.

PR #2204, opened 2026-07-06
[spec] Fix minor error on v128.store_lane oob condition

raoxiaojia:
Fixed a minor error in v128.store_lane's oob condition (`N` is bit-width instead of byte-width).


rossberg (approved):
Thanks!

PR #2206, opened 2026-07-07
[spec] Fix `Localval_ok/unset` making `local.set` untypeable

### LINKED ISSUE #2205: [spec] `Localval_ok`'s rule for uninitialised locals prevents typing of `local.set`
`Localval_ok`'s current rule for uninitialised values (used by `Frame_ok` later) types it at `BOT`:
https://github.com/WebAssembly/spec/blob/ef4771a0b2fc1fc2f33b917640705f8e8a9a0315/specification/wasm-3.0/7.1-soundness.configurations.spectec#L341-L352

It seems that the intention is so that an uninitialised value cannot be read (prevented by the `UNSET` flag), but can match the value type that a later `local.set` would want to set. However this doesn't actually work, because `local.set x`'s typing rule is contravariant on `C.locals[x]`, and putting `BOT` as the type of local prevents writing any value to the local. Similarly putting a `top` wouldn't work (and Wasm has no top type anyway) because the value then can't be used.

I think the correct fix is to let `Localval_ok/unset` be typed at an arbitrary `UNSET t` instead of `UNSET BOT` (probably need a validity guard on the type `t` being valid as well?), basically giving the rule freedom to choose the precise type it needs for the later `get/set` operations.

raoxiaojia:
Fixes #2205 .

On the validity guards: I don't think `Localval_ok/set` needs a guard, because `store_ok s` + `s |- val : t` should be able to derive that. Only the `unset` version needs it as the type is arbitrary.

### DIFF
--- a/specification/wasm-3.0/7.1-soundness.configurations.spectec
+++ b/specification/wasm-3.0/7.1-soundness.configurations.spectec
@@ -343,7 +343,8 @@ rule Localval_ok/set:
   -- Val_ok: s |- val : t
 
 rule Localval_ok/unset:
-  s |- eps : UNSET BOT
+  s |- eps : UNSET t
+  -- Valtype_ok: {} |- t : OK
 
 rule Frame_ok:
   s |- {LOCALS (val?)*, MODULE moduleinst} : C ++ {LOCALS lct*}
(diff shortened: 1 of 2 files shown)


rossberg (approved):
Thanks!

PR #2207, opened 2026-07-09
[spec] Additional possible implementation limits

rossberg:
Some oversights (e.g., number of active handlers) and some motivated by [SpaceWasm](https://github.com/nasa/spacewasm#interpreter-limitations). I don't consider any of them controversial.


tlively (approved):

tlively:
LGTM, but I wonder if there are any limits an implementation could think up that we would push back on. Maybe we should give up on this being an exhaustive list?

rossberg:
Good question. In a way, I see this list less as a boundary for implementors, but more as a warning for producers: these are the values not to get excessive with individually, even if an overall module can still be fairly large. More importantly, it is the list of things that some ecosystems might want to make concrete, like the JS-API does.

PR #2210, opened 2026-07-15
[spec] specify host functions in spectec

zilinc:
It breaks the prose backend.
The official spec still needs to splice the relevant additions from spectec.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -401,16 +401,16 @@ Furthermore, the resulting store must be :ref:`valid <valid-store>`, i.e., all d
 
 $${rule: {Step/call_ref-hostfunc-*}}
 
-Here, :math:`\X{hf}(s, \val^n)` denotes the implementation-defined execution of host function :math:`\X{hf}` in current store :math:`s` with arguments :math:`\val^n`.
-It yields a set of possible outcomes, where each element is either a pair of a modified store :math:`s'` and a :ref:`result <syntax-result>`
-or the special value :math:`\bot` indicating divergence.
+Here, ${:$hostcall(_HOSTFUNC hf, s, val^n)} denotes the implementation-defined execution of host function ${:_HOSTFUNC hf} in current store ${:s} with arguments ${:val^n}.
+It yields a set of possible outcomes, where each element is either a pair of a modified store ${:s'} and a :ref:`result <syntax-result>`
+or the special value ${hostcallresult:BOT} indicating divergence.
 A host function is non-deterministic if there is at least one argument for which the set of outcomes is not singular.
 
 For a WebAssembly implementation to be :ref:`sound <soundness>` in the presence of host functions,
 every :ref:`host function instance <syntax-funcinst>` must be :ref:`valid <valid-hostfuncinst>`,
 which means that it adheres to suitable pre- and post-conditions:
-under a :ref:`valid store <valid-store>` :math:`s`, and given arguments :math:`\val^n` matching the ascribed parameter types :math:`t_1^n`,
-executing the host function must yield a non-empty set of possible outcomes each of which is either divergence or consists of a valid store :math:`s'` that is an :ref:`extension <extend-store>` of :math:`s` and a result matching the ascribed return types :math:`t_2^m`.
+under a :ref:`valid store <valid-store>` ${:s}, and given arguments ${:\val^n} matching the ascribed parameter types ${:t_1^n},
+executing the host function must yield a non-empty set of possible outcomes each of which is either divergence or consists of a valid store ${:s'} that is an :ref:`extension <extend-store>` of ${:s} and a result matching the ascribed return types ${:t_2^m}.
 All these notions are made precise in the :ref:`Appendix <soundness>`.
 
 .. note::
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -330,25 +330,27 @@ and returning from it.
 Invocation of :ref:`function reference <syntax-ref.func>` :math:`(\REFFUNCADDR~a)`
 ..................................................................................
 
-1. Assert: due to :ref:`validation <valid-call>`, :math:`S.\SFUNCS[a]` exists.
+1. Let :math:`z` be the current state.
 
-2. Let :math:`f` be the :ref:`function instance <syntax-funcinst>`, :math:`S.\SFUNCS[a]`.
+2. Assert: due to :ref:`validation <valid-call>`, :math:`z.\SFUNCS[a]` exists.
 
-3. Let :math:`\TFUNC~[t_1^n] \Tarrow [t_2^m]` be the :ref:`composite type <syntax-comptype>` :math:`\expanddt(\X{f}.\FITYPE)`.
+3. Let :math:`\X{fi}` be the :ref:`function instance <syntax-funcinst>`, :math:`z.\SFUNCS[a]`.
 
-4. Let :math:`\FUNC~x~\local^\ast~\instr^\ast` be the :ref:`function <syntax-func>` :math:`f.\FICODE`.
+4. Let :math:`\TFUNC~[t_1^n] \Tarrow [t_2^m]` be the :ref:`composite type <syntax-comptype>` :math:`\expanddt(\X{fi}.\FITYPE)`.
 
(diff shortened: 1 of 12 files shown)


zilinc:
@rossberg We talked about upstreaming this part, and here we go. It should be a good starting point, but there're still TODOs.

specification/wasm-3.0/4.0-execution.configurations.spectec
```diff
@@ -329,3 +329,18 @@ def $growmem(meminst, n) = meminst'
...
+syntax hostcallresult =
+  | RES store result hint(show %, %)
+  | BOT
+
+def $hostcall(hostfunc, store, val*) : hostcallresult*  hint(show %4 <- %1(%2, %3))
+def $hostcall hint(builtin)
+
+def $lift_result(result) : instr*  hint(show %)
+def $lift_result(_VALS val*) = val*
+def $lift_result(`(REF.EXN_ADDR a) THROW_REF) = (REF.EXN_ADDR a) THROW_REF
```
rossberg, line 345: Note that the representation of exception results changes in #2189.

specification/wasm-3.0/4.3-execution.instructions.spectec
```diff
@@ -170,17 +170,32 @@ rule Step_read/call:
...
   ----
   -- if $funcinst(z)[a] = fi
   -- Expand: fi.TYPE ~~ FUNC t_1^n -> t_2^m
   -- if fi.CODE = FUNC x (LOCAL t)* (instr*)
   -- if f = {LOCALS val^n ($default_(t))*, MODULE fi.MODULE}
 
+rule Step/call_ref-hostfunc-res:
+  s; f; val^n (REF.FUNC_ADDR a) (CALL_REF yy)  ~>  s'; f; $lift_result(result)
+  ----
+  -- if $funcinst((s; f))[a] = fi
```
rossberg, line 187: Does it not work to use pattern matching here, like in the hand-written version?

rossberg (commented):
Thanks, this looks good so far.

specification/wasm-3.0/4.3-execution.instructions.spectec
zilinc: I think the reason was that there's a `MODULE` field in the record as well, so it has to appear in the pattern. But otherwise I don't see any problems. I can do it if it's preferred.

rossberg: Ah, right. In that case leave it as is.

zilinc: I was wondering why I needed the parens around `(s; f)` in the `$funcinst` call. Maybe the reason is in the latex generation:
```
-\mbox{if}~ (s ; f){.}\mathsf{funcs}{}[a] = {\mathit{fi}} \\
+\mbox{if}~ s ; f{.}\mathsf{funcs}{}[a] = {\mathit{fi}} \\
```

rossberg:
Quick reminder: please never, ever force-push to a PR branch, since that breaks half of GitHub's review work flow. ;)

specification/wasm-3.0/4.3-execution.instructions.spectec
rossberg: Yes, exactly. Or more precisely, the show hint for $funcinst removes the call parens..

zilinc:
@f52985 I'm stuck on the reduction rules. I have a function call on the rhs of a reduction, which is not supported by the AL pipeline. Would you please take a look? @rossberg Do we want to generate prose for them? Currently there's no prose for the manually written rules.

rossberg:
We don't necessarily need prose for these (since they are mostly interesting as part of a formalisation). But of course, the pipeline shouldn't break over them.

f52985:
> @f52985 I'm stuck on the reduction rules. I have a function call on the rhs of a reduction, which is not supported by the AL pipeline. Would you please take a look?

https://github.com/f52985/spec/tree/hostfunc

In this branch, I implemented a fix that allows function calls—or, more generally, any expression whose type is a list of instructions—on the right-hand side of a reduction rule. It also seems to work correctly with the interpreter backend. Please check if this works, and let me know if something goes wrong.

zilinc:
> > @f52985 I'm stuck on the reduction rules. I have a function call on the rhs of a reduction, which is not supported by the AL pipeline. Would you please take a look?
> 
> https://github.com/f52985/spec/tree/hostfunc
> 
> In this branch, I implemented a fix that allows function calls—or, more generally, any expression whose type is a list of instructions—on the right-hand side of a reduction rule. It also seems to work correctly with the interpreter backend. Please check if this works, and let me know if something goes wrong.

Thanks heaps Dongjun. I'll give it a go and let you know.

zilinc:
@f52985 Thanks agian. The pipeline now works without errors. The prose merges the cases in a somewhat unexpected way, but since we are not using the prose for any of the rules here, it should be fine.

zilinc:
@rossberg This should be ready. The new rules look almost identical as before. I edited the doc a bit to reflect some meta-variable changes, and also tried to fix the manual prose for the normal function call case.

rossberg (commented):
Ah, it wasn't necessary to change the prose for the normal function case, since that's gone anyway with #2189.

But while you're converting the host function section to SpecTec, would you mind including the inline math as well? For example, ``:math:`(s, \val^n)` `` becomes `${:(s, val^n)}`.

document/core/exec/instructions.rst
```diff
@@ -397,39 +399,18 @@ A host function may also modify the :ref:`store <syntax-store>`.
...
+It yields a set of possible outcomes, where each element is either a pair of a modified store ${:s'} and a :ref:`result <syntax-result>`
+or the special value ${hostcallresult:BOT} indicating divergence.
 A host function is non-deterministic if there is at least one argument for which the set of outcomes is not singular.
 
 For a WebAssembly implementation to be :ref:`sound <soundness>` in the presence of host functions,
 every :ref:`host function instance <syntax-funcinst>` must be :ref:`valid <valid-hostfuncinst>`,
 which means that it adheres to suitable pre- and post-conditions:
-under a :ref:`valid store <valid-store>` :math:`S`, and given arguments :math:`\val^n` matching the ascribed parameter types :math:`t_1^n`,
-executing the host function must yield a non-empty set of possible outcomes each of which is either divergence or consists of a valid store :math:`S'` that is an :ref:`extension <extend-store>` of :math:`S` and a result matching the ascribed return types :math:`t_2^m`.
+under a :ref:`valid store <valid-store>` ${:s}, and given arguments ${:\val^n} matching the ascribed parameter types ${:t_1^n},
```
rossberg:
```suggestion
under a :ref:`valid store <valid-store>` ${:s}, and given arguments ${:val^n} matching the ascribed parameter types ${:t_1^n},
```

document/core/exec/instructions.rst
```diff
@@ -397,39 +399,18 @@ A host function may also modify the :ref:`store <syntax-store>`.
...
-     \wedge & \deftype \approx \TFUNC~[t_1^n] \Tarrow [t_2^m] \\
-     \wedge & \bot \in \X{hf}(S; \val^n) \\
-     \end{array} \\
-   \end{array}
+$${rule: {Step/call_ref-hostfunc-*}}
 
-Here, :math:`\X{hf}(S; \val^n)` denotes the implementation-defined execution of host function :math:`\X{hf}` in current store :math:`S` with arguments :math:`\val^n`.
-It yields a set of possible outcomes, where each element is either a pair of a modified store :math:`S'` and a :ref:`result <syntax-result>`
-or the special value :math:`\bot` indicating divergence.
+Here, ${:$hostcall(_HOSTFUNC hf, s, val^n)} denotes the implementation-defined execution of host function ${:_HOSTFUNC hf} in current store ${:s} with arguments ${:val^n}.
```
rossberg:
```suggestion
Here, ${:$hostcall(_HOSTFUNC hf, s, val^n)} denotes the implementation-defined execution of host function ${:hf} in current store ${:s} with arguments ${:val^n}.
```

rossberg (commented):
Thanks! Two minor oversights, I believe.

document/core/exec/instructions.rst
zilinc: Thanks for spotting them. The other one was an oversight; this one was intentional, but the intention was not quite correct.

rossberg (approved):

PR #2211, opened 2026-07-19
[spec] Test subsumption in return call result types

CharlieTap:
While validating a GC module in my runtime, I found that return_call and return_call_indirect were comparing result types for exact equality instead of using result type matching. This was a left over from implementing tail call before gc, and caused a certain valid fixture to fail.

This PR adds tests for both instructions. For each instruction, I've included a valid case showing that (ref array) matches (ref eq), and an invalid case showing that the reverse direction is rejected. The cases are intentionally minimal rather than repeating the broader subsumption coverage already present.

### DIFF
--- a/test/core/gc/type-subtyping.wast
+++ b/test/core/gc/type-subtyping.wast
@@ -113,6 +113,23 @@
 )
 
 (module
+  (func $f (result (ref array)) unreachable)
+  (func (result (ref eq))
+    return_call $f
+  )
+)
+
+(assert_invalid
+  (module
+    (func $f (result (ref eq)) unreachable)
+    (func (result (ref array))
+      return_call $f
+    )
+  )
+  "type mismatch"
+)
+
+(module
   (rec (type $f1 (sub (func))) (type (struct (field (ref $f1)))))
   (rec (type $f2 (sub (func))) (type (struct (field (ref $f2)))))
   (rec (type $g1 (sub $f1 (func))) (type (struct)))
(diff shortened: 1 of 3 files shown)


test/core/gc/type-subtyping.wast
```diff
@@ -399,6 +416,27 @@
 (assert_trap (invoke "fail1") "indirect call type mismatch")
 (assert_trap (invoke "fail2") "indirect call type mismatch")
 
+(module
+  (type $t (func (result (ref array))))
+  (table 0 funcref)
+  (func (result (ref eq))
+    i32.const 0
+    return_call_indirect (type $t)
```
rossberg: Nit: for consistency with the rest of the file, use folded expression style throughout.

test/core/gc/type-subtyping.wast
```diff
@@ -399,6 +416,27 @@
 (assert_trap (invoke "fail1") "indirect call type mismatch")
 (assert_trap (invoke "fail2") "indirect call type mismatch")
 
+(module
+  (type $t (func (result (ref array))))
```
rossberg: This and the following test are slightly misplaced here: this section of the file is concerned with runtime subtyping, but these two tests are about the static subtyping around the return_call_indirect instruction. I would merge them with the other new tests.

rossberg (changes requested):
Thanks!

rossberg (commented):
Sorry for the churn, but on second thought, these tests belong in the respective test files for `return_call` and `return_call_indirect` (and `return_call_ref`, for that matter). The subtyping.wast file is meant to test the correctness of the subtyping algorithm itself (for GC types), not whether it is applied in all the right places.

Since the respective test files are not under `gc/`, though, it would be good to avoid the use of GC types, e.g., use `(ref $functype)` vs `funcref` instead, which should work independent of the presence of GC.

CharlieTap:
No worries thats actually a nicer solution 👍🏼

rossberg (commented):
Thanks! Can you also add a respective test for return_call_ref?

CharlieTap:
Its seems to already have these covered in slightly different examples:

https://github.com/WebAssembly/spec/blob/025bcbf393f51e6e5feac6767413405ca47a0b35/test/core/return_call_ref.wast#L226

and

https://github.com/WebAssembly/spec/blob/025bcbf393f51e6e5feac6767413405ca47a0b35/test/core/return_call_ref.wast#L280

rossberg (approved):
Ah, fair enough!

PR #2212, opened 2026-07-23
[spec] a missing condition in Instr_ok2/frame

zilinc:
This is uncovered in Wasm 2.0's soundness appendix when @maximelegoupil works on the type soundness proof. Back porting it to Wasm 3.0.

### DIFF
--- a/specification/wasm-3.0/7.1-soundness.configurations.spectec
+++ b/specification/wasm-3.0/7.1-soundness.configurations.spectec
@@ -37,6 +37,7 @@ rule Instr_ok2/label:
 rule Instr_ok2/frame:
   s; C |- FRAME_ n `{f} instr* : eps -> t^n
   -- Frame_ok: s |- f : C'
+  -- if C'.RETURN = t^n
   -- Expr_ok2: s; C' |- instr* : t^n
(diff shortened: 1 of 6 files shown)


specification/wasm-3.0/7.1-soundness.configurations.spectec
```diff
@@ -37,6 +37,7 @@ rule Instr_ok2/label:
 rule Instr_ok2/frame:
   s; C |- FRAME_ n `{f} instr* : eps -> t^n
   -- Frame_ok: s |- f : C'
+  -- if C'.RETURN = t^n
   -- Expr_ok2: s; C' |- instr* : t^n
   -- Resulttype_ok: {} |- t^n : OK
```
rossberg, line 42: Ah, yes. At the same time, this line should actually be redundant, shouldn't it? Can you try to remove it?

zilinc: Is that what #2183 was trying to fix?

rossberg: Ah, you're right! :)

rossberg (approved):

PR #2214, opened 2026-07-27
[interpreter] Tweak parsing of simd shuffle indices

### LINKED ISSUE #2209: some scripts expect an "unknown operator" for negative alignment
Hi,

I'm currently adding SIMD support in Owi. I'm using the reference test suite, which contains some test cases where a `"unknown operator"` error is expected when a module contains `v128.load align=-1` (a negative alignment). Is it on purpose? Wouldn't it be better to switch to "alignment must be positive" or something similar?

rossberg:
Fix #2209.

@redianthus, PTAL.


redianthus (approved):

redianthus:
Yes, that's exactly what I expected. Thanks for taking care of it!

PR #2216, opened 2026-07-27
Rename "scope" as "environment" in explainer and add binary encoding

titzer:


titzer:
Sorry for the noise, wrong upstream branch.

PR #2223, opened 2026-08-04
[spec test] results type must exact match in the return call

KKiiim:
required by spec. Results types of callee and caller in return call must be an exact match


KKiiim:
@eqrion Hi, could you help to take a look. Thanks!

rossberg (approved):
Makes sense.

PR #2224, opened 2026-08-10
Merge wide-arithmetic proposal into the specification

alexcrichton:
This commit is an attempt of mine to merge the [wide-arithmetic proposal](https://github.com/WebAssembly/wide-arithmetic/) into the main specification here. I'm not actually entirely sure if this is something I should do or defer to someone else, but I figure I can at least make a skeleton here! I've done a merge from the spec's current status back into wide-arithmetic, sorted out things like CI and such, and now this is merging back into the main specification. This is intended to preserve the git history but I can also squash thing down if that's additionally desired.

I'll note that I think all the spec changes are largely unreviewed, so I'm happy to adjust things as necessary. I'd request special attention to the `TEST.md` files as I suspect I messed up something there, but my belief is that this should otherwise largely be complete.


document/core/text/instructions.rst
```diff
@@ -297,6 +297,17 @@ $${grammar: {
...
+.. _text-wideop:
+
+.. math::
+   \begin{array}{llclll}
+   \phantom{\production{instruction}} & \phantom{\Tplaininstr_I} &\phantom{::=}& \phantom{thisisenough} && \phantom{thisshouldbeenough} \\[-2ex] &&|&
+     \text{i64.add128} &\Rightarrow& \I64.\ADD128 \\ &&|&
+     \text{i64.sub128} &\Rightarrow& \I64.\SUB128 \\ &&|&
+     \text{i64.mul\_wide\_s} &\Rightarrow& \I64.\MULWIDE\K{\_s} \\ &&|&
+     \text{i64.mul\_wide\_u} &\Rightarrow& \I64.\MULWIDE\K{\_u} \\
+   \end{array}
```
tlively: The text format should be doable in spectec now.

spectec/test-interpreter/TEST.md
```diff
@@ -134,5 +134,8 @@ spectec 0.5 generator
 - print_i32: 1
 - print_i32: 2
 - print: ()
+../test-interpreter/spec-test-3/wide-arithmetic.wast:25.1-28.44
+- Test failed at ../test-interpreter/spec-test-3/wide-arithmetic.wast:25.1-28.44 (Algorithm not found: WIDEOP (interpreting CaseV(WIDEOP, [CaseV(I64, []), CaseV(ADD128, [])]) at ))
+Test failed for ../test-interpreter/spec-test-3/wide-arithmetic.wast
```
tlively, line 139: I believe this shows the meta-interpreter failing. We probably want to fix that before merging (although the meta-interpreter is not formally part of the standards process).

alexcrichton: I might need some help on this one and the other failure in CI as well. One failure on CI is [this one](https://github.com/WebAssembly/spec/actions/runs/31426130688/job/93578157105) which asserts that wasm-3.0 is the same as wasm-latest where I believe that this PR is the first divergence from that.

This failure I believe is due to the fact that `spectec/test-interpreter/spec-test-3` is a symlink to `../../test/core` which is also no longer valid since this is the first proposal to go beyond Wasm 3.0.

Should I disable the "ensure-wasm-latest" CI check? Should I copy the testsuite before-this-PR to spec-test-3 in what looks like a similar manner of how spec-test-{1,2} were copied?

tlively: > Should I disable the "ensure-wasm-latest" CI check?

I guess if we don't intend wide-arithmetic to be part of Wasm 3.0, then we will want to create a new wasm-3.1 (or wasm-4.0???) directory that can start tracking wasm-latest instead. I'm not sure what @rossberg's plan is for the numbering, though.

The new directory will have to be created by copying the current wasm-latest directory while [using a merge commit to preserve its history](https://stackoverflow.com/questions/16937359/copy-file-preserving-history), then this merge will have to update both wasm-latest and the new wasm-3.1 (or whatever) directory identically.

(The goal of all this is to have all proposals update just wasm-latest so merges to or from the main spec repo apply changes to a single canonical location, while also having versioned directories with meaningful histories.)

> Should I copy the testsuite before-this-PR to spec-test-3 in what looks like a similar manner of how spec-test-{1,2} were copied?

Yes, that sounds like it's probably the expected workflow. I'm not sure what the expectations are around backporting new tests to these snapshots, but that doesn't directly affect this merge.

rossberg: [Sorry for the delay, back from holiday]

Right, the plan of record is that from 3.0 on, every new proposal merge (except perhaps a real big one) becomes a minor version bump. So that means that this PR should introduce a -3.1 spectec dir, and make -latest an alias for that. (And also update the checking script.)

As for text-interpreter/spec-test-3, I'd be fine with just leaving the symlink alone until we get an actual version 4. If we created a copy for every minor version, we'd quickly multiply the number of test suite runs of the meta interpreter, and they are very slow already.

The test failure with the meta interpreter is due to it not knowing how to compute with the new wide arithmetic operators. It will probably need an extension to list of primitives in backend-interpreter/numerics. Feel free to give it a go, otherwise I can try to review the changes and update the interpreter, but it might take a week or two before I get to it.

alexcrichton: I'm in no rush to get this merged myself, so taking longer is no worries. Would it be ok for the repo to be prepped to have this proposal landed, and then I'll arrange the commits and such to land on top of the scaffolding in-place?

rossberg: Well, not really, at least not on the main branch. All the structural changes ultimately ought to be part of the merge PR itself, since the repo would be in an inconsistent state otherwise. But we can (and should) build up everything on a pre-merge 3.1 branch.

Please feel free to prepare that branch, and then I can look into adding the necessary infrastructure adjustments.

PR #2225, opened 2026-08-11
[ci] Fix spec publishing in forks

kmiller68:
publish-spec lists ensure-wasm-latest in its needs, but that job only runs in WebAssembly/spec. A job whose needs include a skipped job is skipped as well, so forks (i.e. proposal repos) stopped publishing their rendered spec to GitHub Pages entirely once they picked up the ensure-wasm-latest job.

Guard publish-spec on nothing having failed instead, which tolerates the skipped job while still blocking publication when a build or the wasm-latest check fails.

### DIFF
--- a/.github/workflows/ci-spec.yml
+++ b/.github/workflows/ci-spec.yml
@@ -14,12 +14,12 @@ on:
 
 jobs:
   ensure-wasm-latest:
-    if: ${{ github.repository == 'WebAssembly/spec' }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
       - name: Diff wasm-latest
+        if: ${{ github.repository == 'WebAssembly/spec' }}
         run: cd specification && bash diff-wasm-latest.sh
 
   build-core-spec:


kmiller68:
I'm not 100% sure this is correct but it came up while I was working on automatically creating diffs for proposals. https://github.com/kmiller68/spec/actions/runs/31501085024 and seemed to be fixed with this PR: https://github.com/kmiller68/spec/actions/runs/31504405268

kmiller68:
CC @tlively Since I think this is caused by https://github.com/webassembly/spec/pull/2068

tlively (commented):
Makes sense, thanks. Do we also need to remove `ensure-wasm-latest` from the `needs` list?

kmiller68:
I don't think we need to, AFAICT, it would mean that failing the `ensure-wasm-latest` wouldn't block publication. Assuming we don't push things that fail a check it shouldn't matter. I also don't know how bad it is to publish something that fails the check. 

Another alternative might be to change: 

```
    ensure-wasm-latest:
      runs-on: ubuntu-latest
      steps:
        - name: Checkout repo
          uses: actions/checkout@v4
        - name: Diff wasm-latest
          if: ${{ github.repository == 'WebAssembly/spec' }}
          run: cd specification && bash diff-wasm-latest.sh
```

So the `ensure-wasm-latest` runs everywhere but just no-ops when not in `WebAssembly/spec`. Let me know if you prefer that and I'll switch.

tlively:
It's surprising that "a job whose needs include a skipped job is skipped as well" does not seem compatible with a solution that leaves ensure-wasm-latest in the needs list. I guess the `if` clause overrides the default condition that all the `needs` jobs must have passed?

Let's switch to your other proposed solution; it seems more obviously correct.

kmiller68:
> It's surprising that "a job whose needs include a skipped job is skipped as well" does not seem compatible with a solution that leaves ensure-wasm-latest in the needs list. 

Yeah, idk. I would have made it an `allSkipped` rather than `anySkipped` at least.

> Let's switch to your other proposed solution; it seems more obviously correct.

Sound good. I'll make the change.

kmiller68:
Here's the CI flow on main in a fork. (includes some diffing changes I made to so there's extra steps): https://github.com/kmiller68/spec/actions/runs/31623671275

tlively (approved):

PR #2226, opened 2026-08-14
[soundness] fix the frame rule again in instr_ok2

zilinc:
This PR fixes the wrong fix in #2212.
It's ported from wasm 2.0 soundness.
Frame_ok implies that `C'.RETURN = eps`.
(credits to @maximelegoupil)


rossberg:
I think it would be more consistent to change Frame_ok to add RETURN to the context. It already adds locals, so it would only be natural if it also adjusted everything else for the inner scope.

PR #2229, opened 2026-08-20
[ci] Render an upstream baseline copy of the spec in proposal repos

kmiller68:
Proposal repos are plain forks of `WebAssembly/spec`, so their published webassembly.github.io/<proposal>/ site has no rendered copy of the unmodified spec to compare against. The proposals often also lag behind the upstream spec so a direct comparison between the rendered specs is full of false differences. A reviewer who wants to see what a proposal actually does to the spec text has to build upstream themselves, or read source diffs, which is particularly awkward for parts of the spec which are generated in the output. e.g. typeset rules, tables, appendices, etc.

With this change a fork also renders every document at the commit where it diverged from its parent repo, and publishes the result under webassembly.github.io/<proposal>/upstream with the same layout as the main site. This will allow us to link to W3C's spec diff service and view a formatted and searchable comparison with a single URL. e.g.

https://services.w3.org/htmldiff?doc1=https://webassembly.github.io/<proposal>/upstream/core/bikeshed/&doc2=https://webassembly.github.io/<proposal>/core/bikeshed/

The baseline is chosen by a new resolve-baseline job, which asks the `gh` API for the repository's parent, fetches the parent's main, and takes `git merge-base`. For most proposals which sync by merging upstream, this resolves to the most recent sync point. When running on `WebAssembly/spec`, a fork with no changes of its own, or when the parent cannot be determined / fetched, the job short-circuits and no upstream copy is generated.

The commit upstream was rendered from is recorded at /upstream/baseline-sha and compared on the next run. When the shas match, the upstream variant is dropped from the build matrix and the published copy is carried over instead of being rendered again. This saves a decent amount of CI time for most (non-downstreaming) commits.

### DIFF
--- a/.github/workflows/ci-spec.yml
+++ b/.github/workflows/ci-spec.yml
@@ -15,6 +15,11 @@ on:
 permissions:
   contents: read
 
+# Path within the published site recording the commit the baseline was rendered
+# from, both as provenance and so a run can tell the copy is still current.
+env:
+  BASELINE_STAMP: upstream/baseline-sha
+
 jobs:
   ensure-wasm-latest:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


kmiller68:
For help with reviewing here are two CI runs: 

- https://github.com/kmiller68/spec/actions/runs/31828913256 a run with no baseline upstream
- https://github.com/kmiller68/spec/actions/runs/31829903826 a run with baseline upstream matching so no-rebuild

Published output from the two current runs:

- https://kmiller68.github.io/spec/ and https://kmiller68.github.io/spec/upstream/
- https://github.com/kmiller68/spec/blob/gh-pages/upstream/baseline-sha — bdd7164bfe18cf0bd5c3d90ef8cc3b8919fb9c0a
- The diff, two markers apart: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fkmiller68.github.io%2Fspec%2Fupstream%2Fcore%2Fbikeshed%2F&doc2=https%3A%2F%2Fkmiller68.github.io%2Fspec%2Fcore%2Fbikeshed%2F (may take about ~10-20s to load)

After writing this I also realized that gh-pages is actually a git repo itself. So in theory there are historical renderings from the merge-base. That said, there are a few reasons I uploaded this anyway: 

1. https://raw.githubusercontent.com/WebAssembly/spec/dbde29a3e/js-api/index.html is served as text rather than html and services.w3.org won't process it.
1. Historical builds of the spec could have different versions of rendering tools. 
1. Not every commit deploys so champions would have to find the most recent deployment when downstreaming.

kmiller68:
One other decision I made is to use `FETCH_HEAD` rather than hardcoding `WebAssembly/spec` as the upstream source. For most proposals these are the same but for ones dependent on another proposal they might depend on a different proposal. e.g. shared everything, which depends on threads (I assume anyway). I'm open to hardcoding `WebAssembly/spec` if folks think that's a better choice though.

kmiller68:
@rossberg @tlively Do either of you want to review this? it seems like you two have touch this file the most recently.

.github/workflows/ci-spec.yml
```diff
@@ -165,43 +282,76 @@ jobs:
...
+          set -euo pipefail
+          mkdir -p _output/upstream/metadata _output/upstream/legacy/exceptions
+          git show "$BASE_SHA:document/index.html" > _output/upstream/index.html
+          mv _upstream/core-rendered-upstream                     _output/upstream/core
+          mv _upstream/js-api-rendered-upstream                   _output/upstream/js-api
+          mv _upstream/web-api-rendered-upstream                  _output/upstream/web-api
+          mv _upstream/code-metadata-rendered-upstream            _output/upstream/metadata/code
+          mv _upstream/legacy-exceptions-core-rendered-upstream   _output/upstream/legacy/exceptions/core
+          mv _upstream/legacy-exceptions-js-api-rendered-upstream _output/upstream/legacy/exceptions/js-api
+          mv _upstream/versions-rendered-upstream                 _output/upstream/versions
```
tlively: Can we avoid listing out all these directories so we can't accidentally forget to add new directories in the future?

.github/workflows/ci-spec.yml
```diff
@@ -12,23 +12,99 @@ on:
...
+          }
+          trap write_outputs EXIT
+          if [ "$REPOSITORY" = "WebAssembly/spec" ]; then
+            echo "Canonical repo, no upstream baseline."
+            exit 0
+          fi
+          # A copy already on the site is left in place unless a newer baseline
+          # is rendered below.
+          published=$(gh api "repos/$REPOSITORY/contents/$BASELINE_STAMP?ref=gh-pages" \
+            --jq .content 2>/dev/null | base64 -d | tr -d '[:space:]') || published=''
```
tlively: https://github.com/kmiller68/spec/actions/runs/31828913256/job/94859702501 logs "base64: invalid input" here. Is that expected?

tlively (commented):
Nice! I will definitely use this.

rossberg (approved):
Cool, that looks super useful, even though I can't pretend to understand the git fu.

(Would there be a way to also rewrite local links to send you to the respective file diff instead of a 503? Then it would work for the non-Bikeshed version.)

.github/workflows/ci-spec.yml
kmiller68: Yeah, it probably should abort early. Fixed, I don't think it was harmful though.

.github/workflows/ci-spec.yml
kmiller68: I think this is a consequence of how the upload-artifact works. I changed it so each job stages it's build artifacts appropriately. Although that feels like something the Makefiles should be doing but those seem somewhat bit-rotted so it probably makes more sense to do that in a different PR.

kmiller68:
CI runs for the latest PR: 

  | # | Scenario | Run | Result |
  |---|---|---|---|
  | 1 | No `/upstream` on the site yet | [33130922877](https://github.com/kmiller68/spec/actions/runs/33130922877) | 17 jobs; both variants render; stamp written |
  | 2 | Merge-base matches the stamp | [33132330301](https://github.com/kmiller68/spec/actions/runs/33132330301) | 10 jobs; upstream variant dropped from the matrix; published copy carried over; stamp untouched |
  | 3 | Merge-base moved | [33134774643](https://github.com/kmiller68/spec/actions/runs/33134774643) | 17 jobs; re-rendered; stamp updated to the new sha |

kmiller68:
> (Would there be a way to also rewrite local links to send you to the respective file diff instead of a 503? Then it would work for the non-Bikeshed version.)

In theory yes, I think we'd want to do that globally though and have a `webassembly.github.io/<proposal>/diff` link for it.

kmiller68:
Actually for the multi-page version it might be kinda weird since this is just flowing through a diff tool. I think we'd end up with every link being a diff, which would drown the diff in noise. Links would probably have to be something the diff tool itself knows about. Maybe worth a feature request to https://github.com/w3c/htmldiff-ui once we know what we want.

kmiller68:
I don't think I have the power to merge so if someone could do it for me that would be appreciated.

PR #2231, opened 2026-08-24
[spec] Preserve host function arguments on divergence

sanguineman:
## Summary

Previously, the reduction rule for host function divergence dropped the call arguments `val^n` from the resulting configuration.

스크린샷 2026-08-24 오후 1 30 08

According to the WebAssembly specificaiton, the post-conditions for executing host functions must match those in the execution rule for invoking host functions. However, a host function that takes one or more arguments, one reduction step by `Step/call_ref-hostfunc-div` removes `val^n` and leaves only `(REF.FUNC_ADDR a) (CALL_REF yy)`. Neither host-function execution rule can then apply because the required operands are missing. 

## Fix 

The updated rule preserves the complete configuration, making the divergence case as exact self-loop as intended.


rossberg (commented):
Good catch, thanks!

rossberg (approved):

PR #2234, opened 2026-08-26
Require Python 3.10 in document README

bvisness:
Commit adc17e1 bumped the required version of Sphinx to 8.1.3, which requires Python >=3.10.


tlively (approved):
Thanks!

PR #2236, opened 2026-08-30
Add explicit permissions blocks to all workflows

gsnedders:
None of the four workflow files declared a permissions block, so every job ran with the default `GITHUB_TOKEN` scope.

What the defaults are depends on the org/repo configuration, and when the repo was created, per https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/

By declaring these explicitly, we guarantee that the actions do not have any write access.


tlively (approved):
Makes sense to me.

Issue #2221, opened 2026-08-02
[spectec] Wasm 1.0: `Module_ok` missing the validated import and export types


lzy0505:
The Wasm 1.0 `Module_ok` rule has the import external types `ixt*` and export external types `xt*` in its premises, but its conclusion returns only `OK`. 

Current rule:

```spectec
relation Module_ok: |- module : OK      hint(show "T-module")
```

The [WebAssembly Specification, Release 1.0 PDF](https://webassembly.github.io/spec/versions/core/WebAssembly-1.0.pdf), Section 3.4:

> “Then the module is valid with external types `it* → et*`.”

The conclusion of the formal rule is:

```text
⊢ {types functype*, funcs func*, tables table*, mems mem*, globals global*,
   elem elem*, data data*, start start?, imports import*, exports export*}
  : it* → et*
```

rossberg:
Right, PTAL at #2238.

Issue #2220, opened 2026-08-02
[spectec] Wasm 1.0: `Limits_sub` missing cases involving an absent maximum


lzy0505:
The Wasm 1.0 `Limits_sub` relation only matches limits for which both sides have an explicit maximum. 

Current spectec rule

```spectec
rule Limits_sub:
  |- `[n_11 .. n_12] <: `[n_21 .. n_22]
  -- if n_11 >= n_21
  -- if n_12 <= n_22
```

The [WebAssembly Specification, Release 1.0 PDF](https://webassembly.github.io/spec/versions/core/WebAssembly-1.0.pdf) Section 4.5.2 allows the maximum to be absent:

```text
n₁ ≥ n₂
────────────────────────────────────────────────────
⊢ {min n₁, max m₁?} ≤ {min n₂, max ε}
```

The fix should follow Wasm 3.0:

```spectec
rule Limits_sub/max:
  |- `[n_11 .. n_12] <: `[n_21 .. n_22?]
  -- if n_11 >= n_21
  -- (if n_12 <= n_22)?

rule Limits_sub/eps:
  |- `[n_11 .. eps] <: `[n_21 .. eps]
  -- if n_11 >= n_21
```

rossberg:
Right, PTAL at #2238.

PR #2238, opened 2026-09-01
[spectec] Backport limits fixes to 1.0/2.0 specs

### LINKED ISSUE #2219: [spectec] Wasm 1.0: `$growmemory` does not enforce the page limit
The Wasm 1.0 execution rule for `memory.grow` succeeds only if the resulting memory size is no greater than `2^16` pages. 

The official [WebAssembly Specification, Release 1.0 PDF](https://webassembly.github.io/spec/versions/core/WebAssembly-1.0.pdf) states:

> “If `len` is larger than `2¹⁶`, then fail.”

`growmem` includes:

```text
len ≤ 2¹⁶
```

Currently the spectec rule is:
```spectec
def $growmemory(mi, n) = mi'
  -- if mi = { TYPE `[i .. j?], BYTES b* }
  -- if i' = $(|b*| / (64 * $Ki) + n)
  -- if mi' = { TYPE `[i' .. j?], BYTES b* 0^(n * $($(64 * $Ki))) }
  -- (if i' <= j)?
```

It is missing the condition `  -- if $(i' <= 65536)`

rossberg:
Fixes #2219, #2220, #2221.

@lzy0505, PTAL.

### DIFF
--- a/specification/wasm-1.0/5-runtime-aux.spectec
+++ b/specification/wasm-1.0/5-runtime-aux.spectec
@@ -107,9 +107,11 @@ def $growtable(ti, n) = ti'
   -- if i' = $(|a*| + n)
   -- if ti' = { TYPE `[i' .. j?], REFS a* eps^n }
   -- (if i' <= j)?
+  -- if i' < $(2^32)
 
 def $growmemory(mi, n) = mi'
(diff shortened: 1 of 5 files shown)


lzy0505:
Thanks! Looks good to me.

PR #2240, opened 2026-09-01
Fix bugs in custom section tests

bvisness:
Custom section tests for branch hinting were actually already failing, but the failures were being suppressed by a bad Makefile rule that was yielding an `0` exit code. In addition, a couple kinds of custom-section exceptions were not being handled like they should have been. (For example, the existing failing branch-hints test was actually completely fine, except that Custom.Invalid was not being caught, so the `assert_invalid_custom` was failing.)


rossberg (approved):
Nice find, thanks!

PR #2241, opened 2026-09-02
Remove pop_num from the validation algorithm

tlively:
It was not used.


rossberg:
Hm, I think I'd like to keep it for completeness, even though none of the example cases use it. If that's too odd, perhaps rather include a case that uses it?

tlively:
Sure. Is it used for anything except for unannotated select?

PR #2242, opened 2026-09-02
Ensure that inline (data)s and (elem)s count toward identifiers

bvisness:
The `(memory (data ...))` and `(table (elem ...))` abbreviations were not correctly counting the indices of the data and elem segments, causing the identifiers' indices on later data and elem segments to be off by one. Because the identifier context needs to be updated as we parse, `memory_fields` and `table_fields` are updated to do some of their work in the first pass.

### DIFF
--- a/interpreter/text/parser.mly
+++ b/interpreter/text/parser.mly
@@ -1117,9 +1117,9 @@ memory :
 
 memory_fields :
   | memorytype
-    { fun c -> fun x loc -> [Memory ($1 c) @@ loc], [], [], [] }
+    { fun c x loc -> [Memory ($1 c) @@ loc], [], [], [] }
   | inline_import memorytype  /* Sugar */
-    { fun c -> fun x loc ->
+    { fun c x loc ->
       [], [],
       [Import (fst $1, snd $1, ExternMemoryT ($2 c)) @@ loc], [] }
   | inline_export memory_fields  /* Sugar */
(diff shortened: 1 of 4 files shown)


interpreter/text/parser.mly
```diff
@@ -1112,20 +1112,23 @@ memoryuse :
...
-      fun () -> $4 c x $sloc }
+      let mff = $4 c in
+      fun () -> mff x $sloc }
 
 memory_fields :
   | memorytype
-    { fun c x loc -> [Memory ($1 c) @@ loc], [], [], [] }
+    { fun c -> fun x loc -> [Memory ($1 c) @@ loc], [], [], [] }
   | inline_import memorytype  /* Sugar */
-    { fun c x loc ->
```
rossberg, line 1121: Nit: these changes are redundant, since the old version is sugar for the new.

interpreter/text/parser.mly
```diff
@@ -1185,23 +1188,26 @@ tableuse :
...
   | tabletype constexpr1
-    { fun c x loc -> [Table ($1 c, $2 c) @@ loc], [], [], [] }
+    { fun c -> fun x loc -> [Table ($1 c, $2 c) @@ loc], [], [], [] }
   | tabletype  /* Sugar */
-    { fun c x loc -> let TableT (_, _, (_, ht)) as tt = $1 c in
+    { fun c -> fun x loc -> let TableT (_, _, (_, ht)) as tt = $1 c in
       [Table (tt, [RefNull ht @@ loc] @@ loc) @@ loc], [], [], [] }
   | inline_import tabletype  /* Sugar */
-    { fun c x loc ->
+    { fun c -> fun x loc ->
```
rossberg: Same here.

rossberg (commented):
Oh, thanks! That must have been a pain to debug, since this code is a mess.

redianthus:
I feel like having a test for this would be a nice addition, would you be willing to add one @bvisness?

bvisness:
I went back and forth on including tests because they are very fiddly, identifier-specific, and...where would they go? Do we have a place for "regression tests" of sorts? I'm happy to add one, it's just so niche that I struggled with it.

rossberg:
I think elem.wast and data.wast would be natural places. They aren't really regression tests (nor do I see a need to distinguish those in the test suite).

bvisness:
I added some tests that fail on main with an out-of-bounds access due to the identifier referring to the dropped data/elem segment.

rossberg (approved):

PR #2243, opened 2026-09-03
[spectec] Fix scoping of constructor parameters during dimension analysis

### LINKED ISSUE #2227: [spectec] Incorrect front-end elaboration
MWE:
```
;; .spectec source
syntax t =
  | ATOM nat? int  -- if (int < 2 /\ nat > 0)?
```
and it's elaborated into:
```
;; elab.spectec:1.1-2.47
syntax t =
  | ATOM(`nat?` : nat?, int : int) {`int?` : int?}
    -- (if ((int < (2 : nat <:> int)) /\ (nat > 0)))?{int <- `int?`, nat <- `nat?`}
```
The quantifier list of the type case and that of the side condition are both wrong. `int?` shouldn't be there.

(Credits to @maximelegoupil and @DCupello1)

The real use case is the `STORE` instruction in the Wasm specs. It was uncovered during the Wasm 2.0 proof.

rossberg:
Fixes #2227. @zilinc, PTAL.

### DIFF
--- a/spectec/src/frontend/elab.ml
+++ b/spectec/src/frontend/elab.ml
@@ -1034,7 +1034,11 @@ and elab_typ_notation is_unboxed env outer_dims tid at (t : typ) (prems : prem n
   let env1 = local_env env in
   let mixop, xts' = elab_typ_notation' env1 tid t in
   let xs', ts' = List.split xts' in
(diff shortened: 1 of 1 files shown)


zilinc:
Thanks! LGTM.

PR #2247, opened 2026-09-08
[spec] Add load rules missing from document

### LINKED ISSUE #2244: [spec] Bugs / questions about `VLOAD` in Wasm 2.0
In the Wasm 2.0 spectec specification, the instruction `VLOAD` can optionally take a `vloadop`: https://github.com/WebAssembly/spec/blob/main/specification/wasm-2.0/1-syntax.spectec#L510
but I can't find its counterpart in the manually written spec. 
There isn't a relevant typing rule for the `eps` case, but there're reduction rules for it: https://github.com/WebAssembly/spec/blob/main/specification/wasm-2.0/8-reduction.spectec#L505-L511

So we'd like to get some clarification on this discrepancy. 

Incidentally, I think the show hints in this block are all wrong---they all take one more argument marker `%`: https://github.com/WebAssembly/spec/blob/main/specification/wasm-2.0/1-syntax.spectec#L506-L513

(cc: @maximelegoupil)

rossberg:
Fixes #2244 and a few other minor rule naming inconsistencies that led to unintended prose.

@zilinc, PTAL.


zilinc:
Looks good. Thanks.

PR #2251, opened 2026-09-22
Fix finding exceptions test files in `run.py`

fitzgen:


tlively (approved):
Thanks! What were the effects of this bug? I guess the exceptions tests were being skipped? Yikes!

fitzgen:
> Thanks! What were the effects of this bug? I guess the exceptions tests were being skipped? Yikes!

If `run.py` was executed without explicit CLI arguments specifying which files/dirs to run, then exceptions tests were being skipped, yes.

